PR #188, opened 2014-09-22
install: add GOPATH or GOBIN to the list of executables

### LINKED ISSUE #186: Append go_bin_path before executing any command.

fatih:
Fixes #186

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -48,7 +48,12 @@ fu! s:gocodeCommand(cmd, preargs, args)
         let a:preargs[i] = s:gocodeShellescape(a:preargs[i])
     endfor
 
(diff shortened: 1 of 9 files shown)


luan:
Hi @fatih, any particular reason for *append* instead of *prepend*? This is causing issues if people have some other binary in their path with a colliding name, because appending makes our `gobin` the last lookup path rather than the first.

I'm getting several reports of people hitting this issue because they had tools compiled on 1.6 and were hanging out in their `$GOPATH/bin` and now that we're using `g:go_bin_path` that is not getting updated, short term fix for us is just removing those stale binaries but seems like a prepend to the `$PATH` would be appropriate.

Thoughts?

Issue #281, opened 2015-01-05
GoDoc can not found correct package path when package name has extension


blu3gui7ar:
An example is 

import "gopkg.in/yaml.v2"

A quick fix might be 

diff --git a/autoload/go/tool.vim b/autoload/go/tool.vim
index b7e040b..6195305 100644
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -32,7 +32,7 @@ function! go#tool#Imports()
     endif

     for package_path in split(out, '\n')
-        let package_name = fnamemodify(package_path, ":t")
+        let package_name = fnamemodify(package_path, ":t:r")
         let imports[package_name] = package_path
     endfor

fatih:
Can you open a PR with a test file that shows the error and your fixes? We can include and merge it back to the master. Thanks.

PR #284, opened 2015-01-08
[Bugfix]GoDoc can not found correct package path when package name has ext

blu3gui7ar:
https://github.com/fatih/vim-go/issues/281

To reproduce the problem in following code:

put cursor on "yaml.Unmarshal", and run `:GoDoc`

```
package main

import (
	"io/ioutil"

	"gopkg.in/yaml.v2"
)

type config struct {
}

func test() {
	if content, err := ioutil.ReadFile("nothing"); err == nil {
		c := config{}
		err = yaml.Unmarshal(content, &c)
	}
}
```

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -32,7 +32,7 @@ function! go#tool#Imports()
     endif
 
     for package_path in split(out, '\n')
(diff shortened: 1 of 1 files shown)


fatih:
Thanks for the PR :+1:

Issue #217, opened 2014-10-30
"-- More --" on write


spf13:
Whenever I save a go file it appears to read it in a less style buffer prompting me for "-- More --". It also doesn't seem to work properly with the formatting / importing. All other aspects of the plugin seem to be working fine.  

Manually running :GoFmt or :GoImports has the exact same effect. 

Not really sure what is causing it or even where to start debugging it.

andrewaustin:
I also have this problem on Ubuntu 14.04.1 using Go 1.4 and vim 7.4 (installed via the vim-nox package). 

Unlike @mtchavez this is a fresh install of Go, so I only have fresh binaries as a result of ```:GoInstallBinaries```. I tried running ```:GoUpdateBinaries``` just to be safe though, but this did not work.

My .vimrc is tiny:

```
execute pathogen#infect()
syntax on
set nu
filetype plugin indent on
Helptags
let g:go_fmt_command = "goimports"
```

Removing ```let g:go_fmt_command = "goimports"``` from my .vimrc fixed the issue. I tried running ```go get golang.org/x/tools/cmd/goimports``` but still had the same issue.

paytonrules:
Since upgrading from go 1.3.3 to go 1.4 I get the problem above, plus numerous others. What I had to do was make sure 1.3.3 was missing from my system, cleanout all the installed binaries for gocode, godoc, etc. and reinstall them using :GoInstallBinaries. Doing it manually did not work.

dewyze:
I have tried many of the solutions here, completely blown away Go and reinstall, updated binaries, and am having the same issues. Interestingly, I have two machines that have the same dotfiles and this is only happening on one. I am using native Mac terminal.

fatih:
@dewyze do they also have the same Vim version? I wonder if it's happening only for a specific version.

dewyze:
I am going to look into that later tonight, I am only have one currently with me. And it is only when using: `let g:go_fmt_command = "goimports"`

dewyze:
I had two different versions of this Plugin on both computers, I used `vim -V` to figure out what was being called and realized that in order for it to work it was passing -1 to [here](https://github.com/fatih/vim-go/blob/master/autoload/go/fmt.vim#L76).

pzduniak:
Same issue using Cygwin.

vessenes:
I too had this problem on Ubuntu 14.04 after upgrading to go 1.4.

[...]

PR #286, opened 2015-01-13
Fix go#fmt#Format for using GoImports

dewyze:
Pull request to fix the [Issue 217](https://github.com/fatih/vim-go/issues/217). 1 needs to be passed to [here](https://github.com/fatih/vim-go/blob/master/autoload/go/fmt.vim#L76).

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -186,7 +186,11 @@ augroup vim-go
 
     " code formatting on save
     if get(g:, "go_fmt_autosave", 1)
-        autocmd BufWritePre *.go call go#fmt#Format(1)
+        if exists("g:go_fmt_command")
+            autocmd BufWritePre *.go call go#fmt#Format(1)
+        else
+            autocmd BufWritePre *.go call go#fmt#Format(-1)
+        endif
     endif
 
 augroup END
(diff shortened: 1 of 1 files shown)


dewyze:
I now realize this is not the final fix, there needs to be a check whether or not the flag is present, but part of the problem is here.

dewyze:
Otherwise it will always run goImports.

dewyze:
Added second commit to use gofmt if `g:go_fmt_command` does not exist.

fatih:
Hi @dewyze Thanks for the fix. I need to check first, because we already are doing a check inside the `Fmt` function. I'm going to look at it asap.

casimir:
Hi, is there any workaround while waiting for the merge? Like setting a variable in ```.vimrc```?

fatih:
I've merged this one: https://github.com/fatih/vim-go/pull/294. The checking shouldn't be inside plugin/go.vim because this issue needs to be fixed inside fmt.vim. We now check for the existence of the binary if it's something else than `gofmt`. @dewyze's PR was causing to branch the logic and cause to check the binary, so basically we do it no with #294

PR #290, opened 2015-01-16
Update docs for function text objects.

dpetersen:
It looks like the docs aren't quite right in https://github.com/fatih/vim-go/pull/272. They show `am`/`im` when the implementation is `af`/`if`. I'm somewhat awful at vimscript so maybe I'm missing something, but only the `f` variant works for me in vim.


fatih:
That's right. It seems I didn't paid enough attention to this. Thanks for fixing :+1:

PR #291, opened 2015-01-17
syntax:hightlights build constraints

caixw:


fatih:
That's really nice. However syntax is expensive and is causing performance problems on some machines. That's why syntax options are disabled by default. Can you make it so it's disabled by default? Once you do we can merge it :shipit:

caixw:
:ok:

fatih:
Thanks :+1:

PR #292, opened 2015-01-19
Add vimproc support for vim-go.

jsfaint:
On windows, when call the system() function, a console window will be showed up, which will effect the user exprience terribly.
for example, when the auto-complete was enabled every single time a key pressed a console will interrupt the auto-comple window.
It's a nightmare :(
With the help of vimproc, that nightmare is gone. :)

### DIFF
--- a/autoload/go.vim
+++ b/autoload/go.vim
@@ -1,0 +1,17 @@
+"Check if has vimproc
+function! go#has_vimproc()
+    if !exists('g:vimgo#use_vimproc')
+    try
+      call vimproc#version()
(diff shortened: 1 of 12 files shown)


fatih:
That's a nice addition :+1:  We need to be careful though, so it doesn't break anything if the user doesn't have vimproc installed. 

Can you rebase it onto master again? It seems after the errcheck change it broke your PR. Once you rebase I can check, test and look at it more detailed.

jsfaint:
OK, I'll close this pr and create a new one:) Thanks

PR #296, opened 2015-01-19
Add vimproc support for vim-go.

jsfaint:
On windows, when call the system() function, a console window will be showed up, which will effect the user exprience terribly.
for example, when the auto-complete was enabled every single time a key pressed a console will interrupt the auto-comple window.
It's a nightmare :(
With the help of vimproc, that nightmare is gone. :)

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -19,7 +19,12 @@ fu! s:gocodeCurrentBuffer()
     return file
 endf
 
-let s:vim_system = get(g:, 'gocomplete#system_function', 'system')
+
+if go#vimproc#has_vimproc()
+    let s:vim_system = get(g:, 'gocomplete#system_function', 'vimproc#system2')
+else
+    let s:vim_system = get(g:, 'gocomplete#system_function', 'system')
+endif
 
 fu! s:system(str, ...)
     return call(s:vim_system, [a:str] + a:000)
--- a/autoload/go/vimproc.vim
+++ b/autoload/go/vimproc.vim
@@ -1,0 +1,21 @@
+"Check if has vimproc
+function! go#vimproc#has_vimproc()
+    if !exists('g:go#use_vimproc')
+        if IsWin()
+            try
+                call vimproc#version()
+                let exists_vimproc = 1
+            catch
+                let exists_vimproc = 0
+            endtry
+        else
+            let exists_vimproc = 0
+        endif
+
+        let g:go#use_vimproc = exists_vimproc
+    endif
+
+    return g:go#use_vimproc
+endfunction
+
+" vim:ts=4:sw=4:et


fatih:
Hi @jsfaint. Thanks for your efforts. I don't use Windows so I can't test it and see the benefits of it. I'll ask some other Windows users and get their feedback. On Mac it's just working without any problem.

autoload/go.vim
```diff
@@ -0,0 +1,17 @@
+"Check if has vimproc
```
fatih: Can you move this to `autoload/go/vimproc.vim` please. I keep vim-go related autoload files in the `autoload/go` folder.

autoload/go.vim
```diff
@@ -0,0 +1,17 @@
...
+function! go#has_vimproc()
+    if !exists('g:vimgo#use_vimproc')
```
fatih: We use `go` as prefix for global variable, so it should be `g:go#use_vimproc`

fatih:
@alexaandru mind you could look at this?

jsfaint:
Hi @fatih I changes the part you metioned. :smile: 

But there is still one part I haven't add support for vimproc.
In `go#fmt#Format(withGoimport)` function, it called `%!` command
```
" execufe gofmt on the current buffer and replace it
silent execute "%!" . command
```

As the vimproc doc said, we can `:VimProcBang` command to replace `:!` command, but `:VimProcBang` don't support range
If I run `:%VimProcBang`, it will show *E481: No range allowed*

But fortunately, this experience is not too bad to force me add vimproc for it. :laughing:

alexaandru:
I don't exactly understand the benefits and unsure what to test, but here's what I tried: I tried the Omnicomplete as well as GoRun features. Attaching screenshots before applying the patch and after. Both work fine before and are broken after applying the patch. With the latest vim-go from master, I did not experience the issue mentioned in the patch initially (that "every single time a key pressed a console will interrupt the auto-comple window"), the autocomplete worked naturally, as it does on Linux, no console interrupted the process.

I did install vimproc as per its install notes (installed MinGW, and build the native extension).

For some reason, uploading pics here doesn't work at the moment, so I put them in https://gist.github.com/alexaandru/d116c229c07983f80b03

jsfaint:
@alexaandru It's werid, there are something goes wrong.
On my platform I install neobundle as plugin manager and neocomplete as auto-complete plugin for windows. and youcomopleteme for linix/mac.
Maybe some plugins were conflict with vim-go.
Because gorun command never work on my windows, but it's ok on linux and mac.
I'll check it tomorrow and let you know the result.

alexaandru:
I only use the Windows box for testing. The only plugins I have installed there are: vundle, vim-go and now vimproc. Don't have anything else, so there's not much to conflict with vim-go. Don't have anything for autocomplete other than using omnicomplete via C-x C-o

jsfaint:
@alexaandru I record two gif to explain the original issue what I talking about.
I use neocomplete for code completion, without vimproc support, the complete window will not show up. it was interrupted by the console window.
without_vimproc
As you said you tested the omnicomplete, it don't have this issue.

After add vimproc support, neocomplete works fine.
with_vimproc

I already reappear the GoRun issue you mentioned, I'll fix it later.
Thank you!

fatih:
I wonder if this is related to neocomplete only.  I think we should only add this to the complete file, all others should be not touched (to avoid possible recessions).

jsfaint:
Although everytime system() called or execution an external command will show a console under Windows.
but only neocomplete need to be fixed. Other conditions can be acceptable.

So I think it's OK to fix complete file only.

plugin/go.vim
```diff
@@ -140,7 +140,11 @@ function! s:GoInstallBinaries(updateBinaries)
                 echo "vim-go: ". basename ." not found. Installing ". pkg . " to folder " . go_bin_path
             endif
 
-            let out = system("go get -u -v ".shellescape(pkg))
+            if go#has_vimproc()
+                let out = vimproc#system2("go get -u -v ".shellescape(pkg))
+            else
+                let out = system("go get -u -v ".shellescape(pkg))
+            endif
```
fatih: Can you also remove these lines please?

autoload/go/def.vim
```diff
@@ -20,8 +20,8 @@ function! go#def#Jump(...)
 
 	let command = bin_path . " -f=" . expand("%:p") . " -i " . shellescape(arg)
 
-	" get output of godef
-	let out=system(command, join(getbufline(bufnr('%'), 1, '$'), LineEnding()))
+  " get output of godef
+  let out=system(command, join(getbufline(bufnr('%'), 1, '$'), LineEnding()))
```
fatih: Can you revert these whitespace changes?

plugin/go.vim
jsfaint: Sorry, I forget this place.

autoload/go/def.vim
jsfaint: OK, I just checked, in def.vim it use tab as indent, but in my environment I use space. I'll revert it.

autoload/go/def.vim
```diff
@@ -39,7 +39,7 @@ function! go#def#JumpMode(mode)
 	let command = bin_path . " -f=" . expand("%:p") . " -i " . shellescape(arg)
 
 	" get output of godef
-	let out=system(command, join(getbufline(bufnr('%'), 1, '$'), LineEnding()))
+  let out=system(command, join(getbufline(bufnr('%'), 1, '$'), LineEnding()))
```
fatih: Can you revert this too?

fatih:
@jsfaint Thanks again :+1: . After you finished please rebase it to a single commit so I can merge it.

jsfaint:
@fatih I just learned how to rebase to single commit, thanks for this opportunity :laughing: 
Now I knew that I shouldn't close the first PR :cry: 
Thank you! :+1:

fatih:
Thanks @jsfaint and congrats to your first contribution to vim-go :+1:  :)

PR #297, opened 2015-01-19
Update readme.md

nbari:
added a ``comand not found header`` explaining how to make the  plugin work when using csh and replaced /bin/bash with /bin/sh since not all the *nix bring by default /bin/bash in whereas /bin/sh is always available


fatih:
Thanks @nbari :+1:

PR #299, opened 2015-01-20
GoImport can import "encoding/json" with `:GoImport json`

ppalucki:
Actually GoImport can work only for narow number of packages. This is **sad**.

The package have to be flat and the import path have to reflect exactly package name.

This fix tries to infer "real" import path when executed on package name.

For example:

```go
package main
func main(){
    json
}
``` 

Actually when GoImport is executed on "json" it cannot import "encoding/json", This merge fixes that by creating mapping between "package name" and "import path" and injecting "real import path" when needed.

As a bonus, this fix can import not just built-in packages but also external packages if they were installed earlier.

Try this:

```sh
go get "github.com/bitly/go-simplejson"
```

and then:
```vim
:GoImport simplejson
```

the effect is:
```go
package main

import (
    "github.com/bitly/go-simplejson"
)
```

Warning - it takes time, especially on big projects - its a problem of `go list all` command.

It is **very useful** because in go, I have too frequently comment/uncomment parts of code and then fix imports manually to get rid of "import not used" error.
I can use `GoImports` to remove unused packages but then I have import all of them manually - a now it would be possible to just point on missing package and do `GoImport`.

ps. this requires +python in vim, because I am not able to write "vim script" - sorry - but it should be possible to rewrite it to just clean vimscript

### DIFF
--- a/autoload/go/import.vim
+++ b/autoload/go/import.vim
@@ -56,6 +56,56 @@ function! go#import#SwitchImport(enabled, localname, path)
     endif
 
     let exists = go#tool#Exists(path)
+    
+    " try to infer import_path from package_name but only if python is
+    " installed and given path doesn't not reflect actual package
+    if has('python')
+    
+py <<EOF
+import subprocess, collections, vim
+def find_import_path(package_name):
+    """find real import path for given package name
+    warning: takes time
+    """
+
+    # prepare mapping between "package name" and "import path" (possbile many)
+    mapping = collections.defaultdict(list)
+
+    # ignore errors and create mapping between name and import path
+    output = subprocess.check_output("go list -e -f {{.Name}};{{.ImportPath}} all".split(' '))
+    for l in output.split("\n"):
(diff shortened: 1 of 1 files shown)


fatih:
Hi @ppalucki. Thanks for the effort. A lot of people are using `GoImports` lately, so they don't deal with manually importing/dropping import. Btw, `GoImports` also imports all automatically, so I wonder why it fails for your? 

I need to check and rewrite this in Viml again as I avoid any non viml extensions (maintenance is easy)

ppalucki:
:-1: for me 

wow :) you are right - my mistake

`GoImports` actually do what I wanted and do this a lot more better than this pull request.

I thought that `GoImports` can only manage imports the same way as `GoImport`

I missed that - I won't be upset if you just discard this pull request

fatih:
You're welcome. I'll close this issue as I really don't have time for now to rewrite it with VimL. But thanks for the efforts, really!

PR #303, opened 2015-01-31
Goinfo typo

friesencr:
Thank you for working on this project, it has accelerated my learning and it is pretty fun having so much power in vim for go!


fatih:
Thanks :)

PR #304, opened 2015-02-01
Support function literals in `af` and `if` text objects

nilium:
This is sort of a preliminary thing more for discussion than anything else. _I strongly discourage merging this_ in because of the flaw mentioned in the commit message right now (below). I'm also not even sure if this is the sane way to be doing this, I just kind of threw it together over the last hour-ish.

The changes basically switch things around such that `if` still does more or less the same thing it already does where `af` will use visual mode to select only the extents of a function definition (i.e., all of `func …(…) { … }`). This works fine assuming that the return type of a function is either a) a tuple or b) anything else or c) nothing. This does not work if the function returns a `func(…)` (regardless of that func's return type), mainly because it looks like a completely normal function literal if read in reverse, which is what the `a` path does with these changes.

So, I figured I'd throw this up as a pull request and see if you or anyone else watching this had any ideas on how to proceed with this.

Edit: For a slightly funnier option, consider a return type of `func() func() func()`.

_Actual commit message follows._

----

Works fine except for the case were a function is returning only a
function (e.g., `func() func() { ... }`). Not sure if there's a good way
around this because that would also be a valid function literal, so the
main issue is trying to sort-of parse this in reverse using just motions
and the character under the cursor doesn't work all that well.

Probably need a forward parser.

### DIFF
--- a/autoload/go/textobj.vim
+++ b/autoload/go/textobj.vim
@@ -2,14 +2,72 @@ if !exists("g:go_textobj_enabled")
     let g:go_textobj_enabled = 1
 endif
 
+let s:funcDecl = '\(^\|\W\|\s\)\@<=func\(\s\|(\).*{'
+let s:funcBegin = '\(^\|\W\|\s\)\@<=func\(\s\|(\)'
+
+function! s:char()
+  " Borrowed from Ingo Karkat's answer on Stack Overflow:
+  " http://stackoverflow.com/a/23323958/457812
+  return matchstr(getline('.'), '\%'.col('.').'c.')
+endfunction
+
 function! go#textobj#Function(mode)
-  if search('^\s*func .*{$', 'Wce', line('.')) <= 0
-        \ && search('^\s*func .*{$', 'bWce') <= 0
-    return
-  endif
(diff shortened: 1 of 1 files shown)


nilium:
Also fails for functions whose signatures cover multiple lines. For example:

```go
var somefunc = func(
	x float64,
	y string,
	z int,
) {
	println("Beep", x, y, z)
}
```

fatih:
Hi @nilium. Thanks for the PR. I'l look into it ASAP :)

fatih:
You are right. It's wonderful that it supports anonymous functions, but it breaks current `af` on global functions. I really want to merge this due the support for closures,  let me check too, maybe others can join too to have a fix for it.

fatih:
I'll close this as I don't see any improvements in short term. I think we need tools who understand the syntax and are able to parse the AST of a given go source code. Vim is not suitable for it (unless we have a go parser written in viml, which I think is an ambitious project). Thanks @nilium for your works.

nilium:
Agreed. Definitely don't want to do this in viml. Might be possible to use oracle, but I just haven't looked into it enough, and haven't needed this with any urgency. Might revisit it later.

PR #308, opened 2015-02-07
Use shiftwidth() instead of '&shiftwidth'

fd0:
I'm using vim-sleuth[1] from tpope to automatically detect indent
settings for files that I edit. The latest version sets 'shiftwidth' to
zero for files that are detected to be indented only with hard tabs, so
that the value of 'tabstop' is used instead of 'shiftwidth'. This
change[2] was done to allow people setting 'tabstop' to their preferred
value, and to have 'shiftwidth' automatically follow.

The vim documentation about 'shiftwidth' states:

     When zero the 'ts' value will be used.  Use the shiftwidth()
     function to get the effective shiftwidth value.

Therefore this commit changes using 'sw' directly and instead calls
shiftwidth().

[1] https://github.com/tpope/vim-sleuth
[2] https://github.com/tpope/vim-sleuth/commit/b6b4c3b237678f1214ea2eca0e3e50eaafb30e3a

### DIFF
--- a/indent/go.vim
+++ b/indent/go.vim
@@ -37,20 +37,21 @@ function! GoIndent(lnum)
   let previ = indent(prevlnum)
 
   let ind = previ
+  let s:shiftwidth = shiftwidth()
 
   if prevl =~ '[({]\s*$'
     " previous line opened a block
-    let ind += &sw
+    let ind += s:shiftwidth
   endif
   if prevl =~# '^\s*\(case .*\|default\):$'
     " previous line is part of a switch statement
-    let ind += &sw
+    let ind += s:shiftwidth
   endif
   " TODO: handle if the previous line is a label.
 
   if thisl =~ '^\s*[)}]'
     " this line closed a block
-    let ind -= &sw
+    let ind -= s:shiftwidth
   endif
 
   " Colons are tricky.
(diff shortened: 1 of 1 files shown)


fatih:
What is the current value of `&sw` ? Still `tabstop` or something else ?

fd0:
I'm sorry, I should've explained it better.

When `&shiftwidth` is set to 0, the indentation for Go files doesn't work at all, whereas for other filetypes (e.g. C) it works perfectly. I discovered this after upgrading to the latest version of vim-sleuth, and was irritated that after typing "{<CR>" no indentation was made. You can easily reproduce this behaviour by opening a Go file and setting `&shiftwidth` to 0.

The vim manual states that when `&shiftwidth` is 0, the value of `&tabstop` is used for indentation internally. In this case, reading `&shiftwidth` yields 0, so vim-go does not indent at all. The manual suggests that instead of reading `&shiftwidth`, the function `shiftwidth()` should be called. When `&shiftwidth` is set to some value other than zero, this value is returned, and the value of `&tabstop` in the case when `&shiftwidth` is zero.

Do you need more clarification?

fatih:
Thanks for the detailed explanation. Now it makes totally sense. I've tested it and this PR indeed fixes it in the way you described. Thanks :+1:

Issue #250, opened 2014-12-02
Oracle


DAddYE:
Hello there!

I was looking for `GoOracle` and I didn't find the command anymore. What happened? 

I found that now the lines are commented: https://github.com/fatih/vim-go/blob/master/ftplugin/go/commands.vim#L79-L88

However I remember I saw a recent PR for improving oracle support.

yasker:
Thank you Fatih for this brilliant work! I hope `:GoCallers` can be the next one in the queue since I found it would be another essential tool to read the code.

hoffoo:
:+1: for :GoCallers

ghodss:
+1 for :GoCallers as well

fatih:
I've created a branch that supports all Oracle commands. I am no longer using a custom UI, instead I'm parsing Oracle's `plain` format. After the PR #311, I've rethought my initial decision and decided to go with a simpler way (using quickfix). I've tested each mode extensively and they all work perfectly. For testing you can use the following example file that I've used for testing it. I also added documentation for each mode, of course more info can be find in the [Oracle User Manual](golang.org/s/oracle-user-manual).

Example file to test all Oracle modes:

```go
package main

import "fmt"

type Storage interface {
	Get() string
}

func main() {
	fmt.Println("vim-go")

	i := &intStorage{}
	s := &stringStorage{}

	done := make(chan bool)

	callStorage(i, done)
	callStorage(s, done)
	<-done

}

func callStorage(s Storage, done chan bool) {
	res := s.Get()
	fmt.Printf("res = %+v\n", res)

	a := 5
	foo(a)

	done <- true
}

type intStorage struct{}

func (i *intStorage) Get() string {
	return "intStorage"
}

type stringStorage struct{}

func (s *stringStorage) Get() string {
	return "stringStorage"
}

func foo(b int) {
	fmt.Println("b", b)
	bar()
}

func bar() {
	fmt.Println("fatih arslan")
}
```

I'll going to merge the pr #319 soon. Please test and any feedback is welcome :)

yasker:
Wow, it works great for me! Still using it but so far so great! Thank you Faith!

fatih:
Oracle support landed now into the master branch.

ghodss:
Just want to pile on thanks for this, this feature is AWESOME, very well done!

akavel:
[...]

PR #311, opened 2015-02-09
Enable oracle referrers command

akavel:
Put the results into quickfix list for easy browsing.

### DIFF
--- a/autoload/go/oracle.vim
+++ b/autoload/go/oracle.vim
@@ -262,44 +262,25 @@ function! go#oracle#Referrers(selected)
 
     " get the referrers list
     if has_key(out.referrers, "refs")
-        let referrers = out.referrers.refs
+        let refs_buf = out.referrers.refs
     else
         redraw | echon "vim-go: " | echon "no referrers available"| echohl None
         return
     endif
 
-    let result = []
-    for ref in referrers
+    let desc = out.referrers.desc
+
+    let referrers = ["Referrers for: " . desc]
+    for ref in refs_buf
         " strip column number, add dummy "..." for correct quicklist format
         " TODO: would be awesome if the contents of the line could be shown
         " here...
         let line = join(split(ref, ":")[0:-2] + ["..."], ":")
-        call add(result, line)
+        call add(referrers, line)
     endfor
 
-"    cgetexpr referrers
-    cgetexpr result
+    cgetexpr referrers
     copen
-"    let title = "Referrers:"
-"    "
-"    " start to populate our buffer content
-"    let result  = [title, ""]
-"
-"    for ref in refs
-"        let line = "\t" . ref
-"        call add(result, line)
-"    endfor
-"
-"    " open a window and put the result
-"    call go#ui#OpenWindow("Referrers", result)
-"
-"    " define some buffer related mappings:
-"    "
-"    " go to definition when hit enter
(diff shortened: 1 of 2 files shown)


fatih:
Hi @akavel 

Thanks for the PR. Couple of things:

* This is how it looks like for me: http://d.pr/i/KIQP This is also the main reason I don't use QuickFix for Oracle. It's just doesn't fit. That's why I've created a custom UI for it. I see that you removed that part, but I'm parsing JSON just for this reason, to have a nice UX around this. We have already ready to use commands that uses those functions and UI. 
* Docs are missing
* Mappings are missing (Plugs).

After you are finished, please don't forget to rebase it :)

akavel:
Thanks for having a look.

1. As to how it looks: um; I'm not sure what's there that you don't like? I'm not being ironic, just asking. As to the helper ui functions, I've seen them and tried first, but I missed that the window disappears after hitting Enter, so I couldn't easily skip from case to case; also I know I sometimes use `:cn`, `:cp` etc. I wasn't sure if the alternative ui is on purpose, or accidental. Can I ask what are the advantages? Also, I thought, maybe there could be option for .vimrc for choosing if one prefers the quickfix list or the special-purpose ui, global for all the commands? What would you think?
2. Could you help me a bit more, where should I add the docs? This PR is totally my first ever foray into vimscript, so I'm absolutely green with regards to conventions etc.
3. Same with regards to the mappings/plugs; what they even are? :( I'd be really grateful for some hints... thanks!

fatih:
If you start to use it first, it seems quickfix is the obvious solution for this kind of task. However I initially didn't liked the output of quickfix and how it looks like. Check the following examples for the `implements` mode:

This is quickfix list:

image

This is when you use the custom UI functions in vim-go (which I've implemented from scratch).:

image

Now one obvious difference is, the custom UI looks better. It's clear and readable. However it doesn't support anything related to quickfix, that's why I've added also some custom mappings to it (such as opening definition or closing the window). Now I agree this is not sufficient, because this leads us to implementing a custom quickfix with time which I don't want to do.

So with your PR I'm slowly rethinking if we should ditch json parsing and move on to parse the initial GNU style plain format. For this to be happen I need to change the current `Implements` and `Callees` modes to use quickfix. If you can wait, let me fix those and revert back to use quickfix. It needs some refactor (but nothing hard to do). Once I'm finished I'll let you know, and btw, probably we are not going to need to parse the json, as oracle also outputs it formatted.

akavel:
Oh wow, I sure didn't expect such turn of events :) Now that you show them compared, I agree your UI looks prettier and cleaner in this case. That said, I'm happy you're seriously considering the switch.

As to waiting... Fatih, given that I've been on the receiving end of PRs too, and know how it is, I'm totally not going to hurry you up or whatever. I'm very grateful to you for the plugin as it is already, and extra so that it is written in such a clean way, that even a green vim apprentice like me can add something he needs to it. On top of that I'd sure be happy if I could contribute something back and for others to be useful (and also with this sneakily drop the burden of future maintenance of this contribution on you), but I accept that this would be an extra bonus, and if and when it happens is fully yours to choose.

Also, I'm happy to try doing any further improvements to this PR if needed to fit the core, so I'll leave it now till you let me know you're ready. Tx!

fatih:
@akavel the branch is ready with the PR #319. All commands are implemented please let me know how it works for you. If there is any problem please report it back to the issue #250. I'll close this and thanks for the initial PR (it made me think differently) :)

PR #315, opened 2015-02-14
indent: fix breaking behaviour with #308, fixes #314

### LINKED ISSUE #314: autoindent not working
(See https://github.com/fatih/vim-go/pull/308 for background.)

I'm having an issue with 229130c, which changed `GoIndent()` from using `&shiftwidth` to using `shiftwidth()`. The issue is that `shiftwidth()` may not yet have sufficient penetration into the installed base of vim users. If `shiftwidth()` does not exist, calls to it return nothing rather than the desired/expected value. Using the result to adjust `ind` leads to `ind` not changing as expected.

I'm running a fully up-to-date OS X 10.10, which comes with vim 7.3. I suspect that `shiftwidth()` was introduced with vim 7.4. I note that http://vimdoc.sourceforge.net/htmldoc/options.html#'shiftwidth' does not mention the new function (though http://vimhelp.appspot.com/options.txt.html does).

Checking the history of indent/go.vim, I see that it originally merely accessed `&shiftwidth`. The change in 229130c was to change these to `shiftwidth()` calls. Instead, you might want to use the approach in the version of indent/go.vim that came with go 1.3.3:

```
...
if exists("*GoIndent")                                                                    
  finish                                                                                  
endif                                                                                     

" The shiftwidth() function is relatively new.                                            
" Don't require it to exist.                                                              
if exists('*shiftwidth')                                                                  
  func s:sw()
    return shiftwidth()                                                                   
  endfunc                                                                                 
else
  func s:sw()
    return &shiftwidth                                                                    
  endfunc                                                                                 
endif                                                                                     

function! GoIndent(lnum)
...
```

`GoIndent()` then calls `s:sw()` where it currently calls `shiftwidth()` directly.

As an aside, I confirm that autoindenting works when I install vim 7.4 on my Mac.

fatih:
Not all Vim versions have shiftwidth function available to be used, so
fallback to &sw if it's not available. This is also written in the Help
page of shiftwidth.

### DIFF
--- a/indent/go.vim
+++ b/indent/go.vim
@@ -9,7 +9,7 @@
 " - general line splits (line ends in an operator)
 
 if exists("b:did_indent")
-    finish
+	finish
 endif
 let b:did_indent = 1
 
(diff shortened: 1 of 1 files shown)


keith-rollin:
I've tried it with both vim 7.3 and 7.4, and it seems to work as expected. Thanks.

I do note, however, that just about all lines in indent/go.vim are marked as changed due to what looks to be a switch in the way lines are indented. I don't know if that was intended or not, but it makes reviewing the change a little challenging.

fatih:
@keith-rollin Yes sorry for that change. I'm using a commong indentation for all vim files and adding it to the bottom of the file. However it seems it was missing here and I've just added and reformatted it. The only change though is adding the fallback function. Thanks for the feedback, I really appreciate it !

PR #327, opened 2015-02-20
Fix oracle scope to split on whitespace

ghodss:
`g:go_oracle_scope` is splitting on non-word characters, which includes valid import path characters like `/` and `.`. This fixes it to split only on whitespace.

### DIFF
--- a/autoload/go/oracle.vim
+++ b/autoload/go/oracle.vim
@@ -82,7 +82,11 @@ func! s:RunOracle(mode, selected) range abort
     if exists('g:go_oracle_scope')
         " let the user defines the scope, must be a space separated string,
         " example: 'fmt math net/http'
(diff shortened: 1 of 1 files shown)


fatih:
Hi @ghodss . This does't work for me. I've tested it with this scope:

```viml
let g:go_oracle_scope = "fmt math"
```

And it doesn't split them. It prepends them like a single string in the form of `... implements "fmt math"`.

ghodss:
You're right, I think the escaping was getting in the way. I changed it so it split first and then escaped the packages. I tested this and it works fine for me for zero or more packages, but I'm a viml newb so feel free to comment on the styling. :)

fatih:
Thanks @ghodss , works perfectly now :+1:  :)

PR #330, opened 2015-02-24
Trim needless bracket

mattn:
```
http.Handle()
```
When the cursor is at `n` on `Handle` in normal mode, and type `caw<c-x><c-o>`, we expect

```
http.DefaultMaxHeaderBytes()
```
But currently
```
http.DefaultMaxHeaderBytes(()
```
This patch remove trailing brackets (if it's not needed).

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -135,6 +135,11 @@ function! go#complete#Info()
     endif
 endfunction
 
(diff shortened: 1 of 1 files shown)


fatih:
Hi @mattn  Thanks for the PR :+1: I'm happy to see you contribute to vim-go :)

PR #331, opened 2015-02-24
Fix triming bracket

mattn:
previous change should be fixed. When trigger omnifunc at `E` on `.Encode()`, still leave `(` at trailing.


fatih:
I've actually tested your previous change, but couldn't find anything serious. Thanks for fixing this edge case though :)

PR #338, opened 2015-02-27
oracle: add full content of referring line

akavel:
Implementation proposal for a feature suggested in https://github.com/fatih/vim-go/issues/250#issuecomment-76425041

### DIFF
--- a/autoload/go/oracle.vim
+++ b/autoload/go/oracle.vim
@@ -56,7 +56,7 @@ func! s:qflistSecond(output)
     " We discard line2 and col2 for the first errorformat, because it's not
     " useful and quickfix only has the ability to show one line and column
     " number
-	let &errorformat = "%f:%l.%c-%.%#:\ %m,%f:%l:%c:\ %m"
+	let &errorformat = "%f:%l.%c-%[%^:]%#:\ %m,%f:%l:%c:\ %m"
 
     " create the quickfix list and open it
     cgetexpr split(a:output, "\n")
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @akavel.

autoload/go/oracle.vim
```diff
@@ -56,7 +56,7 @@ func! s:qflistSecond(output)
     " We discard line2 and col2 for the first errorformat, because it's not
     " useful and quickfix only has the ability to show one line and column
     " number
-	let &errorformat = "%f:%l.%c-%.%#:\ %m,%f:%l:%c:\ %m"
+	let &errorformat = "%f:%l.%c-%[%^:]%#:\ %m,%f:%l:%c:\ %m"
```
fatih: What does this change do ?

akavel: In short words: this fixes parsing of lines that contain colon (":") in the "message" ("%m") part.

The final part of the first pattern in the old code assumed that there are no colons (":") in the message ("%m"), by ignoring (with "%.%#:", i.e. `".*:"` when translated to regular regexp) all the stuff till the *very last colon* in the line. 

The change inverts the logic, by assuming that we only want to ignore "line 2 and col2" (as mentioned in the comment), and that they don't have a colon. So, we ignore only all the stuff till the *very first colon after %l.%c*. The pattern when translated to more "regular regexp" would look like:

    "FILE:LINE\.COLUMN-[^:]*: MESSAGE"

Does this make it more clear?

fatih: Got it, makes sense now. I wasn't sure about messages containing `:`, so it was a fix but seems it has some quirks in it. Thanks for the contribution btw, much appreciated.

PR #340, opened 2015-02-28
Snippet to handle error and return

nowk:
Not sure how widely used this would be, but I have found myself doing this a lot in a recent project refactoring their HTTP handlers. So here it goes...

    errr<tab>

Expands to -

    if err != nil {
      _ <- cursor puts you here; Handle error
      return
    }

### DIFF
--- a/gosnippets/UltiSnips/go.snippets
+++ b/gosnippets/UltiSnips/go.snippets
@@ -167,6 +167,14 @@ if err != nil {
 ${0}
 endsnippet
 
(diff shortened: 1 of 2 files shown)


fatih:
Hi @nowk. Thanks for the PR.  Can you  rename it to `errh` ? , we have also `errn` (return directly) and `errt` (handle testing errors).

nowk:
Hey @fatih, renamed.

fatih:
@nowk Thanks :+1:

PR #343, opened 2015-03-03
Return basename from BinPath function

kalekseev:
If basename not found in PATH then restore PATH and return empty string
If basename found just return it. Otherwise we could face situation when BinPath returns something like this `D:\home\go\bin/gocode`. Related problem with vimproc https://github.com/Shougo/vimproc.vim/issues/192

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -131,7 +131,14 @@ function! go#tool#BinPath(binpath)
         return ""
     endif
 
-    return basename
+    let $PATH = old_path
+
+    let sep = '/'
+    if IsWin()
+        let sep = '\'
+    endif
+
+    return go_bin_path . sep . basename
 endfunction
 
 " following two functions are from: https://github.com/mattn/gist-vim 
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -124,17 +124,14 @@ function! go#tool#BinPath(binpath)
     let old_path = $PATH
     let $PATH = $PATH . PathSep() .go_bin_path
 
-    if !executable(binpath) 
+    if !executable(basename)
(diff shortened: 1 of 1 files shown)


fatih:
Your fix is actually changing some of the logic there. Does this works for you:


```viml
    let sep = '/'
    if IsWin() 
        let sep = '\'
    endif

    return go_bin_path . sep . basename
```

Just change this part and try again.

kalekseev:
Your fix doesn't solve the problem, although it looks more like vimproc problem (it doesn't handle right paths with `\`).

Anyway I would say that logic in my patch make sense while old code not.
Let me go through the code of the BinPath.

It accept binpath. From logic of function i assume it can be in one of the following format `gofmt` or `/usr/bin/gofmt`, because after that we take only `basename` from it and trying to find it in PATH. But instead of `basename` we search `binpath` in PATH https://github.com/fatih/vim-go/blob/c45812eeb75599acd1ee521ac721a184ce725793/autoload/go/tool.vim#L127

Also we don't restore PATH if `!executable(binpath) `, even more `if go_bin_path` never true. From eval.txt:

```
Conversion from a String to a Number is done by converting the first digits
to a number.  Hexadecimal "0xf9" and Octal "017" numbers are recognized.  If
the String doesn't start with digits, the result is zero.
...
Note that in the command >
	:if "foo"
"foo" is converted to 0, which means FALSE.  To test for a non-empty string,
use empty(): >
	:if !empty("foo")
```

I think problem with my patch is that I thought you change PATH by design, I've added commit which restore PATH and use your code for sep selection.

fatih:
Exactly. Thanks for adding it. This function is called in a lot of places, so it's in a critical hot path.  Let me test it first, once I feel it's good to go  I'll merge it.

fatih:
Thanks for the PR @kalekseev. I've used it since morning and didn't encounter any problem.

PR #349, opened 2015-03-08
Added HTTP handler snippet

namit:


fatih:
Hi @namit This is to general to be added. I appreciate your contribution, but we can later add more generic and useful snippets if needed.

namit:
I don't know what you mean by "too general to be added" and I see myself wanting to use this just like I want to use a for loop. Anyway, thanks for considering.

PR #355, opened 2015-03-13
Use vimporc shellescape

kalekseev:
I have had the same problem as @alexaandru for completiion with vimproc on windows https://gist.github.com/alexaandru/d116c229c07983f80b03

This patch fixes the problem. Tested with and without vimproc on windows and linux.

Vimproc requires that arguments for its functions [were escaped](https://github.com/Shougo/vimproc.vim/issues/192#issuecomment-76941582)
 `s:gocodeShellescape` doesn't satisfies vimproc requirements for escaping (at least on windows), so I replaced it with `vimproc#shellescape`

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -31,6 +31,9 @@ fu! s:system(str, ...)
 endf
 
 fu! s:gocodeShellescape(arg)
+    if go#vimproc#has_vimproc()
+        return vimproc#shellescape(a:arg)
+    endif
     try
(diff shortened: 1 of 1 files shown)


fatih:
Hi @kalekseev. I'm going to check it out and test as soon as a I have some spare time. Thanks for the contribution!

fatih:
Sorry for the late delay, I'm very busy lately. I've tested it on Mac OS X and works quite perfect. Thanks :+1:

PR #363, opened 2015-03-23
Focused test command (go test -run 'Test...$')

nowk:
Separate test command to run only the test immediate to your cursor.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -128,7 +128,7 @@ function! go#cmd#Test(compile, ...)
     endif
 endfunction
 
-function! go#cmd#TestFocused(...)
+function! go#cmd#TestFunc(...)
     let test = search("func Test", "bcsW")
 
     if test == 0
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -129,20 +129,25 @@ function! go#cmd#Test(compile, ...)
 endfunction
 
 function! go#cmd#TestFocused(...)
-    let line = getline(test)
-    let name = split(split(line, " ")[1], "(")[0]
-    let flag = "-run '" . name . "$'"
     let test = search("func Test", "bcsW")
 
-    let a1 = ""
-    if len(a:000)
-        let a1 = a:1
+    if test == 0
+        echo "vim-go: [test] no test found immediate to cursor"
+    else
+        let line = getline(test)
+        let name = split(split(line, " ")[1], "(")[0]
+        let flag = "-run '" . name . "$'"
 
-        " add extra space
-        let flag = " " . flag 
-    endif
+        let a1 = ""
+        if len(a:000)
+            let a1 = a:1
 
-    call go#cmd#Test(0, a1, flag)
+            " add extra space
+            let flag = " " . flag 
+        endif
+
+        call go#cmd#Test(0, a1, flag)
+    end
 endfunction
 
 function! go#cmd#Coverage(...)
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -129,10 +129,10 @@ function! go#cmd#Test(compile, ...)
 endfunction
 
 function! go#cmd#TestFocused(...)
-    let test = search("func Test", "bcs")
     let line = getline(test)
     let name = split(split(line, " ")[1], "(")[0]
     let flag = "-run '" . name . "$'"
+    let test = search("func Test", "bcsW")
 
     let a1 = ""
     if len(a:000)
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -133,21 +133,22 @@ function! go#cmd#TestFunc(...)
 
     if test == 0
         echo "vim-go: [test] no test found immediate to cursor"
-    else
-        let line = getline(test)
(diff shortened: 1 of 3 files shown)


nowk:
Added a `TestLast` function as a companion to the focused command, since you can only run those in the `_test.go` file itself.

ftplugin/go/commands.vim
```diff
@@ -9,6 +9,8 @@ nnoremap <silent> <Plug>(go-run) :<C-u>call go#cmd#Run(expand('%'))<CR>
 nnoremap <silent> <Plug>(go-build) :<C-u>call go#cmd#Build('')<CR>
 nnoremap <silent> <Plug>(go-install) :<C-u>call go#cmd#Install()<CR>
 nnoremap <silent> <Plug>(go-test) :<C-u>call go#cmd#Test(0, '')<CR>
+nnoremap <silent> <Plug>(go-test-focused) :<C-u>call go#cmd#TestFocused('')<CR>
```
fatih: Let us rename all `Focused` to `Function`. So the internal functions should be renamed too. The command should be `:GoTestFunction`. It's more simpler and obvious what it does.

autoload/go/cmd.vim
```diff
@@ -124,6 +135,35 @@ function! go#cmd#Test(compile, ...)
     endif
 endfunction
 
+function! go#cmd#TestFocused(...)
+    let test = search("func Test", "bcsW")
+
+    if test == 0
+        echo "vim-go: [test] no test found immediate to cursor"
```
fatih: Return here early and don't use else below.

autoload/go/cmd.vim
```diff
@@ -124,6 +135,35 @@ function! go#cmd#Test(compile, ...)
     endif
 endfunction
 
+function! go#cmd#TestFocused(...)
+    let test = search("func Test", "bcsW")
```
fatih: Can you change the flags to `bsnW` ? Don't move the cursor. Also I would be happy if you could add the flag meanings as comment so someone else reading the code can have some idea :)

autoload/go/cmd.vim
```diff
@@ -124,6 +135,35 @@ function! go#cmd#Test(compile, ...)
...
+        endif
+
+        call go#cmd#Test(0, a1, flag)
+    end
+endfunction
+
+" default cache, in case TestLast is called before anything
+let s:go_cmd_test_args_cache = [0, ""]
+
+function! go#cmd#TestLast()
```
fatih: I didn't like the caching functionality right now. May be it's because of the naming. It's good but it's somehow doesn't fit into the current code. In which cases are you using it?  (I'm curious about the usage)

fatih:
Hi @nowk , great work! I made some comments. Beside that please add the necessary commands and mappings to  the doc please: https://github.com/fatih/vim-go/blob/master/doc/vim-go.txt

autoload/go/cmd.vim
nowk: My main though for this was, 

1. `:GoTestFunction`
2. Fix failing code (most likely in another buffer, tab, etc..)
3. `:GoTestLast` (To rerun the test, without jumping back to the test buffer)

However, having this on my own setup for the last week or so, I haven't found myself going back to this too often. Might just be muscle memory of going back to the test or just running `:GoTest`, not sure...

Now that I think about, not terribly good with names, but something like `:GoTestReplay` or `:GoTestRerun` might sound better?

I'll remove it from this PR and open another for this specifically to marinate over it a bit longer.

fatih: Sounds good, thanks :+1: This will keep the the current code also to be simple.

nowk:
Hey @fatih, I pushed up the changes you suggested and removed the `GoTestLast` (will create another PR for that shortly).

I opted to make `:GoTestFunction` to `:GoTestFunc` though, shorter, and since go uses `func` thought it might sound better, let me know what you think.

autoload/go/cmd.vim
```diff
@@ -124,6 +128,37 @@ function! go#cmd#Test(compile, ...)
...
+function! go#cmd#TestFunc(...)
+    " search flags legend (used only)
+    " 'b' search backward instead of forward
+    " 'c' accept a match at the cursor position
+    " 'n' do Not move the cursor
+    " 'W' don't wrap around the end of the file
+    "
+    " for the full list
+    " :help search
+    let test = search("func Test", "bcnW")
```
nowk: Had to use `bcnW`, `s` + `n` is not allowed, but this does keep the cursor in place while still allowing to run the test if immediately on the test line, eg. `func TestSomething...`

fatih:
Thanks a lot. I like `:GoTestFunc` better :+1:  Would you mind adding the docs for command and mapping? If not I can add it myself, no worries.

autoload/go/cmd.vim
fatih: You are right, I've also used `bcnW` (replaces `s` with `n`). However I've wrote it wrong in my own comment. Thanks for changing it.

nowk:
Oh right, forgot about that, I'll do that.

nowk:
Command and Mapping documentation added, let me know if I missed anything.

fatih:
Thanks @nowk

nowk:
No, thank you! vim-go ftw.

PR #369, opened 2015-04-04
Fix :GoImports when g:go_fmt_options is set

arodland:
goimports doesn't accept "go fmt" options, so don't try to append them
to the command line when we're calling goimports.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -90,8 +90,12 @@ function! go#fmt#Format(withGoimport)
         let fmt_command = bin_path
     endif
 
-    " populate the final command with user based fmt options
-    let command = fmt_command . ' ' . g:go_fmt_options
+    if a:withGoimport == 1
+        let command = fmt_command
+    else
+        " populate the final command with user based fmt options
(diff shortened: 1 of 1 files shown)


arodland:
Without this, when I run `:GoImports` I get my program echoed to stdout, and the imports aren't modified. Not sure how that happens but it's clearly wrong.

autoload/go/fmt.vim
```diff
@@ -90,8 +90,12 @@ function! go#fmt#Format(withGoimport)
...
     endif
 
-    " populate the final command with user based fmt options
-    let command = fmt_command . ' ' . g:go_fmt_options
+    if a:withGoimport == 1
+        let command = fmt_command
+    else
+        " populate the final command with user based fmt options
+        let command = fmt_command . ' ' . g:go_fmt_options
+    endif
```
fatih: Can you change it to

 ```
if a:withGoimport != 1
   let command = fmt_command . ' ' . g:go_fmt_options
endif
```

And also add comment to the code why we do this, like you have wrote already in the PR description :)

Issue #366, opened 2015-03-25
Syntax highlighting doesn't know about some operators


cespare:
vim-go recognizes these as operators:

```
+ - * / % += -= == < > != <= >= :=
```

It does not recognize these as operators:

```
& | ^ << >> &^ *= /= %= &^= && || <- ++ -- = ! ...
```

And these two, it recognizes the last two characters as being an operator separate from the first character, which is incorrect because the whole thing is a single operator:

```
<<= >>=
```

I tried fixing this in my fork a while ago:

https://github.com/cespare/vim-go/commit/beaf6eef87d56b8356a921bc1550273714a39c17

My change is too broad, though, and causes too many false positive matches. (For instance today it broke a big `/* */` comment block.)

I'll try giving it another shot at some point but I wanted to note it in the issue tracker in case someone else is motivated to tackle it sooner :)

cespare:
Also if you write

``` go
a*b + c*d
```

(note that this is how gofmt will manage the spaces) then the `*` will not be recognized as operators.

cespare:
Trying again with https://github.com/cespare/vim-go/commit/8e1e1bd99fcaf6b21531b993c29280fb4743e914. If it looks OK after a while I'll send over a PR.

fatih:
Hi @cespare 

Thanks for opening this issue. I think the best way to not lose your changes is to open a PR and letting it sit it here. At least we can see each iteration and your improvements (or other people's). Also a PR is more visible for other others, so they can dive easier.

If not, I can open the PR from your branch myself, but it would be better if it's from you :)

cespare:
Done, see #372.

PR #372, opened 2015-04-04
Highlight all Go operators

cespare:
See #366 for context.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -236,20 +236,20 @@ hi def link     goTodo              Todo
 
 " Operators; 
 if g:go_highlight_operators != 0
-	syn match goOperator /:=/
-	syn match goOperator />=/
-	syn match goOperator /<=/
-	syn match goOperator /==/
-	syn match goOperator /!=/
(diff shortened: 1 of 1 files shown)


cespare:
Hi @fatih, friendly bump :)

This has been working great for me for the past week.

fatih:
Hi @cespare I'm currently going over the issues and PR's :) 

I've tested it and it works quite fine, just you had this one note :+1: 

>My change is too broad, though, and causes too many false positive matches. (For instance today it broke a big /* */ comment block.)

I've double checked this (for example by opening `go/src/flag/flag.go` file and it looks fine and I couldn't see any problems. I think the problem doesn't persist any more, right ?

cespare:
Right. That note was about a previous version of this change that I made a few months ago which had a variety of small issues. I put more time and thought into this iteration and as I've used it it seems to be working well.

fatih:
Thanks @cespare. This is a very awesome contribution :+1:

cespare:
Glad to help. Thanks for the great plugin :)

PR #377, opened 2015-04-09
fix 'No match: `=dir`'

kovetskiy:
Sometimes when I'm doing `:GoImport %lib%`, I see this error:

```Error detected while processing function go#import#SwitchImport..go#tool#Exists..go#tool#ExecuteInDir:
line    7:
E480: No match: `=dir`
```

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -76,10 +76,10 @@ function! go#tool#ExecuteInDir(cmd) abort
     let cd = exists('*haslocaldir') && haslocaldir() ? 'lcd ' : 'cd '
     let dir = getcwd()
     try
-        execute cd.'`=expand("%:p:h")`'
+        execute cd . fnameescape(expand("%:p:h"))
         let out = system(a:cmd)
     finally
-        execute cd.'`=dir`'
+        execute cd . fnameescape(dir)
     endtry
     return out
 endfunction


fatih:
This happens if you remove your current folder. That's why you see it sometimes. Can you please check if that is the case?

For example I have something like this in my vimrc which always changes the directory to a valid one:

```viml
autocmd BufEnter * silent! lcd %:p:h
```

kovetskiy:
No, directory is exists. 
`:echo getcwd()` showing `/tmp`

fatih:
This function is taken from @tpope so I don't want to remove or change unless it's really an error. Just googled and found these issues:

https://github.com/tpope/vim-fugitive/issues/371
https://github.com/tpope/vim-fugitive/issues/535

Is this about your wildignore may be?

kovetskiy:
Yes, `:echo &wildignore` getting me `*/tmp/*,*.so,*.swp,*.zip`.

And now, if I don't want to see autocomplete for */tmp/* directories, you think I shouldn't have possibility to use GoImport features?

kovetskiy:
@fatih I can implement changing wildignore, and restoring to previous value after execution `system(a:cmd)`.

Ping me, if you agreed with me.

tpope:
For the record, this has given me enough grief that nowadays I do `execute cd fnameescape(...)`. Without the `fnameescape()` you can get tripped up on spaces and the like.

fatih:
Hey @tpope thanks for the insight. @kovetskiy mind try it with `fnameescape()` ?

kovetskiy:
@fatih Okay, I added fnameescape().

fatih:
@kovetskiy thanks for the PR :+1:

PR #382, opened 2015-04-11
Remove trailing slash when you :GoImport github.com/user/pkg/

nowk:
Removes the trailing slash when using `:GoImport`, this can happen when you happen to `<tab>` your way to the full path and forget to remove the slash yourself.


fatih:
Thanks @nowk :+1:

PR #383, opened 2015-04-12
Use double quotes for regexp in TestFunc

xdl:
On Windows, I was having false negatives when using GoTestFunc, and it turns out that the regexp for go test -run doesn't match with single quotes, and so will always pass since it can't find the test. Using double quotes fixes this.


fatih:
Tested on Mac OS X and works fine. Thanks @xdl for the PR :+1:

PR #391, opened 2015-04-16
Provide default value for rename input

mbrt:
In the prompt, the default value now is the previous symbol.
You see: vim-go: rename 'xxxx' to: xxx
and then you can edit the symbol by starting from the previous name. Many code refactor tools provides this. It is useful when you want to rename a word to something similar without retyping it.


fatih:
Looks good. Tested on my local environment and works fine. Thanks for the PR :+1:

PR #393, opened 2015-04-20
Improve to detect 'no documentation'

rhysd:
## Problem

When `godoc` can't find any package for corresponding query like below,

    :GoDoc foobar

`:GoDoc` opens the window for 'not found' result like as below.

    2015/04/21 03:58:02 open /usr/local/Cellar/go/1.4.2/libexec/src/foobar: no such file or directory

## Solution

I fixed this problem with checking the content of result.
With this fix, `:GoDoc` command correctly echoes 'not found' result without opening window as
below.

    No documentation found for "foobar".

### DIFF
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -71,6 +71,14 @@ function! s:godocWord(args)
     return [pkg, exported_name]
 endfunction
 
+function! s:godocNotFound(content)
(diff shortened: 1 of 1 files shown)


fatih:
Looks good thanks. Godoc needs some love always :)

rhysd:
Thanks!

PR #396, opened 2015-04-21
Operator highlighting cleanup/fixes

### LINKED ISSUE #394: syntax highlighting on combined operators not working as expected
````*foo```` will highlight the ````*```` as an operator as expected.

````!*foo```` will only highlight the ````!```` operator and not the ````*````.
````(*foo)```` will not highlight the ````*```` operator.

Not sure if this has been discussed. Is it intended behaviour? I'll hopefully have a look at it tomorrow and see if there's any easy fix.

relevant: #372 and 1042109ca30d442db08a5f0988837e9401d4d597

cespare:
* Fix highlighting of X*foo where X is any other syntax match by
  removing unnecessary comment-matching protection
* Remove some unnecessary escaping in character classes

Fixes #394.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -236,13 +236,11 @@ hi def link     goTodo              Todo
 
 " Operators; 
 if g:go_highlight_operators != 0
-	" match single-char operators:          - + % < > ! & | ^ =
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @cespare

PR #400, opened 2015-04-24
fix handling paths in error list. close #399

### LINKED ISSUE #399: Path is doubled in error list when we issue the :GoInstall during package development
Windows 7 64-bit, Gvim version 7.4.699
$GOPATH = c:\go
Steps to reproduce:

:cd c:\go\src
:!md a
:cd a
:new a.go

Write something, with mistake for testing:

package a

func F() int { return 1 

:w|GoInstall

Then in quick fix is:
C:\go\src\a\c:\go\src\a\a.go|5| syntax error: unexpected EOF

mattn:


fatih:
Many thanks  @mattn  :+1:

PR #402, opened 2015-04-27
Use Dispatch when available

alexlafroscia:
This has `vim-go` check for the `:Make` command provided by Tim Pope's [`vim-dispatch`](https://github.com/tpope/vim-dispatch) before using `make`, and if it's available, uses that instead.  This allows the `GoBuild` and `GoRun` commands to be run in a Tmux split instead, which is both asynchronous and a _lot_ faster.

I did a little profiling, and it seems like the performance increase is pretty significant

**`:GoBuild` profiling**

| Time | `make` | `Make` (Dispatch) |
| :-- | :--         | :-- |
| Total | 1.055178 s | 0.173488 s |
| Self | 0.016310 s | 0.008696 s |

If you want to check out the full profiling that I did, you can find the files [here](https://gist.github.com/alexlafroscia/d7d351c6fc9834efebe9)

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -6,17 +6,6 @@ if !exists("g:go_dispatch_enabled")
     let g:go_dispatch_enabled = 0
 endif
 
-" Run either `make` or `Make` depending on the user's configuration
-" `Make` required the vim-dispatch plugin by Tim Pope.
-" https://github.com/tpope/vim-dispatch
-function! s:runMake(...)
-    if g:go_dispatch_enabled && exists(':Make') == 2
-        silent exe 'Make'
-    else
-        silent exe 'make!'
-    endif
-endfunction
-
 function! go#cmd#Run(bang, ...)
     let goFiles = '"' . join(go#tool#Files(), '" "') . '"'
 
@@ -38,7 +27,11 @@ function! go#cmd#Run(bang, ...)
         let &makeprg = "go run " . expand(a:1)
     endif
 
-    call s:runMake()
+    if g:go_dispatch_enabled && exists(':Make') == 2
+        silent! exe 'Make!'
+    else
+        exe 'make!'
+    endif
     if !a:bang
         cwindow
         let errors = getqflist()
@@ -79,7 +72,11 @@ function! go#cmd#Build(bang, ...)
     endif
 
     echon "vim-go: " | echohl Identifier | echon "building ..."| echohl None
-    call s:runMake()
+    if g:go_dispatch_enabled && exists(':Make') == 2
+        silent! exe 'Make'
+    else
+        silent! exe 'make!'
+    endif
     redraw!
     if !a:bang
         cwindow
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -23,7 +23,11 @@ function! go#cmd#Run(bang, ...)
         let &makeprg = "go run " . expand(a:1)
     endif
 
-    exe 'make!'
+    if exists(':Make') == 2
+        silent! exe 'Make'
+    else
+        exe 'make!'
+    endif
     if !a:bang
         cwindow
         let errors = getqflist()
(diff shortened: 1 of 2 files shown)


fatih:
Hi @alexlafroscia. This is nice, thanks for the PR. :+1:  

I would like to have it behind a setting though. In my case, I'm using Macvim and sometimes I switch to iTerm. However I don't want `dispatch` to be invoked when I use MacVim or even in iTerm. But it's already installed. So we should have a global option: `g:go_dispatch_enabled` which is by default `0` (disabled). If someone wants to use it, it should be added by them:

`let g:go_dispatch_enabled = 1`

Makes this sense for you?

I also need to test it, will do it in following days. Thanks again for the PR.

alexlafroscia:
Absolutely! I'll add that in.  This is my first time doing anything with a Vim plugin, so I'll try to look through this codebase or others to see if there's a pattern for that that I can follow.

fatih:
Create a new function, in the same file. It should be a local function (such as `runMake()`). And check the setting there. In the `runMake()` function check it with:

```
if g:go_dispatch_enabled && exists(':Make') == 2
...
endif
```

Once you done I can review it.

alexlafroscia:
Excellent, I'll get that taken care of tonight!

alexlafroscia:
Hey, just to verify, you _don't_ want me to make a new function in the `autoload/go/tools.vim`?

autoload/go/cmd.vim
```diff
@@ -2,6 +2,21 @@ if !exists("g:go_jump_to_error")
...
 endif
 
+if !exists("g:go_dispatch_enabled")
+    let g:go_dispatch_enabled = 0
+endif
+
+" Run either `make` or `Make` depending on the user's configuration
+" `Make` required the vim-dispatch plugin by Tim Pope.
+" https://github.com/tpope/vim-dispatch
+function! s:runMake(...)
```
fatih: Can you make it that it accepts a boolean, such as `isSilent`. If it exists add `silent`, if not don't do it. Currently `:GoRun` is broken, because we do display the output of `:GoRun`

autoload/go/cmd.vim
```diff
@@ -23,7 +38,7 @@ function! go#cmd#Run(bang, ...)
         let &makeprg = "go run " . expand(a:1)
     endif
 
-    exe 'make!'
+    call s:runMake()
```
fatih: Should be called  `s:runMake("")`

autoload/go/cmd.vim
fatih: Or you know what, remove this function all together. We us different options for both :GoRun and :GoBuild. For :GoRun we have `exe make!` but for :GoBuild we have `silent! exe 'make!'`.

doc/vim-go.txt
```diff
@@ -599,6 +599,14 @@ Default is enabled. >
...
   let g:go_def_mapping_enabled = 1
 <
+                                                    *'g:go_dispatch_enabled'*
+
+Use this option to enable/disable the use of Dispatch to execute the
+`:GoGun` and `:GoBuild` commands.  More information about Dispatch is
+available at https://github.com/tpope/vim-dispatch. Default is disabled. >
+
+  let g:go_dispatch_enabled = 0
+<
```
fatih: :+1:

autoload/go/cmd.vim
alexlafroscia: Will go :+1:

alexlafroscia:
I wasn't sure what your preference was, but I had `:GoRun` use `:Make!` (which opens a new Tmux window) and `:GoBuild` use `:Make` (just opens a new pane).  This was based on the Dispatch help information, which recommended using `:Make!` for long-running tasks, like running the a whole test suite.

Speaking of tests, I would love to be able to integrate Dispatch with the `:GoTest` command, but it didn't seem to use the `make`.  Maybe that would be something for a future PR?

fatih:
Build on Go is fast, it should be a foreground build. Thus your `:Make` decision is correct for `:GoBuild`. The essence of :GoRun is to capture the output. For example I've just tried on Mac with iTerm and somehow I wasn't successful to see any output. Were you able to do it? 

However this is good to go. I'm merging this. (it doesn't break current functionality)

For GoTest, yes it doesn't call make. I had a reason for it though, not remember now. I'll plan to look at it in more detail. We can as you said do it in the future with another PR.

Thanks for your contribution :+1:

PR #403, opened 2015-04-28
fix neosnippet support

choplin:
Because NeoSnippetSource affects only current buffer,
snippets file is not loaded after the second time.


fatih:
Thanks for the fix @choplin :+1:

PR #405, opened 2015-04-30
tool: generate path based on the current buffer directory

fatih:


mattn:
Thank you & Sorry about I couldn't take a time to look.

PR #413, opened 2015-05-04
Change highlighting setting according to documenation

mitch000001:
The documentation tells that this setting is disabled by default, but it was not.


fatih:
Thanks @mitch000001. Good catch :+1:

PR #422, opened 2015-05-13
fmt: don't call command twice for successful output

fatih:
Previously we were calling gofmt/goimports twice when the command was
successful. First we call it so we can catch any errors and display
it(in quickfix window), if there is no errors, we called it again,
purely so we could pipe the commands output (which in our case is the
formatted output) directly to the current buffer. This was very fast
for gofmt, but with goimports it was slow.

Now, we don't call the command again if it's successful, instead we
just replace the output of the first command with the current buffer.
This is actually so obvious and I'm thinking why I didn't I this in the
first place. Humans ...

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -95,8 +95,12 @@ function! go#fmt#Format(withGoimport)
 
     " execute our command...
     let out = system(command . " " . l:tmpname)
+    let splitted = split(out, '\n')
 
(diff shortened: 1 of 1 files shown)


fatih:
Just remembered that there was a problem with Vim's own history  when we try to replace it, but again it could be something entire different. I'm going to use this for a couple of days extensively to catch any errors before I merge this PR.

fatih:
This is good to go. No significant problems found yet.

PR #429, opened 2015-05-16
Gopath detection and support for godep. Closes #425

### LINKED ISSUE #425: Godeps/gb aware tools
This is just an idea. vim-go should be able to detect `Godeps`, a gb project, and many other popular options and change the `GOPATH` automatically so tools like `godef`, `gocode`, `oracle` etc.. uses the intended vendorized source codes. I've played around it locally (not pushed yet), but it works quite perfect. I'll push with updates once I'm finished with it. Any feedback is welcome again here.

fatih:
This PR includes new features and improvements which brings better support for multiple `GOPATH`s  and tools that support this (such as `godep`). This also creates a foundation for other new project based building tools, such as `gb`. 

The  changes are : 

* A new `:GoPath` command that displays the current set `GOPATH`. It changes the current `GOPATH` when an argument is passed, i.e: `:GoPath /home/fatih/mypath`
* A new `:GoPathClear` command that clears the previous set `GOPATH` via
`:GoPath` and restores it to the initial` GOPATH` value. the current set
`GOPATH`.
* A new `go_autodetect_gopath` setting. This is enabled by default. When enabled, vim-go automatically modifies the `GOPATH` if a project uses a certain package tool (such as `godep`). For now only `godep` is supported. If a user edits a file and invokes any `vim-go` command, such as `:GoDef`, `:GoBuild`, etc.. the `GOPATH` will include the `Godeps` folder. For example any go-to-definition via `:GoDef` will jump to the source code inside `Godeps`. 
* Changes to vim-go code base to have better support for multiple OS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,17 +21,23 @@ disabled/enabled easily.
 * Compile your package with `:GoBuild` , install it with `:GoInstall`
 * `:GoRun` quickly your current file/files
 * Run `:GoTest` and see any errors in quickfix window
+* Automatic `GOPATH` detection based on the directory structure (i.e: `godep`
+  vendored projects)
+* Change `GOPATH` with `:GoPath`, restore back to original at any time with
+  `:GoPathClear`
 * Create a coverage profile and display annotated source code in browser to see
   which functions are covered with `:GoCoverage`
 * Lint your code with `:GoLint`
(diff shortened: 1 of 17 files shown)


fatih:
This is good to go. I'll merge it.

bketelsen:
you are awesome!

PR #434, opened 2015-05-21
support git url.<base>.insteadOf and custom import paths of binaries

bhcleek:
Fix GoUpdateBinaries to support using the git url.<base>.insteadOf
configuration. `go get` is already fixed on the tip of the go sources, but
GoUpateBinaries with go1.4 fails to update any binary whose remote.origin.url
is not the custom import path.

relates to golang/go#10791


fatih:
Thanks @bhcleek :+1:

PR #436, opened 2015-05-22
fixed godocNotFound to not match 'os' pkg contents

blitzrk:
The no such file string appears in 'os' package's documentation, so it erroneously claims that documentation wasn't found. Forcing it to only match that string if it is found at the end of the content allows `:GoDoc os` to succeed, but `:GoDoc osx`, for example, to still fail.

### DIFF
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -76,7 +76,7 @@ function! s:godocNotFound(content)
         return 1
     endif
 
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @blitzrk :+1:

blitzrk:
Thanks for making such a great plugin!

Issue #423, opened 2015-05-14
Oracle scope dependency


tsenart:
Hi Fatih,

As we bespoke on Twitter, I'm having trouble using scope dependent :Go oracle calls. You mentioned we could set it with `g:go_oracle_scope` and I'm trying to think of other ergonomic ways to do this. Here's a perhaps naive idea:  We could traverse the current working directory up until the repo root and then recursively look for `main` packages. We'd then either run the scope dependent `:Go` oracle calls with all of the found main packages, or, with another configuration, let the user pick which one to use. The user would be also able to set the path (relative or absolute) of a `main` package as an argument to the call.

fatih:
Hi @tsenart 

Thanks for bringing this up. I agree that we should make scope handling smarter to handle common cases. For now I'm doing some changes already, which can be seen here:

https://github.com/fatih/vim-go/blob/master/autoload/go/oracle.vim#L82

I'll look into this into more detail once I have some time (I'm currently have a lot of things to do).

tsenart:
Hi @fatih! Thanks for looking into this and food building vim-go. I wouldn't be the same go programmer without it.

fatih:
I'll close this in favor of #437. I'm not sure traversing and finding any main package would be helpful. But the new `:GoOracleScope` has a very nice autocompletion support (it completes only go import paths) and it's really useful. I think we should use it a little bit and get a good sense about it. So other improvements can be made upon on it.

PR #437, opened 2015-05-22
Oracle improvements

fatih:
1. Add new `:GoOracleScope` with following usages:
  *. `:GoOracleScope` prints the current custom scope
  *. `:GoOracleScope fmt github.com/fatih/color` sets the custom scope from given arguments. It overrides the `g:go_oracle_scope` argument.
  *. `:GoOracleScope ""` clears the scope and unsets it

2. Add package completion to `:GoOracleScope`. So  when we hit `tab` for the following command `:GoOracleScope gith...`  it'll be completed to `:GoOracleScope github.com/...` with possible results to complete the finish. It completes any package in `GOPATH`

3. This PR also includes a fix where the completion would show duplicates paths for completion results, such as `github.com/fatih/color` and `github.com/fatih/color/`. 

These are improvements for #423. 

/cc @tsenart

### DIFF
--- a/autoload/go/oracle.vim
+++ b/autoload/go/oracle.vim
@@ -1,4 +1,3 @@
-" -*- text -*-
 "  oracle.vim -- Vim integration for the Go oracle.
 "
 "  Part of this plugin was taken directly from the oracle repo, however it's
@@ -46,7 +45,7 @@ endfun
 " via regex.
 func! s:qflistSecond(output)
     " backup users errorformat, will be restored once we are finished
-	let old_errorformat = &errorformat
+    let old_errorformat = &errorformat
 
     " match two possible styles of errorformats:
     "
@@ -56,13 +55,13 @@ func! s:qflistSecond(output)
     " We discard line2 and col2 for the first errorformat, because it's not
     " useful and quickfix only has the ability to show one line and column
     " number
-	let &errorformat = "%f:%l.%c-%[%^:]%#:\ %m,%f:%l:%c:\ %m"
+    let &errorformat = "%f:%l.%c-%[%^:]%#:\ %m,%f:%l:%c:\ %m"
 
     " create the quickfix list and open it
     cgetexpr split(a:output, "\n")
     cwindow
 
-	let &errorformat = old_errorformat
+    let &errorformat = old_errorformat
 endfun
 
 func! s:getpos(l, c)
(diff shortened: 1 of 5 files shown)


autoload/go/oracle.vim
```diff
@@ -143,6 +143,25 @@ func! s:RunOracle(mode, selected) range abort
     return out
 endfun
 
+function! go#oracle#Scope(...)
+	if len(a:000)
+        if len(a:000) == 1 && a:1 == '""'
```
tsenart: I never reviewed Vimscript before but I'm wondering about the indentation here...

tsenart:
:+1:

autoload/go/oracle.vim
fatih: You are right, the indentation was broken. Just fixed it :+1:

autoload/go/util.vim
```diff
@@ -36,3 +36,14 @@ function! go#util#IsWin()
...
+
+" StripPath strips the path's last character if it's a path separator.
+" example: '/foo/bar/'  -> '/foo/bar'
+function! go#util#StripPathSep(path)
+	let last_char = strlen(a:path) - 1
+	if a:path[last_char] == go#util#PathSep()
+		return strpart(a:path, 0, last_char)
+	endif
+
+	return a:path
```
tsenart: Don't wanna nitpick but is this the right indentation here?

fatih: Yeah the problem is, some files have the indent definition on the bottom, such as:

```
" vim:ts=4:sw=4:et
```

With that, it automatically tries to adapt. However this missing in both `oracle.vim` and `util.vim` (probably it's missing in some other files too). I'll add them whenever I see it's missing :)

fatih:
@tsenart thanks for the initial feedback. If you hit any problems just open an issue please so we can tackle it.

tsenart:
Thanks for making it happen!

On Monday, May 25, 2015, Fatih Arslan <notifications@github.com> wrote:

> @tsenart <https://github.com/tsenart> thanks for the initial feedback. If
> you hit any problems just open an issue please so we can tackle it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/fatih/vim-go/pull/437#issuecomment-105197998>.
>


-- 
Sent from Gmail Mobile

Issue #440, opened 2015-05-25
Remove go_jump_to_error and use the bang(!) argument


fatih:
I think we should completely remove the variable: `g:go_jump_to_error` and use the `!` identifier. This is how `:make` works. For example if you call it like `:make!` it will not jump to the error, `:make` will jump to the error. This is idiomatic vim usage and fit's better into the vim usage. We have already this support for both `:GoBuild` and `:GoRun`, but not for commands such as `:GoTest`, `:GoInstall`

fatih:
Closed with #441

PR #441, opened 2015-05-25
commands: use bang attribute to jump to errors

fatih:
Previously we were using a custom global variable to control the
behavior of jumpting to an error when we called any command in cmd.vim
(such as :GoInstall, :GoTest, etc..)

However we don't need this actually because the default :make command
in vim has the behavior of no jumping to the error when the `!` attribute
is passed to the command, i.e: `:make!`. We have this feature already
for `:GoBuild` and `:GoRun` but those were also conflicting with our
manual `g:go_jump_to_error` setting. Basically even if you call
`:GoBuild` (without `!`) it would not jump if `g:go_jump_to_error` was
set to `0`.

This PR adds the bang attribute, removes the `go_jump_to_error` setting
for the following commands:

* `:GoBuild`
* `:GoRun`
* `:GoInstall`
* `:GoTest`
* `:GoTestFunc`
* `:GoTestCompile`
* `:GoVet`
* `:GoCoverage`

Now basically the behavior is just like `:make`:

* `:GoBuild` jumps to the error
* `:GoBuild!` does not jump to the error

Relates issue #440

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -35,8 +35,10 @@ function! go#cmd#Build(bang, ...)
 
     cwindow
     let errors = getqflist()
-    if !empty(errors) && !a:bang
-        cc 1 "jump to first error if there is any
+    if !empty(errors) 
+        if !a:bang
+            cc 1 "jump to first error if there is any
+        endif
     else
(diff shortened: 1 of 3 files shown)


fatih:
This is good to go.

PR #450, opened 2015-06-02
when open existing go files, fencs become only utf-8.

mattn:


mattn:
This is part that I wrote at the first. So this was my bug. :-(

fatih:
@mattn thanks! I don't have much time recently, going to check and merge it as soon as possible.

mattn:
no prob!

fatih:
@mattn I've tried it now however I'm confused about one thing. For example if we open a new file utf-8 is applied, so `s:current_fileformats` is nonempty now.

If I open another second file, it never changes the encoding, because `s:current_fileformats` is non empty. Example ('ve put echo statements):

```vim
function! s:gofiletype_pre()
    echo "pre"
    if s:current_fileformats == ''
        echo "inside pre"
        let s:current_fileformats = &g:fileformats
        let s:current_fileencodings = &g:fileencodings
        set fileencodings=utf-8 fileformats=unix
        setlocal filetype=go
    endif
endfunction
```

Here I see `pre` and `inside pre` for the first file. However every other file only prints `pre`. Is this normal? This means other files encoding never change to `utf-8` anymore. I'm not sure what this fixex, would be happy if you give more detail.

mattn:
Sorry, I don't still figure out what probrem is. I'll open new PR in later.

PR #451, opened 2015-06-03
set tabstop=8 in gofiletype_pre function

rocknrollMarc:
As my ADHD made me actualy count the spaces on numerous examples, I figured that "fmt" is allways in sync with main in package main and thats 8 spaces :) just feels right like that.


fatih:
@rocknrollMarc I'm not sure I want to set it. Gofmt is already changing it that way. And people can decide what to use in vim by themself, for example I have this in my vimrc:

```
au BufNewFile,BufRead *.go setlocal noet ts=4 sw=4 sts=4
```

I always see four spaces, not 8. I don't want to touch people's setting so thanks for the PR.

rocknrollMarc:
Ah cool yes Ill try that :) good idea

PR #452, opened 2015-06-04
Don't break when file path contains space

eliasgs:
When using :GoDef on files in a directory containing space, vim would break
with something like: `E484: Can't open file /var/folders/...`

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -21,7 +21,7 @@ function! go#def#Jump(...)
 	let old_gopath = $GOPATH
 	let $GOPATH = go#path#Detect()
 
-	let command = bin_path . " -f=\"" . expand("%:p") . "\" -i " . shellescape(arg)
+	let command = bin_path . " -f=" . shellescape(expand("%:p")) . " -i " . shellescape(arg)
 
 	" get output of godef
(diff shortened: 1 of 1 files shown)


fatih:
Hi @eliasgs. Thanks for pr. I'm currently a little busy. I'm going to look at this and test it once I have some spare time. In the meanwhile could you give me a detail how to reproduce it? As I know it's not valid to have spaces in a golang import path?

eliasgs:
Hi @fatih.  The reason I ran into this issue, is that my current GOPATH is a symbolic link to a Dropbox folder which for some reason includes spaces. I guess `expand("%:p")` just gives the full path.

So to reproduce, you can create a symbolic link to a directory containing spaces and use the link as your GOPATH or simply `export GOPATH="/some/path/with spaces"`. Then running `:GoDef` on a file in your GOPATH, should give you the `E484` error.

Thanks:)

fatih:
@eliasgs I think it would be better if you use `shellescape`. Could you try this out:  https://github.com/fatih/vim-go/pull/454

If this works, just change your PR and I'll merge yours ;)

eliasgs:
I agree:) It works fine, thanks!

fatih:
Thanks @eliasgs :+1:

PR #455, opened 2015-06-07
goFormatSpecifier width and precision match

Intyre:
Fix to match and highlight width and precision in goFormatSpecifier.
```
%s %5s %-5s %5.5f %.5f
```
Using a part of the [Python syntax](https://github.com/hdima/python-syntax/blob/44f1855902b6e2836ff95bec3fc221ad17021ee3/syntax/python.vim#L306).


fatih:
Thanks @Intyre, awesome contribution :+1: :)

PR #463, opened 2015-06-20
Basic implementation of :GoGenerate command

andviro:
Hi, Fatih! First of all, great big thanks for your work on this excellent plugin. I have implemented simple go-generate support, by basically taking snippets of your code for go-build command. Hope that it will be useful.

Andrew.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -242,6 +242,42 @@ function! go#cmd#Vet(bang)
         redraw | echon "vim-go: " | echohl Function | echon "[vet] PASS" | echohl None
     endif
 endfunction
+"
+" Generate runs 'go generate' in similar fashion to go#cmd#Build()
+function! go#cmd#Generate(bang, ...)
+    let default_makeprg = &makeprg
+    let gofiles = join(go#tool#Files(), '" "')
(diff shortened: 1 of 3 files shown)


fatih:
Hi @andviro, thanks for the new feature! I'm going to look at and review as soon as possible.

autoload/go/cmd.vim
```diff
@@ -242,6 +242,42 @@ function! go#cmd#Vet(bang)
...
+    echon "vim-go: " | echohl Identifier | echon "generating ..."| echohl None
+    if g:go_dispatch_enabled && exists(':Make') == 2
+        silent! exe 'Make'
+    else
+        silent! exe 'make!'
+    endif
+    redraw!
+
+    cwindow
+    let errors = getqflist()
```
fatih: I'm not sure we need parsing quickfix list as `go generate` is only a wrapper. However the tool itself might output an error so I'm ok with it.

fatih:
@andviro I've tested it and it works quite well. Thanks a lot contributing to vim-go :+1: I know there is a duplication with build, run, generate, etc.. commands. I'm eventually going to refactor them in near future.

PR #464, opened 2015-06-20
fix autocomplete on windows

jittuu:
Hi, Fatih! First of all, great big thanks for your work on this excellent plugin.

I'm trying to use `vim-go` on my **Windows** PC. It is working fine for other features except gocode autocomplete. I always got this error when I try to use autocomplete.

`Omni completion (^O^N^P) Pattern not found`

With the help of Google, I also try to print vim debug log and I got the following:

```
continuing in function go#complete#Complete..<SNR>53_gocodeAutocomplete..<SNR>53_gocodeCommand
Calling shell to execute: ""gocode" "-in=C:\Users\soemoe\AppData\Local\Temp\VIA4E9E.tmp" "-f=vim" "autocomplete" "C:\Users\soemoe\code\go\src\github.com\jittuu\main.go" "76" >C:\Users\soemoe\AppData\Local\Temp\VIo4EBE.tmp 2>&1"-- Omni completion (^O^N^P) -- Searching...-- Omni completion (^O^N^P) Pattern not found-- INSERT ---- INSERT --
```

After I made the changes (removed shellescape for gocode) as in this pull request, it is working fine in my Windows PC.

Please let me know if there is any impact for this changes as it is my first ever vimscript code. :smiling_imp:

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -61,7 +61,7 @@ fu! s:gocodeCommand(cmd, preargs, args)
     let old_gopath = $GOPATH
     let $GOPATH = go#path#Detect()
 
-    let result = s:system(printf('%s %s %s %s', s:gocodeShellescape(bin_path), join(a:preargs), s:gocodeShellescape(a:cmd), join(a:args)))
+    let result = s:system(printf('%s %s %s %s', bin_path, join(a:preargs), s:gocodeShellescape(a:cmd), join(a:args)))
 
     let $GOPATH = old_gopath
 


fatih:
Hm. I don't know if this really solves you problem. The `pattern not found` error comes only if you didn't compiled your binary. 

What exactly did you solved by removing the shell escape function?

jittuu:
hmm.. it is working fine now (even without compiling because it is inside the same package).

All I did was, I upgrade gvim to 7.4 while fixing editing symbolic link issue in windows. I'm not sure it also fixed it.

I'll try to reproduce if I can find my old gvim installer and will update here again.

Again, thank you for the response and great plugin.

PR #471, opened 2015-07-01
Add gb-vendor plugin's folder to gopath

euank:
This adds a 'vendor' folder under the discovered 'gb' gopath to your gopath. It does not generalize to vendor folders in your existing gopath.

Per the TODO a few lines above, it might make sense to turn this into a configurable option (e.g. an array of 'vendor', 'third_party' searched for as subdirectories of each part of your gopath), but that change is a little more involved and this change does solve a problem for me and, perhaps, others.


fatih:
Hi @euank. Thanks for this simple change. I was going to change this function, however this can be merged without any problem. Once I refactor, I'll just change it too. Thanks!

Issue #412, opened 2015-05-03
Add :GoInstallDeps to install missing third party dependencies


fatih:
Useful if you build something and you hit the error `cannot find package ...`. We'll going to call :GoInstallDeps which will fetch missing dependencies. We already have knowledge about dependencies via `:GoDeps` and `:GoFiles`, should be trivial to implement.

avrtno:
It's very simple to implement:

    get get path/to/package

You can fetch dependencies in a loop until all are resolved using :GoDeps, but it's simpler to use 'go get' which does this already.

fatih:
We can do it already with `:GoImport!`, closing this as this is not something top priority.

PR #473, opened 2015-07-01
:GoImport! to import with go get

mattn:
Sometimes I want to add import with go-get.

### DIFF
--- a/autoload/go/import.vim
+++ b/autoload/go/import.vim
@@ -39,7 +39,7 @@
 " The backslash '\' is the default maplocalleader, so it is possible that
 " your vim is set to use a different character (:help maplocalleader).
 "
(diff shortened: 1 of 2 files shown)


fatih:
Hi @mattn. Thanks for the PR. I've added long time ago an issue about this: https://github.com/fatih/vim-go/issues/412 I had something different in mid, but this is imho more suitable, it's backward compatible and it suits in the flow. Can you please modify the doc too and add the necessary new information to it? https://github.com/fatih/vim-go/blob/master/doc/vim-go.txt#L164

fatih:
Ping @mattn. If you are busy let me know, I'll do it, no worries.

mattn:
do it please.

mattn:
thank you

fatih:
Thank you to @mattn :+1:

PR #476, opened 2015-07-04
Use delete plus append to replace file on GoFmt

### LINKED ISSUE #459: go_fmt_autosave is slow
" delete any leftover before we replace the whole file. Suppose the
                                    " file had 20 lines, but new output has 10 lines, only 1-10 are
                                    " replaced with setline, remaining lines 11-20 won't get touched. So
                                    " remove them.
    1              0.000004         if line('$') > len(splitted)
                                        execute len(splitted) .',$delete'
                                    endif
                            
                                    " setline iterates over the list and replaces each line
    1             13.487943         call setline(1, splitted)

I let g:go_fmt_autosave = 1 in my vimrc, then the save is very slow.
run  `!goimports -w=true %`  in vim is fast.  if you can use this instead of setline? or it's only happened on my computer

luan:
This should fix #459

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -118,16 +118,9 @@ function! go#fmt#Format(withGoimport)
         let default_srr = &srr
         set srr=>%s 
 
-        " delete any leftover before we replace the whole file. Suppose the
-        " file had 20 lines, but new output has 10 lines, only 1-10 are
-        " replaced with setline, remaining lines 11-20 won't get touched. So
-        " remove them.
-        if line('$') > len(splitted)
-            execute len(splitted) .',$delete'
-        endif
-
-        " setline iterates over the list and replaces each line
-        call setline(1, splitted)
+        " replace the file with the new contents
+        execute '%delete'
+        call append(0, splitted)
 
         " only clear quickfix if it was previously set, this prevents closing
         " other quickfixes


fatih:
Hi @luan. Thanks for the PR. The `fmt.vim` file is an important part of vim-go which is changed a lot and I try to not introduce any errors. Your change particular is something which changed later with: https://github.com/fatih/vim-go/commit/99a1732e40e3f064300d544eebd4153dbc3c60c7 

So we need to be sure it works it doesn't trigger the window change event. This causes really a pain when you have two windows with the same buffer. I'm looking now into it, if I something that works well we can use it.

fatih:
This unfortunately breaks vim when you use two views with the same file. The problem is delete which causes this. I can't merge this. I've tried to reimplement it with `append` but it requires to use `delete` on all file content, which again makes it even worse. I'm closing this. Thanks.

luan:
Humm, I understand that this breaks when you use two views with the same file, but in my humble opinion that is more acceptable than waiting a few seconds every time you save. In any case, we need to find a way around this, I know a lot of people that end up switching to https://github.com/google/vim-codefmt because of the performance issues of your fmt code.
I'll be using my fork of your plugin for now, and will try to think of a way around this. Just to re-iterate, waiting a few seconds every time you save a file is not acceptable, more so than breaking when you use two views of the same file (which I myself consider somewhat of an edge case).

fatih:
Did you checked #459?  I don't know what `couple of seconds` means I've profiled it and also provided a way so people can post their profile logs, so I can look. Otherwise a couple of seconds doesn't mean much. (for me for example it's not like that)

I'm aware of codefmt, however the only complaining about fmt is currently the one in #459. The fmt code is already improved and I'm constantly improving it. However, `append` is really not acceptable as it mess up the layout. This shouldn't be happen. You might look in a solution where it doesn't mess it up, however if you except it me to merge it like this I can't sorry. 

I'm open to contributions that will improve things without breaking other things. Otherwise I'm going to be the one that is maintaining all this stuff and I know that in one week I'll get another issue that vim-go breaks the layout. Vim-go is already huge enough and this all is done on my free time. So anything new that makes it worse is not acceptable.

For you speed isse, please profile and post your log. So we can check what's wrong.

luan:
Sorry if I came out rude, I didn't mean to seem pushy. The problem I have is exactly the same as #459.

Here's the profile for that:
```

FUNCTION  go#fmt#Format()
Called 1 time
Total time:   7.787686
 Self time:   7.546543

count  total (s)   self (s)
                                " save cursor position and many other things
    1              0.000004     let l:curw=winsaveview()
                            
                                " needed for testing if gofmt fails or not
    1              0.000004     let l:tmpname=tempname()
    1              0.001531     call writefile(getline(1,'$'), l:tmpname)
                            
                            
    1              0.000005     if g:go_fmt_experimental == 1
                                    " save our undo file to be restored after we are done. This is needed to
                                    " prevent an additional undo jump due to BufWritePre auto command and also
                                    " restore 'redo' history because it's getting being destroyed every
                                    " BufWritePre
                                    let tmpundofile=tempname()
                                    exe 'wundo! ' . tmpundofile
                                endif
                            
                                " get the command first so we can test it
    1              0.000004     let fmt_command = g:go_fmt_command
    1              0.000002     if a:withGoimport  == 1 
                                    let fmt_command  = g:go_goimports_bin
                                endif
                            
                                " if it's something else than gofmt, we need to check the existing of that
                                " binary. For example if it's goimports, let us check if it's installed,
                                " if not the user get's a warning via go#path#CheckBinPath()
    1              0.000003     if fmt_command != "gofmt"
                                    " check if the user has installed goimports
    1   0.000077   0.000018         let bin_path = go#path#CheckBinPath(fmt_command) 
    1              0.000003         if empty(bin_path) 
                                        return 
                                    endif
                            
                                    " change GOPATH too, so goimports can pick up the correct library
    1              0.000003         let old_gopath = $GOPATH
    1   0.001100   0.000012         let $GOPATH = go#path#Detect()
                            
    1              0.000002         let fmt_command = bin_path
    1              0.000001     endif
                            
                                " populate the final command with user based fmt options
    1              0.000002     let command = fmt_command . ' ' . g:go_fmt_options
                            
                                " execute our command...
    1   0.240258   0.000262     let out = system(command . " " . l:tmpname)
    1              0.000273     let splitted = split(out, '\n')
                            
    1              0.000005     if fmt_command != "gofmt"
    1              0.000016         let $GOPATH = old_gopath
    1              0.000001     endif
                            
                            
                                "if there is no error on the temp file replace the output with the current
                                "file (if this fails, we can always check the outputs first line with:
                                "splitted =~ 'package \w\+')
    1              0.000001     if v:shell_error == 0
                                    " remove undo point caused via BufWritePre
    1              0.000015         try | silent undojoin | catch | endtry
                            
                                    " do not include stderr to the buffer, this is due to goimports/gofmt
                                    " tha fails with a zero exit return value (sad yeah).
    1              0.000004         let default_srr = &srr
    1              0.000015         set srr=>%s 
                            
                                    " delete any leftover before we replace the whole file. Suppose the
                                    " file had 20 lines, but new output has 10 lines, only 1-10 are
                                    " replaced with setline, remaining lines 11-20 won't get touched. So
                                    " remove them.
    1              0.000004         if line('$') > len(splitted)
                                        execute len(splitted) .',$delete'
                                    endif
                            
                                    " setline iterates over the list and replaces each line
    1              7.543638         call setline(1, splitted)
                            
                                    " only clear quickfix if it was previously set, this prevents closing
                                    " other quickfixes
    1              0.000005         if s:got_fmt_error 
                                        let s:got_fmt_error = 0
                                        call setqflist([])
                                        cwindow
                                    endif
                            
                                    " put back the users srr setting
    1              0.000006         let &srr = default_srr
    1              0.000001     elseif g:go_fmt_fail_silently == 0 
                                    "otherwise get the errors and put them to quickfix window
                                    let errors = []
                                    for line in splitted
                                        let tokens = matchlist(line, '^\(.\{-}\):\(\d\+\):\(\d\+\)\s*\(.*\)')
                                        if !empty(tokens)
                                            call add(errors, {"filename": @%,"lnum":     tokens[2],"col":      tokens[3],"text":     tokens[4]})
                                        endif
                                    endfor
                                    if empty(errors)
                                        % | " Couldn't detect gofmt error format, output errors
                                    endif
                                    if !empty(errors)
                                        call setqflist(errors, 'r')
                                        echohl Error | echomsg "Gofmt returned error" | echohl None
                                    endif
                                    let s:got_fmt_error = 1
                                    cwindow
                                endif
                            
    1              0.000001     if g:go_fmt_experimental == 1
                                    " restore our undo history
                                    silent! exe 'rundo ' . tmpundofile
                                    call delete(tmpundofile)
                                endif
                            
                                " restore our cursor/windows positions
    1              0.000598     call delete(l:tmpname)
    1              0.000015     call winrestview(l:curw)
```

For a more realistic example this is the file I'm using: https://github.com/cloudfoundry-incubator/rep/blob/master/cmd/rep/main_test.go

My initial point was that I believe that waiting 7s every time you save such file is worse than having the layout breakdown in certain corner cases. But my proposal was that we come up with a solution that fits both cases. I don't think me and #459 are the only ones having the problem, I just think that people might not want to keep opening the same issue after you say that you'd be looking into it.

For more details this is the vim version I'm using currently:

```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Jun 22 2015 15:27:21)
MacOS X (unix) version
Included patches: 1-488
Compiled by Homebrew
Huge version without GUI.  Features included (+) or not (-):
+acl             +farsi           +mouse_netterm   +syntax
+arabic          +file_in_path    +mouse_sgr       +tag_binary
+autocmd         +find_in_path    -mouse_sysmouse  +tag_old_static
-balloon_eval    +float           +mouse_urxvt     -tag_any_white
-browse          +folding         +mouse_xterm     -tcl
++builtin_terms  -footer          +multi_byte      +terminfo
+byte_offset     +fork()          +multi_lang      +termresponse
+cindent         -gettext         -mzscheme        +textobjects
-clientserver    -hangul_input    +netbeans_intg   +title
+clipboard       +iconv           +path_extra      -toolbar
+cmdline_compl   +insert_expand   +perl            +user_commands
+cmdline_hist    +jumplist        +persistent_undo +vertsplit
+cmdline_info    +keymap          +postscript      +virtualedit
+comments        +langmap         +printer         +visual
+conceal         +libcall         +profile         +visualextra
+cryptv          +linebreak       +python          +viminfo
+cscope          +lispindent      -python3         +vreplace
+cursorbind      +listcmds        +quickfix        +wildignore
+cursorshape     +localmap        +reltime         +wildmenu
+dialog_con      +lua             +rightleft       +windows
+diff            +menu            +ruby            +writebackup
+digraphs        +mksession       +scrollbind      -X11
-dnd             +modify_fname    +signs           -xfontset
-ebcdic          +mouse           +smartindent     -xim
+emacs_tags      -mouseshape      -sniff           -xsmp
+eval            +mouse_dec       +startuptime     -xterm_clipboard
+ex_extra        -mouse_gpm       +statusline      -xterm_save
+extra_search    -mouse_jsbterm   -sun_workshop    -xpm
   system vimrc file: "$VIM/vimrc"
     user vimrc file: "$HOME/.vimrc"
 2nd user vimrc file: "~/.vim/vimrc"
      user exrc file: "$HOME/.exrc"
  fall-back for $VIM: "/usr/local/share/vim"
Compilation: /usr/bin/clang -c -I. -Iproto -DHAVE_CONFIG_H   -F/usr/local/Frameworks -DMACOS_X_UNIX  -Os -w -pipe -march=native -mmacosx-version-min=10.10 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1
Linking: /usr/bin/clang   -L. -fstack-protector -L/usr/local/opt/openssl/lib -L/usr/local/opt/readline/lib -L/usr/local/opt/libyaml/lib -L/usr/local/opt/gdbm/lib  -L/usr/local/lib -F/usr/local/Frameworks -Wl,-headerpad_max_install_names -o vim        -lm  -lncurses -liconv -framework Cocoa  -L/usr/local/lib -llua -fstack-protector  -L/System/Library/Perl/5.18/darwin-thread-multi-2level/CORE -lperl -framework Python   -lruby-static -framework CoreFoundation -lgmp -lobjc -L/Users/pivotal/.rubies/ruby-2.1.6/lib
```

I'm trying with latest patch now, I'll comment again if anything changes.

luan:
Update: no change on latest patch

PR #484, opened 2015-07-12
Correct show import section in tagbar plugin.

ithamsteri:
Before: Screenshot 1
After: Screenshot 2


fatih:
Thanks @ithamsteri. Awesome fix :+1:

PR #488, opened 2015-07-18
Open *.slide files in browser automatically

marcosnils:
The following PR allows *.slide files to be opened into the browser
through the "GoPresent" command. In addition go_slide_start_server can
be set to start "present" automatically.

@fatih this the first thing I've ever done for `vim`. Even though I've based into your current vim-go please take a careful look into it. 

Hopefully it's useful enough to merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ disabled/enabled easily.
   your custom vim function.
 * Tagbar support to show tags of the source code in a sidebar with `gotags`
 * Custom vim text objects, such a `a function` or `inner function`
+* Automatically open `present` slides with `:GoPresent`
 
 ## Install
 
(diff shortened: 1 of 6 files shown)


fatih:
Hi @marcosnils 

Thanks for the PR. I'm not sure we want something like this to be included into vim-go. Especially where you have no control of running process, like the present one. What happens if you call `:GoPresent twice? What happens if `present` is not available?  Also the extension `slide` is not go specific and can be used by others. I'm not keen to add this. I know you put some work here, and don't want to see cruel here but I'm always in favor of simplicity. I hope you understand me. 

Maybe we can add something which improves `slide` editing. For example we might add a feature to automatically run/play a selected `.play` line. Once the cursor is on that line, we just run `:GoPresentPlay`, which loads the selected file into a scratch buffer and runs and shows the output (like `:GoRun`).  Let me know what you think of it.

Thanks.

marcosnils:
@fatih I realized about the `present` process as well that's the reason why I made it optional in the first place. 
The motivation for this modification arose when I was making a presentation and I found myself closing and re-opening them several times. I understand your reasons for not merging and they seem completely reasonable, I'll keep using it nevertheless 

Keep up the excellent work! :smile:

PR #489, opened 2015-07-21
Ensure $GOPATH in current shell is passed to tools.

### LINKED ISSUE #472: Go tool commands does not respect $GOPATH in current shell.
As the title says. It seems like the tool only gets the `$GOPATH` value from the default shell environment and not the current one VIM is running in.

After running `:GoImplements` on an indentifier

```
vim-go: oracle: directory /Users/chakrit/Documents/example/src/db is not beneath any of these GOROOT/GOPATH directories: /Users/chakrit/Documents/go/src, /Users/chakrit/Documents/gopath/src
```

However running `:GoPath` prints the right value from the current shell (project's root)

```
/Users/chakrit/Documents/example/lib/:/Users/chakrit/Documents/example/
```

chakrit:
The `system` call spawns a new subshell. If you set your `$GOPATH` in your `.zshenv` (or bash equivalent) that value will be used instead of what you set in your current shell (and VIM's) breaking `gorename` and `go oracle` and is counterintuitive from the VIM user's POV. This changes ensure `$GOPATH` picked up by the tools are consistent with the shell that VIM started in.

Fix #472

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -93,7 +93,7 @@ function! go#tool#ExecuteInDir(cmd) abort
     let dir = getcwd()
     try
         execute cd . fnameescape(expand("%:p:h"))
-        let out = system(a:cmd)
+        let out = system("env GOPATH=" . $GOPATH . " " . a:cmd)
     finally
         execute cd . fnameescape(dir)
     endtry


fatih:
Hi @chakrit. Thanks for the fix. Is this Windows compatible? It seems not, but its outside of my knowledge so I'm not sure.

chakrit:
@fatih i'll test on windows this weekend and lets you know.

fatih:
Sounds good :+1:

chakrit:
@fatih tested and working. i think `vim` on windows already have access to some unix tools so `env` works.

chakrit:
For the record, tools are installed using [Chocolatey](https://chocolatey.org) with the command:

```sh
$ choco install vim git golang python
```

And the plug in was installed using [vim-plug](https://github.com/junegunn/vim-plug) since it's easy to install on windows. (just a 1-file drop-in)

fatih:
Thanks @chakrit. I'm still looking at it, however it seems `system()` is already picking up the environment variables if those are set: http://vim.1045645.n5.nabble.com/Set-environment-variables-in-Vim-td1177621.html I'm not sure if there are any edge cases though.

chakrit:
@fatih Hmm, maybe this is specific to my shell? I'm running `zsh` inside `tmux` : /

fatih:
@chakrit can you just try to use vim-go with bash? I don't see any problems here otherwise :/

fatih:
@chakrit I'm closing this as the GOPATH should be passed to VIM from outside. Please check your environment variable and how they are sourced when you open an app to see what fails.

chakrit:
> I'm closing this as the GOPATH should be passed to VIM from outside.

This isn't the point. GOPATH is already passed to VIM from outside, it's just not further passed to subshells spawned by VIM and the PR is making sure here that it does.

chakrit:
And I don't think zsh isn't some obscure shells that shouldn't be supported.

fatih:
I was saying that this is not possible because vim-go is working like that for 1.5 years already and apps like godef would just break because they all rely on GOPATH, which is indeed passed down to the process.

chakrit:
@fatih I don't get it. How does this break godef? Just passing the same GOPATH value breaks it?

fatih:
> And I don't think zsh isn't some obscure shell that shouldn't be supported.

I'm using ZSH too and it works. My point is that this is not a fix, because otherwise the setup of several ten thousands plugin users (including me) would just not work from the beginning. I'm not agains to fix your problem, but this imho is no the real fix in any case. 

>  I don't get it. How does this break godef? Just passing the same GOPATH value breaks it?

I'm saying that godef works already in the current context without the fix. If what you said would be true (which is not as I'm trying to explain above), it would break all tools from the very beginning of this project.  I'm against these kind of hacks, as `env` would possible only work on POSIX  compatible OS. For example I'm suspicious that this even would work on Windows (probably it'll break windows support). 

The problem you encounter is a different one, I don't know  what the solution would be right now, but we need to find the real problem first before we go and change it.

chakrit:
Ok, that clears it up better.

The problem I have is that everything works just fine outside of VIM. All tools work. So not sure what to check further.

fatih:
Can you change your shell explicitly? Just add this to your vimrc:

```
set shell='/bin/sh'
```

I'm curious if somehow zsh is changing it. I'm currently in #vim channel at freenode IRC and this one of the suggestions. We also had a issue with this a while back ago, so I've wrote this to our FAQ: https://github.com/fatih/vim-go/wiki/FAQ-Troubleshooting#im-using-fish-shell-but-have-some-problems-using-vim-go

guns:
@chakrit: .zshenv is sourced by zsh on every invocation of the shell, not just interactive sessions. The only bash equivalent of this I am aware of is setting the BASH_ENV variable.

Are you sure you want this? Setting your GOPATH in your .zshprofile or .zshrc will affect your interactive sessions without polluting your non-interactive ones.

If you set GOPATH in .zshenv, you are asking zsh to override any inherited value of GOPATH, so you shouldn't be surprised when it does so.

fatih:
@guns you are right indeed, I've also got a similar response from our #vim channel at Freenode: https://github.com/fatih/vim-go/issues/472#issuecomment-140359028

chakrit:
@fatih @guns Ah, so that was the problem. Thanks for getting down to it! GoRename works now but oddly GoDef no longer works w/ vim-go@master but I think that might be a different problem. I'll report back if there's anything else.

Do you guys want to add this to the FAQ?

chakrit:
I recalled it being easier to put in `.zshenv` because some tools (with their shell handling) would not pick it up correctly. That's why I put it there but I have since forgotten why.

fatih:
@chakrit would be happy if you could add it to the WIKI/FAQ, also let us close #472

PR #492, opened 2015-07-23
Improve fmt-in-write by storing results in a tempfile

guns:
The current implementation of `gofmt -w` is incomplete as it cannot be
used from a BufWritePre hook. The command is run on a buffer's backing
file, but if the user's buffer has not been flushed to disk, gofmt
processes the existing file and we discard the user's changes.

We can avoid this situation by using a tempfile and moving the file into
place after a successful fmt.

Also note that `let &syntax = &syntax` is the preferred way of resetting
the syntax engine for a particular buffer.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -57,6 +57,10 @@ function! go#fmt#Format(withGoimport)
     " save cursor position and many other things
     let l:curw=winsaveview()
 
+    " Write current unsaved buffer to a temp file
+    let l:tmpname = tempname()
+    call writefile(getline(1, '$'), l:tmpname)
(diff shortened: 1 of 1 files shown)


fatih:
Hi @guns 

Many thanks! This improvement really rocks right now :) It doesn't suffer from buffer changes at all and is much more better than my current implementation. I've never thought of using a temporary file, nice trick there :+1:  I'll definitely going to merge this one, just let me use it for a couple of days and see if I hit any problems. Other than this, I don't see any problems.

fatih:
This is good to go. I've used it without any problems. Thanks again @guns :+1:

Issue #461, opened 2015-06-16
GoLint support for entire package?


ivulfson:
Looks like GoLint only supports running on the current file.  Would it be possible to add support for running golint on the entire package, when editing one of the package files?

fatih:
Hi @ivulfson.

`golint` supports packages: https://github.com/golang/lint#usage so yes it's possible. Thanks for bringing up the idea. I'll mark this as an improvement and I'm going to look at it once I have proper time.

fatih:
This is now supported with the merged #493 PR. @ivulfson Please test it and report us any problems you have :)

ivulfson:
Confirmed working, after I changed my .vimrc thusly:

```
-au FileType go nmap <leader>l  :GoLint<CR>
+au FileType go nmap <leader>l  :GoLint .<CR>
```

That only became apparent after looking at the code diff. :)  Not sure if that's what I'm supposed to do, though.

PR #493, opened 2015-07-24
Improve :GoCommand argument handling

guns:
Hi @fatih,

This patch series concentrates on `:GoCommand` argument handling:

1. GoRun, GoTest{,Func,Compile}, GoCoverage, GoGenerate, GoErrcheck, GoLint,
   and GoVet can now all handle an arbitrary number of command arguments so
   that the user can pass more flags, packages, etc. to respective programs.

   This allows for maximum flexibility while still providing simple defaults.

   For instance, #461 is addressed with these changes.

2. All command arguments are now properly escaped for shell invocation,
   which will avoid bugs involving users that have go source paths with vim
   wildcards such as `%` and `#`.

3. All command arguments that will invoked via `makeprg` are escaped with the
   `{special}` option of `shellescape` to avoid a double wildcard expansion.

   Example:

```vim
" I have go source file at /src/99%/file.go
set makeprg=go\ build\ /src/99%/file.go
:make
" make expands the '%' wildcard with the current file!
" => stat /src/99/src/99%/file.go/file.go: no such file or directory
```

These are all issues I've had with vim-go because of my idiosyncratic setup
and tastes. I would love to address any concerns about these patches, so
please don't hesitate to critique.

Thank you very much for vim-go. It really pulls together all the fantastic go
tooling into a very first-class vim experience.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -54,13 +54,11 @@ endfunction
 " suitable for long running apps, because vim is blocking by default and
 " calling long running apps will block the whole UI.
 function! go#cmd#Run(bang, ...)
-    let goFiles = '"' . join(go#tool#Files(), '" "') . '"'
-
     let old_gopath = $GOPATH
     let $GOPATH = go#path#Detect()
 
     if go#util#IsWin()
-        exec '!go run ' . goFiles
+        exec '!go run ' . go#util#Shelljoin(go#tool#Files())
         if v:shell_error
             redraws! | echon "vim-go: [run] " | echohl ErrorMsg | echon "FAILED"| echohl None
         else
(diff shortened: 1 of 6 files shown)


fatih:
Hi @guns. Thanks much! You are in fire :) I'll look at it probably tomorrow or this evening (my timezone). A lot of things has changed in this PR and I want to go over them in a very detailed way. This is really a huge (but a good one) change.

guns:
> A lot of things has changed in this PR and I want to go over them in a very
> detailed way.

Of course. Thank you for reviewing it. I broke up the patches into per-command
changes with some rationale in the patch headers in hopes that it would be
easier to follow. If you'd like to comment in a code-review style via Github
inline comments, I'd be happy to respond in that fashion.

Cheers!

shazow:
Yay this is fantastic! @guns: Huge thanks from a rabid vim-go user. :)

fatih:
This is a very sharp and detailed PR with not problems at all. I've really tested it for 1 hour long with all kind of edge cases I can imagine with no luck :) Kudos to @guns, really impressive work. I don't have nothing to add. Merging.

guns:
Thanks for the review @fatih. Many thanks for vim-go as well. :)

PR #497, opened 2015-07-25
fix :GoRun with g:go_dispatch_enabled

ebfe:
Running `:Make!` will not update the quickfix list (requires `:Copen`).
As a result the errors from the last build/run are shown which is
exceptionally confusing.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -77,7 +77,7 @@ function! go#cmd#Run(bang, ...)
     endif
 
     if g:go_dispatch_enabled && exists(':Make') == 2
-        silent! exe 'Make!'
+        silent! exe 'Make'
     else
         exe 'make!'
     endif


fatih:
Hi @ebfe. The `Make!` is intentional because :GoRun can be a long running task, so that's why it's of type background build. Also we call `:cwindow` here https://github.com/fatih/vim-go/blob/master/autoload/go/cmd.vim#L86, maybe we should call it after the errors are populated https://github.com/fatih/vim-go/blob/master/autoload/go/cmd.vim#L87

ebfe:
`:cwindow` doesn't help. All it shows is the old quickfix list content. To populate the quickfix list with the errors from the `:Make!` command one has to call `:Copen`.

fatih:
Thanks @ebfe for the clarification. I

Issue #499, opened 2015-07-26
GoDrop throws Not enough args for go#imports#SwitchImport


nowk:
Not sure if this is just me. But when I do this:

    :GoImport fmt
    ...
    :GoDrop fmt
    # GoDrop then gives me `Not enough arguments for go#imports#SwitchImport`

PR #500, opened 2015-07-26
Fix Not enough args for go#import#SwitchImport when using :GoDrop

nowk:
This fixes the `:GoDrop <pkg>` issue for me., looks like a missing arg in the call. #499


fatih:
Thanks @nowk :+1:. This must be introduced with the recent changes.

PR #506, opened 2015-07-31
Fix `<Plug>(go-import)` mapping

ichizok:
It lacks a 4th argument.


guns:
I can confirm this bug. This bug, and #500 both existed as of bb64fa1, before the command overhaul.

fatih:
Thanks @ichizok

PR #508, opened 2015-08-01
Add a mapping for going to definition under cursor in a new window

tmrts:
Maps `<Ctrl-W>d` and `<Ctrl-W>D` to split a window and go to definition commands, like `gd`.

### DIFF
--- a/ftplugin/go.vim
+++ b/ftplugin/go.vim
@@ -30,6 +30,9 @@ endif
 
 if get(g:, "go_def_mapping_enabled", 1)
    nnoremap <buffer> <silent> gd :GoDef<cr>
+
+   nnoremap <buffer> <silent> <c-w>d :split<bar>:GoDef<cr>
+   nmap <buffer> <silent> <c-w>D <c-w>d
 endif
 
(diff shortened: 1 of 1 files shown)


fatih:
Hi @tmrts 

There is already mappings for showing it in a `split`, `vertical` and `tab` views. And you can use them like:

```
au FileType go nmap <Leader>ds <Plug>(go-def-split)
au FileType go nmap <Leader>dv <Plug>(go-def-vertical)
au FileType go nmap <Leader>dt <Plug>(go-def-tab)
```

tmrts:
@fatih I didn't know about those but, `gd` is vim's own mapping for goto definition and `<c-w>d` is vim's split the window, goto definition mapping. 
Since, you mapped `gd` I figured `<c-w>d` should be mapped as well

fatih:
I didn't know that. Can you point me to the official doc so I can see how it's being used? If it makes sense we can override the mappings.

tmrts:
@fatih you can find the documentation [here](http://vimdoc.sourceforge.net/htmldoc/tagsrch.html#CTRL-W_d)

PR #509, opened 2015-08-02
GoErrCheck: Only append gopath to report files if necessary

grimborg:
The paths in the `GoErrCheck` reports contain the `GOPATH` twice: instead of `/Users/whatever/gopath/src/github.com/whatever/projectname`, I get `/Users/whatever/gopath/src//Users/whatever/gopath/src/github.com/whatever/projectname`. Pressing `<cr>` on an error opens an empty buffer (with the wrong path).

I've fixed it by checking if the filename (`tokens[1]`) starts with `/`: in this case it's an absolute path, and `GOPATH` should not be appended.

I've tested this on OSX (I always get absolute paths), both with vim and MacVim (7.4).

### DIFF
--- a/autoload/go/errcheck.vim
+++ b/autoload/go/errcheck.vim
@@ -27,7 +27,12 @@ function! go#errcheck#Run(...) abort
             let tokens = matchlist(line, mx)
 
             if !empty(tokens)
-                call add(errors, {"filename": expand(go#path#Default() . "/src/" . tokens[1]),
+                if tokens[1][0] == "/"
+                    let filename = tokens[1]
+                else
+                    let filename = expand(go#path#Default() . "/src/" . tokens[1])
+                endif
+                call add(errors, {"filename": filename,
                             \"lnum": tokens[2],
                             \"col": tokens[3],
                             \"text": tokens[4]})


fatih:
Hi @grimborg. Thanks for the PR! I'm going to look into it soon (probably at the weekend).

fatih:
I think the problem is `errcheck` uses a different command executing process. We use something else in other commands. I've pushed a fix for this to this branch: https://github.com/fatih/vim-go/tree/fix-errcheck-executing  (commit: https://github.com/fatih/vim-go/commit/fb0de36f8a9e3f0b89924170d7f790e21a605052)

Can you please test it? If it works we can merge it.

grimborg:
Great, it works! Thanks, @fatih!

fatih:
Awesome!

hori-ryota:
I have same problem yet. but @grimborg code work good... what should I do?

I tested
```
OSX: 10.10.5
Go: 1.5

VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Aug 26 2015 22:53:28)
MacOS X (unix) version
Included patches: 1-826
Compiled by Homebrew
```

PR #510, opened 2015-08-04
Correct misspellings

syohex:


fatih:
Thanks @syohex, much appreciated :+1:

PR #514, opened 2015-08-05
Change message to PASS after errcheck analyzed without error.

wingyplus:
Currently `:GoErrCheck` show `errcheck analysis...`. It doesn't know how long errcheck analyze complete.


fatih:
Thanks @wingyplus

Issue #353, opened 2015-03-10
Add :GoPlayRun command, to run play.golang.org snippets directly


fatih:
I'm tired creating a new buffer, copying the whole content and running it. We should have something 

```viml
:GoPlayRun http://play.golang.org/p/3WhFYO7qFC
```

Which automatically downloads the script, opens a new buffer and runs it. We have the opposite already `:GoPlay` which creates a new script on http://play.golang.org. This would be also handy.

wingyplus:
I experiment about this feature that use curl to fetch and `nokogiri` to extract code from html in example.

`curl -s http://play.golang.org/p/3WhFYO7qFC | nokogiri -E utf-8 --type html -e 'puts $_.css(\"textarea#code\").children.text'` 

I didn't find any tools better than nokogiri for query selector. What do you recommend about it? I can send a pull request if possible.

fatih:
Append  `.go` to the url and call curl :)

fatih:
Example:

```go
curl http://play.golang.org/p/NyNHXKNxhe.go
```

wingyplus:
Wow, I doesn't know it can. :D

fatih:
I don't see any use case for this.  Closing.

PR #515, opened 2015-08-05
play: run code from go playground snippet

wingyplus:
ref #353, @fatih please review it. :)

### DIFF
--- a/autoload/go/play.vim
+++ b/autoload/go/play.vim
@@ -47,12 +47,9 @@ function! go#play#Run(url)
         return
     endif
 
-    let file = tempname().".go"
-    let curl_command = "curl -s -o " . file . " " . a:url . ".go"
-    let command = curl_command . " && " . "go run " . file
-    let result = system(command)
-
-    call delete(file)
+    let url = a:url . ".go"
+    let command = "curl -s " . url
+    let source_code = system(command)
 
     if v:shell_error
         echo 'A error has occured. Run this command to see what the problem is:'
@@ -60,7 +57,7 @@ function! go#play#Run(url)
         return
     endif
 
-    echo result
+    call go#ui#OpenWindow(tempname() . "_playground.go", split(source_code, "\n"))
 endfunction
 
 function! s:get_visual_content()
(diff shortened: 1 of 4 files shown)


fatih:
Hi @wingyplus. Thanks for PR! My idea was actually to open a new scratch buffer and paste the content to it and **then** run it. Because the reason for this is, I want to play around with it, and eventually going to share it again back with `:GoPlay`. Have a look at:

* Check out http://learnvimscriptthehardway.stevelosh.com/chapters/52.html (Scratch Splits section) to get more info
* Also we have some helper functions, which you can read to get some idea for how to creating scratch views:  https://github.com/fatih/vim-go/blob/master/autoload/go/ui.vim#L4

wingyplus:
@fatih Thank in advice. I modified `go#ui#OpenWindow` helper function (because I grep from source, it's no where to use it.) to show real file name before save and it can modifiable. but it doesn't save to the real file. I think you will review it and discuss about solution at the lastest commit.

Thanks.

fatih:
Thanks @wingyplus. I'm busy lately, I'll look it at soon or later.

wingyplus:
Thanks @fatih

fatih:
Hi @wingyplus. Just tested it and there are couple of things that needs to be fixed:

1. `:GoPlayRun` command should be available in any case, not only when a go file is opened. (Just like `:GoInstallBinaries`, etc..
2. The split window should be vertical as nowadays all displays have more space vertically instead of in the bottom of screen
3. If you call `:GoPlayRun` twice, it overrides the currently open file. Instead it should open another new vertical split view.

wingyplus:
Thanks for comment @fatih. I'll fix tonight.

fatih:
Any new changes for this @wingyplus? It seems we the step three in the list above is still not implemented.

wingyplus:
@fatih Sorry, I'm very busy at the time.Currently, I implemented 1, 2 steps in last two commits (b2cfa6e, 341de9b
).

fatih:
No worries. May be I can pick this up if it's ok for you?

wingyplus:
Yes, you can. :)

fatih:
I'm going to take another stab for this. Thanks @wingyplus for your previous work.

PR #518, opened 2015-08-12
Avoid setting 'filetype' twice

BohrShaw:
Setting 'filetype' the second time is as expensive as the first time:

The previous filetype settings are undone using b:undo_ftplugin and
ftplugin, indent, etc. get fully re-sourced because variables like
b:did_ftplugin are removed then.

Thus use :setfiletype to set the 'filetype' only if not done yet.


fatih:
Thanks @BohrShaw. This looks good.

PR #520, opened 2015-08-15
add g:go_vet_autosave and add opening quickfix by GoVet

kovetskiy:


fatih:
Hi @kovetskiy. Thanks for the PR. I'm busy lately, going to look at it soon or later.

plugin/go.vim
```diff
@@ -132,6 +132,11 @@ augroup vim-go
         autocmd BufWritePre *.go call go#fmt#Format(-1)
     endif
 
+    " running go vet on save
+    if get(g:, "go_vet_autosave", 0)
+        autocmd BufWritePre *.go call go#cmd#Vet(0)
```
fatih: Maybe `BufWriteCmd` is better as we don't change the file at all ?

kovetskiy:
@fatih ping

fatih:
Hi @kovetskiy any update on this?

fatih:
Ping @kovetskiy  :)

kovetskiy:
@fatih I'm here, sorry, I have been busy.

I have a problem with `BufWritePost/BufWriteCmd` and `cc 1`. After running `cc 1` in `BufWritePost or BufWriteCmd` context, vim opens first file in quickfix list, but file `filetype` is empty (`set filetype?` will show `filetype=`), and I can't investigate why.

This can be reproduced with current version.

fatih:
@kovetskiy maybe we should abandon this change because https://github.com/scrooloose/syntastic is already doing it for us. A lot of people are using it. There are many other tools and I think we should add something that is scalable to all other commands, such as `golint`, `errcheck`, etc..

fatih:
Closing this. We have gometalinter which executes many other linters too. What we should to is integrate the linter instead of go vet.

PR #529, opened 2015-09-02
cmd: close quickfix window if successful

### LINKED ISSUE #526: :GoInstall doesn't close the quickfix window when succesful
I used to rely a lot on `:GoBuild`, and the quickfix window with the errors closes itself when there are no more build errors.

`:GoInstall` gives me the exact same quickfix window with compilation errors, but once it succeeds with no errors, the errors for the last unsuccessful build stay open. Am I missing something, can this be closed automatically already somehow?

I'm using the v1.1 tag.

fatih:
Fixes #526

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -108,6 +108,9 @@ function! go#cmd#Install(bang, ...)
             cc 1 "jump to first error if there is any
         endif
         return
+    else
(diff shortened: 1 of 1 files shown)


vdemario:
Thanks @fatih !

fatih:
You're welcome @vdemario :)

PR #533, opened 2015-09-06
fix bug of handling of trailing /

rxwen:
- the GOPATH may be in the form of /home/user/go/ or /home/user/go, the
  trailing / matters. the match pattern may fail if there is trailing /

### DIFF
--- a/autoload/go/package.vim
+++ b/autoload/go/package.vim
@@ -66,7 +66,8 @@ function! go#package#ImportPath(arg)
         return -1
     endif
 
-    return substitute(path, workspace . '/src/', '', '')
+    let srcdir = substitute(workspace . '/src/', '//', '/', '')
+    return substitute(path, srcdir, '', '')
 endfunction
 
 function! go#package#FromPath(arg)


fatih:
Hi @rxwen  Thanks for the contribution. 

Can I ask what this solved for you exactly? Could you also provide a test case (source code) so I can test it my self or see what's happening. Thanks!

rxwen:
Hi @fatih 
Sorry for not elaborating the problem clear enough.

The test scenario is:
1. Set GOPATH to /home/user/go/
2. Run vim to edit $GOPATH/src/github.com/golang/lint/golint/golint.go
3. Move cursor to a function, say usage()
4. run :GoCaller to analysis its callers
5. It should fail

Change the GOPATH to /home/user/go (without trailing /), and try again. It should run ok

The problem is the ImportPath function assumes there is no trailing / in the go pkg paths, so I improved it to support path with or without trailing /.

fatih:
Thanks for the clarification @rxwen I'll look at it later or soon and if everything is ok I'll merge it :+1:

fatih:
This looks good. Thanks @rxwen for the fix :+1:

PR #538, opened 2015-09-13
Add default tab size for Go language.

prodis:


fatih:
Hi @prodis . gofmt handles all formatting related settings. The one your add is user specific settings and we shouldnt' change it ourselves. Thanks for the pr though.

PR #539, opened 2015-09-14
fmt.vim: fix fmt for buffer names existing in nonexisting directories

fatih:
Related problem: https://twitter.com/mitchellh/status/643174982437113860

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -116,7 +116,7 @@ function! go#fmt#Format(withGoimport)
         " below fail
         "
         " :vs bar/foo.go
-        " :!mkidr -p bar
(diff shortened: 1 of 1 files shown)


fatih:
@mitchellh does this solves your problem?

fatih:
@guns there is a test case inside the code, I found a very simple solution, but wonder if it's due something else. Would appreciate if you could have a look at it.

fatih:
Using this fix now for a day it's even more buggy as it changes the filepath. Abonding it. I'll look into a better solution.

mitchellh:
Sorry I didn't get a chance to test this I just saw it now. Let me know if you have any other ideas. Thanks!

PR #540, opened 2015-09-17
Strip spaces from path in GoImport

glittershark:
Currently if I type `GoImport fmt <CR>`, the string `import "fmt "`
will be inserted in my file. This strips leading and trailing spaces
from the path argument to GoImport.


fatih:
Thanks @glittershark. This is a useful addition and fix to vim-go.

PR #547, opened 2015-09-23
Discard quickfix errors when quickfix doesn't locate a file to jump to

### LINKED ISSUE #287: current editing file lost after ：GoRun  output error text
this is a sample:

package main

func main() {
	println("read 0 byte, err read tcp 127.0.0.1:8800: i/o timeout: []")
}

athuth:
Tries to fix #287 by dropping from the QuickFix list any items that don't locate existing files.

Known issue: If `cc 1` jumps to a new buffer immediately, the open-buffer message  `"{filename}" {linenumber}L, {colnumber}C` hides the messages that tell the user about the dropped errors. Haven't been able to figure out how to get around that. The messages are just a little "extra" anyway.

Please review when able and let me know if you find any other issues.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -83,12 +83,33 @@ function! go#cmd#Run(bang, ...)
         exe 'make!'
     endif
 
-    cwindow
-    let errors = getqflist()
+    " Remove any nonvalid filename from the qflist to avoid opening an empty
+    " buffer. See https://github.com/fatih/vim-go/issues/287 for details.
+    let qflist = getqflist()
+    let errors = []
+    let is_readable = {}
+
+    for item in qflist
+        let filename = bufname(item.bufnr)
+        if !has_key(is_readable, filename)
+            let is_readable[filename] = filereadable(filename)
+        endif
+        if is_readable[filename]
+            call add(errors, item)
+        endif
+    endfor
+
+    for k in keys(filter(is_readable, '!v:val'))
+        echo "vim-go: " | echohl Identifier | echon "[run] Dropped " | echohl Constant | echon  '"' . k . '"'
+        echohl Identifier | echon " from QuickFix list (nonvalid filename)" | echohl None
+    endfor
+
+    call setqflist(errors)
     if !empty(errors) && !a:bang
         cc 1 "jump to first error if there is any
     endif
-
+    cwindow
+    
     let $GOPATH = old_gopath
     let &makeprg = default_makeprg
 endfunction
--- a/compiler/go.vim
+++ b/compiler/go.vim
@@ -21,14 +21,15 @@ else
     CompilerSet makeprg=go\ build
 endif
 
-CompilerSet errorformat=
-      \%-G#\ %.%#,
-      \%-G%.%#panic:\ %m,
-      \%Ecan\'t\ load\ package:\ %m,
-      \%A%f:%l:%c:\ %m,
-      \%A%f:%l:\ %m,
-      \%C%*\\s%m,
-      \%-G%.%#
+" Define the patterns that will be recognized by QuickFix when parsing the output of GoRun.
+" More information at http://vimdoc.sourceforge.net/htmldoc/quickfix.html#errorformat
+CompilerSet errorformat =%-G#\ %.%#                   " Ignore lines beginning with '#' ('# command-line-arguments' line sometimes appears?)
+CompilerSet errorformat+=%-G%.%#panic:\ %m            " Ignore lines containing 'panic: message'
+CompilerSet errorformat+=%Ecan\'t\ load\ package:\ %m " Start of multiline error string is 'can\'t load package'
+CompilerSet errorformat+=%A%f:%l:%c:\ %m              " Start of multiline unspecified string is 'filename:linenumber:columnnumber:'
+CompilerSet errorformat+=%A%f:%l:\ %m                 " Start of multiline unspecified string is 'filename:linenumber:'
+CompilerSet errorformat+=%C%*\\s%m                    " Continuation of multiline error message is indented
+CompilerSet errorformat+=%-G%.%#                      " All lines not matching any of the above patterns are ignored
 
 let &cpo = s:save_cpo
 unlet s:save_cpo


fatih:
Hi @athuth 

Thanks for the PR. Can you please give more detail what was wrong and what this is now fixing for us? I don't know if it's a good idea to check all existing files if they exist or not. This could slow down the overall performance.

athuth:
Hey @fatih - here's [a playground snippet][1] that shows the behavior this is trying to fix with `:GoRun`.

Say you're working on a project and you want to show some trace/info/warning lines at runtime, something important enough to print but not important enough to save to a file. Unless your project has only one or two files, you probably pass `log.Lshortfile` or `log.Llongfile`. The compiler `errorformat` picks those up as errors and then tries to open files with names like `"INFO 23:00:00 main.go"`, which of course don't exist, but QuickFix puts them in unlisted buffers anyway. If the first item is one of these, `:GoRun` jumps back into a blank file that doesn't exist, annoying and unhelpful.

Using the `log` package is not the only way you could encounter this, but it's a pretty straightforward case of a feature of the language that doesn't play nice with `:GoRun`.

As far as performance impact, I admit I haven't looked at it. Maybe someone who has a large Go project can try it out and see if it makes a noticeable difference; I'm just starting to get into Go so I have nothing big to test it on, and didn't notice any change with only a few items in the QuickFix list. 

Important to keep in mind, it only checks files that show up in the QuickFix list, which normally would not be every file in the project. And QuickFix has to process each of these items, and create unlisted buffers for all the files it finds, currently, so if you drop some items from the qflist you save some processing by QuickFix.

I could be better about avoiding redundant calls to `filereadable()`, if it's an expensive call. For example:

         let qflist = getqflist()
         let errors = []
    -    let dropped = {}  " use dict to mimic a set/avoid duplicates
    +    let seen = {}
         for item in qflist
             let filename = bufname(item.bufnr)
    -        if filereadable(filename)
    +        if !(get(seen, filename) && get(errors, item)) && filereadable(filename)
                 call add(errors, item)
    +            let seen[filename] = 'added'
             else
    -            let dropped[filename] = ''
    +            let seen[filename] = 'dropped'
             endif
         endfor
         for k in keys(dropped)
             echo "GoRun: Dropped" k "from QuickFix list (file not readable in working directory)"
         endfor
         call setqflist(errors)

That might be a smarter approach, but it's a bit less straightforward and I don't know how it would affect performance, or if it even needs to be optimized.

  [1]: http://play.golang.org/p/ry4pmLqw_y

fatih:
Hi @athuth 

Thanks for tackling this issue, much appreciated! I have two things in mind:

1. I think we can solve this entirely by leveraging it to the parser section in `compiler/go.vim` file. There are a lot details that could detect non files and not include them. Did you check it with `:help errorformat` ?

2. I you have only one file and have ten errors, your current implementation (not the one in your comment) will check the same file ten times. As you said, it needs to changed and improved. Can you add the fixes in your comment to the final PR ?

athuth:
@fatih Both good points.

1. I had the same thought at first but it seems impossible to predict exactly which messages/patterns are errors. I expect some false positives are unavoidable. Have you checked to make sure [the first commit of this PR](https://github.com/athuth/vim-go/commit/6d8b9281f2cb774e319282ab7f3a47bf1c5294b6) correctly documents the goals of the existing patterns?

 It might be possible to improve or add to the errorformat patterns in a way that plays more nicely with the `log` package. I'll look into it when I have an opportunity.

2. I have pushed a commit to avoid multiple redundant checks. A little simpler and better implementation than the example I suggested above.

autoload/go/cmd.vim
```diff
@@ -83,12 +83,28 @@ function! go#cmd#Run(bang, ...)
...
-    let errors = getqflist()
+    let qflist = getqflist()
+    let errors = []
+    let is_readable = {}
+    for item in qflist
+        let filename = bufname(item.bufnr)
+        if !has_key(is_readable, filename)
+            let is_readable[filename] = filereadable(filename)
+        endif
+        if is_readable[filename]:
```
fatih: You need to drop the `:` at the end, it's not a valid syntax.

autoload/go/cmd.vim
```diff
@@ -83,12 +83,28 @@ function! go#cmd#Run(bang, ...)
...
+            call add(errors, item)
+        endif
+    endfor
```
fatih: Add newline after `endfor` please

autoload/go/cmd.vim
```diff
@@ -83,12 +83,28 @@ function! go#cmd#Run(bang, ...)
...
+    for k in keys(filter(is_readable, '!v:val'))
+        echo "vim-go: " | echohl Identifier | echon "[run] Dropped file " | echohl Constant | echon  '"' . k . '"'
+        echohl Identifier | echon " from QuickFix list (not readable in working directory)" | echohl None
+    endfor
```
fatih: Add newline after `endfor` please

autoload/go/cmd.vim
```diff
@@ -83,12 +83,28 @@ function! go#cmd#Run(bang, ...)
         exe 'make!'
     endif
 
-    cwindow
-    let errors = getqflist()
+    let qflist = getqflist()
+    let errors = []
+    let is_readable = {}
```
fatih: Add a newline please

autoload/go/cmd.vim
```diff
@@ -83,12 +83,28 @@ function! go#cmd#Run(bang, ...)
         exe 'make!'
     endif
 
-    cwindow
-    let errors = getqflist()
+    let qflist = getqflist()
```
fatih: Can you put a document above this line and explain why we do the iterating. Something like: `Do not add non valid filenames to avoid opening an empty buffer. Checkout https://github.com/fatih/vim-go/issues/287 for the problem`

compiler/go.vim
```diff
@@ -21,14 +21,15 @@ else
...
-      \%-G%.%#
+" Define the patterns that will be recognized by QuickFix when parsing the output of GoRun.
+" More information at http://vimdoc.sourceforge.net/htmldoc/quickfix.html#errorformat
+CompilerSet errorformat =%-G#\ %.%#                   " Ignore lines beginning with '#' ('# command-line-arguments' line sometimes appears?)
+CompilerSet errorformat+=%-G%.%#panic:\ %m            " Ignore lines containing 'panic: message'
+CompilerSet errorformat+=%Ecan\'t\ load\ package:\ %m " Start of multiline error string is 'can\'t load package'
+CompilerSet errorformat+=%A%f:%l:%c:\ %m              " Start of multiline unspecified string is 'filename:linenumber:columnnumber:'
+CompilerSet errorformat+=%A%f:%l:\ %m                 " Start of multiline unspecified string is 'filename:linenumber:'
+CompilerSet errorformat+=%C%*\\s%m                    " Continuation of multiline error message is indented
+CompilerSet errorformat+=%-G%.%#                      " All lines not matching any of the above patterns are ignored
```
fatih: Thanks for adding the comments, much appreciated :+1: I'm usually spending 10/20 minutes in `errorformat` before jumping and fixing stuff here :)

autoload/go/cmd.vim
```diff
@@ -83,12 +83,28 @@ function! go#cmd#Run(bang, ...)
...
+        if !has_key(is_readable, filename)
+            let is_readable[filename] = filereadable(filename)
+        endif
+        if is_readable[filename]:
+            call add(errors, item)
+        endif
+    endfor
+    for k in keys(filter(is_readable, '!v:val'))
+        echo "vim-go: " | echohl Identifier | echon "[run] Dropped file " | echohl Constant | echon  '"' . k . '"'
+        echohl Identifier | echon " from QuickFix list (not readable in working directory)" | echohl None
```
fatih: Remove the word `file`. We are actually dropping a text here which is not a file. Also the sentence `not readable in working directory` could be better when we rename it to: `(nonvalid file)` (or a better wording?). So in the end it would something like:

```
[run] Dropped "read 0 byte, err read tcp 127.0.0.1:8800: i/o timeout: []" from QuickFix list (nonvalid file)
```

fatih:
Thanks @athuth 

I've pulled and test it. There was an error with the `:` character, once I've removed it evertything. Can you please fix the remaining issues and squash all your changes into one commit. 

I think this is a wonderful contribution. Thank you that you put your time here, much appreciated :+1:

athuth:
@fatih That's what I get for being lazy and testing changes in the installed plugin, then typing them directly into the GitHub web interface, instead of just cloning the branch locally to begin with. Lesson learned.

Fixed and squashed. I'm happy to be able to contribute.

fatih:
Thanks @athuth, merged :+1:

mrnugget:
Thanks for this @athuth! And thanks to @fatih for merging :)

PR #559, opened 2015-10-03
fix 558: fix goTplIdentifier regex so it doesn't go over limits

### LINKED ISSUE #558: syntax: go template identifier doesn't end properly
As shown in the image below, the identifier coloring is going all the way to just before `>` so the `}}` delimiter is missed. 

vim-go-tmpl-syntax-error

DavidGamba:
fixes #558


fatih:
Thanks @DavidGamba for the fix :+1:

PR #564, opened 2015-10-09
Fix path output for GoErrCheck

pdf:
With recent changes to the way errcheck is called (using ExecuteInDir),
prepending the GOPATH results in invalid expansion.


fatih:
Thanks @pdf!

I think the time has come to add tests to vim-go. This is the 4-5 time we fixed the file name in errcheck, everytime because something has changed (because of me or of something else). I'll look into it.

PR #565, opened 2015-10-09
Use absolute paths for gometalinter, so that filenames resolve correctly

pdf:
Because gometalinter is run using ExecuteInDir, it outputs filenames
relative to the source file being edited, which means that you cannot
open them from quickfix.  By using absolute paths, quickfix output is
usable for jumps.

This change also drops the superfluous go_metalinter_path.

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -6,10 +6,6 @@ if !exists("g:go_metalinter_enabled")
     let g:go_metalinter_enabled = ['vet', 'golint', 'errcheck']
 endif
 
-if !exists("g:go_metalinter_path")
-    let g:go_metalinter_path = "./..."
-endif
-
 if !exists("g:go_golint_bin")
     let g:go_golint_bin = "golint"
 endif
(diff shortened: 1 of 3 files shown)


pdf:
It should be noted that `gometalinter` has broken output for `errcheck` (alecthomas/gometalinter#53), but other linters appear to function correctly.

autoload/go/lint.vim
```diff
@@ -18,17 +14,23 @@ if !exists("g:go_errcheck_bin")
     let g:go_errcheck_bin = "errcheck"
 endif
 
-function! go#lint#Gometa(path_to_lint) abort
+function! go#lint#Gometa(...) abort
+    if a:0 == 0
+        let goargs = expand('%:p:h')
```
fatih: You only pass the current file's path. Previously we were recursively looking for all files under the current buffer's directory (with `./...`). This means our `:GoMetaLinter` functionality has changed. I've added `g:go_metalinter_path` just for that, because for my case I want that it goes and lints all files. 

I think I don't get an error because of this simple line in my vimrc:

```
autocmd BufEnter * silent! lcd %:p:h
```

So the fix here is to return absolute paths from gometalinter, not passing a folder just to make it return absolute paths.

pdf: None of the other linters do recursive linting by default, why is GoMetaLinter a special case?  There's no option to tell gometalinter to output full paths, so that's not a possible solution.

Passing a single file to gometalinter is not valid input, you must pass it a path (`./...` is a path).  You should still be able to call `GoMetaLinter ./...` to get that behaviour, like some of the other linters.

I don't like the idea of forcing people to add stuff to their vimrc to get default behaviour working, and changing the working directory by default will produce unintended behaviour for people.

fatih: > None of the other linters do recursive linting by default, why is GoMetaLinter a special case? There's no option to tell gometalinter to output full paths, so that's not a possible solution.

We know the current directory via Vim. So we can prepend it to gometalinter output if we want, but I think this would make it more complex than simpler.

> I understand not wishing to introduce breaking behaviour, but this is pretty fresh code, so there are probably not too many people affected, and at least this behaviour is consistent with the other linters.

Thinking now I think this is fair. Just let me use it a little bit and get a feeling for this.

fatih:
Thanks @pdf for your contributions.

pdf:
@fatih most welcome, thanks for the prompt merges :sunglasses:

PR #568, opened 2015-10-14
Added an explanation of customizing quickfix windows which proved hard to find online in other sources.

karlmutch:
Quickfix window customization is a little hard to find out about online and was a bit of a mystery so I have added a reference to this in the README.  This note could be easily moved to a wiki or help section as appropriate


fatih:
Hi @karlmutch 

I'm using the same setting in my Vimrc too. Maybe we should add this to the wiki? Because this is more a Vim setting rather than vim-go. Just add it to the FAQ please (anyone should have access to it). I'm closing this PR in hope you'll add it to the FAQ. If there is anything to be done please let me know. Thanks!

PR #575, opened 2015-10-15
Treat cygwin vim same as normal unix vim.

### LINKED ISSUE #537: GoRun says go run: no go files listed
I'm trying to launch simple hello world. But when I type :GoRun the result is go run: no go files listed.
Writing go run hello.go it a console works just fine. Writing !go run hello.go from within vim is just fine also. GoFmt works, GoBuild works.

What can be wrong?

shawnbon206:
This is to fix #537 
If we cant have it this way, maybe we can put a variable in there to override this behavior via .vimrc

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -27,7 +27,7 @@ endfunction
 
 " IsWin returns 1 if current OS is Windows or 0 otherwise
 function! go#util#IsWin()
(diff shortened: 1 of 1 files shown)


poxu:
Please merge it. It's pretty imprtant for cygwin users.

fatih:
Thanks @shawnbon206 :+1 This looks good.

PR #580, opened 2015-10-20
Add note on Syntastic lag to README

paked:
There have been a few different issues regarding making `vim-go` and `Syntastic` play well together. The fix I have added to the README in this PR seems to work the best out of all I have seen.

If there is another (better) fix, of course this should be replaced with that... But I believe it is important to have this kind of thing in the README.


fatih:
Hi @paked Usually I'm adding these kind of details into the wiki, but I see a lot of questions come up, so thanks!

PR #590, opened 2015-11-02
Disuse go_fmt_options with GoImports

hori-ryota:
When using `go_fmt_options` for `GoFmt`, `GoImports` also use it.

I want to use `-s` option with `GoFmt`, but `GoImports` will not work.


fatih:
Thanks @hori-ryota much appreciated!

PR #593, opened 2015-11-05
Update def.vim to enable split mode option

drhurd:
Updated def.vim to allow user to set `g:godef_split` to take advantage of split mode options. Plagiarized from [vim-godef](https://github.com/dgryski/vim-godef/blob/master/plugin/godef.vim#L46-L52). Will default to normal behavior if not set.


fatih:
Hi @drhurd. We already support split mode via mappings. For what is this intended?

fatih:
Closing as we already have support for it. Thanks.

PR #601, opened 2015-11-15
syntax: Add support for all the values of GOARCH.

tsuna:
The master list is at
https://github.com/golang/go/blob/master/src/go/build/syslist.go


fatih:
Thanks @tsuna

PR #604, opened 2015-11-16
Fix goCharacter syntax region to be contained

### LINKED ISSUE #603: goCharacter syntax region is uncontained
Using the `gotexttmpl` and `gohtmltmpl` syntax files, the following file content is highlighted incorrectly:

```go
Text after the single quote shouldn't be highlighted as a string.
```
PR is inbound.

moorereason:
This commit sets the goCharacter syntax region to be contained; otherwise,
single-quotes are highlighted as strings outside of the gotplAction regions.

Fixes #603


fatih:
Thanks @moorereason for this fix :+1:

PR #605, opened 2015-11-17
syntax/go: do not treat builtins as keywords

mmlb:
builtins have a very different property from keywords, the identifiers are
settable. It is perfectly valid to override the builtins, though hopefully used
sparingly.

Instead of using `syn keyword`, we now use `syn match` and detect a function
call by searching for an open parentheses.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -117,8 +117,8 @@ syn match       goDeclaration       /\<func\>/
 
 
 " Predefined functions and values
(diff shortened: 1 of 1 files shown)


fatih:
Hi @mmlb 

Can you show me what this changes exactly?

mmlb:
before:
image


after:
image

fatih:
Thanks @mmlb this makes totally sense. :+1:

PR #611, opened 2015-11-20
Fix template variable highlighting

### LINKED ISSUE #610: Template syntax doesn't handle foo{{$bar}}baz
The syntax highlighting doesn't properly handle text like this:

```
foo{{$bar}}baz
```

Here's a screenshot comparing that with `foo{{$bar }}baz` (the space makes it work):

screen_20151120123556

cespare:
Fixes #610


fatih:
Thanks @cespare

Issue #595, opened 2015-11-05
fileformat changed by GoFmt


cochran-at-niche:
I am running gvim on windows 7. When I open a new .go file, I can run the following vim command in order to check the file format:

`:set ff?`

The result is: `fileformat=dos`, which I was I expect, and what I want.

If I then run the GoFmt command, like so:

`:GoFmt`

And again check the fileformat, it now returns `fileformat=unix`, which is undesirable. Furthermore, all of my line endings have changed from crlf to lf, which I do not want (git keeps complaining to me about it).

If I run `:verbose set ff?`, it tells me that the fileformat was `Last set from ~\vimfiles\bundle\vim-go\autoload\go\fmt.go`

If I revert that file (fmt.go) to a version prior to this commit: 01578fd8632870f8285266b93c1e219df19c4002 the problem does not occur. Hence, I believe that the problem was introduced in that commit.

I should also note that I am using goimports for my GoFmt command, as established by this line in my vimrc:

`let g:go_fmt_command = "goimports"`

wingyplus:
I think it changed by `gofmt` not vim-go itself (see https://github.com/golang/go/issues/3961 for more details).

cochran-at-niche:
@wingyplus Then how is the "fileformat" variable, which belongs to vim, getting changed from "dos" to "unix"? If GoFmt is responsible, it should have no effect on the vim variable. Furthermore, when I revert the fmt.go file (ONLY that file), it fixes the problem - how is that explained?

cochran-at-niche:
I think you're correct. The commit I referenced above introduced the use of gofmt's -w option, followed by reloading the current buffer. Since gofmt changes all my line endings to lf, when the current buffer is reloaded, it therefore detects the fileformat as unix.

I submitted a pull-request which fixes this issue: #615

cochran-at-niche:
The pull-request was merged in, fixing the issue. This issue can be closed.

PR #615, opened 2015-11-24
Fix issue #595 - fileformat changed by gofmt

cochran-at-niche:
A simple fix for issue #595.

When :GoFmt is called from within Vim, it runs the gofmt command with the -w parameter, which modifies the source file (actually, it modifies a temp file, which is then swapped in, but you get the idea). One effect of this is that all line endings are changed to lf. When the current buffer is reloaded, vim detects the fileformat as 'unix' due to the lf line endings, even if the fileformat was previously 'dos'. This makes it impossible to maintain windows-style line endings (crlf), if you want them.

Prior to this recent commit: 01578fd8632870f8285266b93c1e219df19c4002, this did not occur. I believe it has to do with the fact that the lines output from gofmt were previously split and inserted line-by-line via Vim itself, which respected the existing fileformat variable. However, in the this commit, the behavior was changed so that gofmt takes care of modifying the file externally, outside of vim (via the -w flag). As a result, the fileformat variable is no longer respected, and is actually changed upon reloading the buffer.

I'm not sure if you'll want to merge this pull-request, or simply accept unix-style line endings as the only valid line-endings for a Go file (this seems to be the convention adopted by the authors of Go: https://github.com/golang/go/issues/3961). In any case, I thought I'd bring it to your attention, because I, at least, wanted to keep using windows-style line endings on my windows machine (I let git do the conversion for me), and I expect that other people might want the same.

Thanks!

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -115,8 +115,8 @@ function! go#fmt#Format(withGoimport)
         try | silent undojoin | catch | endtry
 
         " Replace current file with temp file, then reload buffer
-        call rename(l:tmpname, expand('%'))
         let old_fileformat = &fileformat
+        call rename(l:tmpname, expand('%'))
         silent edit!
         let &fileformat = old_fileformat
         let &syntax = &syntax
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -116,7 +116,9 @@ function! go#fmt#Format(withGoimport)
 
         " Replace current file with temp file, then reload buffer
         call rename(l:tmpname, expand('%'))
+        let old_fileformat = &fileformat
         silent edit!
+        let &fileformat = old_fileformat
         let &syntax = &syntax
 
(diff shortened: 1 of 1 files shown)


fatih:
Hi @cochran-at-niche 

This totally makes sense. I'm ok to merge it as I really don't want to change anythign. Just the order of saving and restoring the filetype seems to be wrong. Shouldn't  we save before the renaming? `silent edit!` just reloads the file actually. Again not sure, can you test it?

cochran-at-niche:
I'm 99% sure that the fileformat variable is changed when the file/buffer is reloaded (`silent edit!`) - that's when vim attempts to detect the fileformat, and ends up setting it to 'unix'. I can confirm that it fixed the issue for me, at least.

In any case, I went ahead and changed it so that the fileformat is saved before the temp file is renamed, just to be safe.

fatih:
Thanks @cochran-at-niche

cochran-at-niche:
You're welcome! :)

Issue #596, opened 2015-11-10
If gofmt is not installed :GoFmt doesn't return an error


fatih:
An issue I've received at DotGo, needs to be debugged and confirmed.

fatih:
Fixed with #617

PR #617, opened 2015-11-25
Check if binary is installed in fmt command. #596

kleww:


fatih:
Also fixes #596. Thanks @kleww :+1:

PR #620, opened 2015-11-26
Use vimproc#get_last_status() instead of v:shell_error if using vimproc.

mattn:
v:shell_error is readonly variable. So vimproc provide way to get last error. Use it instead of v:shell_error.

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -19,12 +19,20 @@ fu! s:gocodeCurrentBuffer()
     return file
 endf
 
-
 if go#vimproc#has_vimproc()
     let s:vim_system = get(g:, 'gocomplete#system_function', 'vimproc#system2')
+    let s:vim_shell_error = get(g:, 'gocomplete#shell_error_function', 'vimproc#get_last_status')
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @mattn

autoload/go/complete.vim
```diff
@@ -19,13 +19,21 @@ fu! s:gocodeCurrentBuffer()
...
 if go#vimproc#has_vimproc()
     let s:vim_system = get(g:, 'gocomplete#system_function', 'vimproc#system2')
+    let s:vim_shell_error = get(g:, 'gocomplete#shell_error_function', 'vimproc#get_last_status')
 else
     let s:vim_system = get(g:, 'gocomplete#system_function', 'system')
+    let s:vim_shell_error = ''
 endif
 
+fu! s:shell_error()
+    let empty(s:vim_shell_error)
```
mattn: ah sorry, this should be `if`.

fatih: Weird I've downloaded and tested it before I've merged and it was working :S

PR #621, opened 2015-11-26
Add go_updatetime to modify value of updatetime

mattn:


fatih:
Hi @mattn , why did you close it?

mattn:
because it should provide way to NOT modify updatetime in default.

PR #622, opened 2015-11-26
Fix check s:vim_shell_error

zchee:
`if` instead of `let`

Ref: https://github.com/fatih/vim-go/pull/620


mattn:
@zchee thanks!

fatih:
I'm still thinking why it was working for me when I tried it locally.

PR #623, opened 2015-11-26
fix bug

mattn:
sorry....


mattn:
merge https://github.com/fatih/vim-go/pull/622

PR #607, opened 2015-11-20
Neovim integration

fatih:
This PR is going to add a proper initial Neovim integration (to make use of Neovim's advanced features). Nothing needs to be installed or to be done. The features will be active only if you use vim-go inside Neovim. 

Currently implemented features:

* An async launcher and base foundation was implemented for the `go` command. This will be used for all upcoming subcommands of the `go` tool.
* `:GoBuild` is now called asynchronously (it doesn't block the UI anymore). 
* A new `go#jobcontrol#Statusline()` can be used to plug into the statusline. This will show the status of the job running asynchronously
* `:GoRun` opens a new vertical terminal and runs the command there. The terminal mode can be changed with `g:go_term_mode`, which is by default `vsplit`. Current options are `vsplit, split or tab`
* Three new mappings to open `:GoRun` terminal in different modes: `<Plug>(go-run-vertical)`,  `<Plug>(go-run-split)` and  `<Plug>(go-run-tab)`: An example usage:
```viml
au FileType go nmap <leader>rt <Plug>(go-run-tab)
au FileType go nmap <Leader>rs <Plug>(go-run-split)
au FileType go nmap <Leader>rv <Plug>(go-run-vertical)
```
* We have two settings for terminal sizes: `g:go_term_height` and `g:go_term_width`. By default a vertical or horizontal is just split by vim automatically. However with these settings we can for example have a terminal with a smaller height when we split it horizontally.
* `:GoTest`, `:GoTestFunc` and `:GoTestCompile` opens and runs in a new terminal. The view mode(split,vertical, tab) is defined with `g:go_term_mode`. 
* If a command inside the term fails (such as `go run`, `go test` ...) we parse now the errors and list them inside in a qf window.
* A `g:go_term_enabled` setting is added to change the behavior of `:GoTest` .If set to `1`, it opens the test commands inside a terminal, if not it runs them in background just like  `:GoBuild` and displays the result in the statusline
* The status_line is improved to show the status per package instead of file. Assume you have three files open, all belonging to the same package, if the package build (`:GoBuild`) is successful, all statusline's will be empty (means SUCCESS), if you it fails all file's statusline will show `FAILED`. 

Please test all features both with Vim and Neovim if you can. Right now documentation, messages and co are just placeholders and will be suited and corrected once it's release ready. Please ask any question. Any feedback is welcome.

PR #626, opened 2015-11-26
Location list

fatih:
Add support for `location list` instead of using `quickfix window`. This allows us to let each window having it's own list of errors displaying. Below is an example of it:

image 2015-11-26 at 7 04 03 pm

Because of consistency, the previous setting: `go_quickfix_height` is renamed to `go_loclist_height`. (the single breaking change we have)

This is a feature I really like. I also refactored all error parsing, populating, displaying, etc.. functions into a single file to make maintaining much more easier in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ disabled/enabled easily.
 * Go to symbol/declaration with `:GoDef`
 * Look up documentation with `:GoDoc` inside Vim or open it in browser
 * Automatically import packages via `:GoImport` or plug it into autosave
-* Compile your package with `:GoBuild`, install it with `:GoInstall`
+* Compile your package with `:GoBuild`, install it with `:GoInstall` or test
+  them with `:GoTest` (also supports running single tests via `:GoTestFunc`)
 * Quickly execute your current file/files with `:GoRun`
-* Run `:GoTest` and see any errors in the quickfix window
 * Automatic `GOPATH` detection based on the directory structure (i.e. `gb`
   projects, `godep` vendored projects)
 * Change or display `GOPATH` with `:GoPath`
@@ -41,6 +41,8 @@ disabled/enabled easily.
   your custom vim function.
 * Tagbar support to show tags of the source code in a sidebar with `gotags`
 * Custom vim text objects such as `a function` or `inner function`
+* All commands support collecting and displaying errors in Vim's location
+  list.
 
 ## Install
 
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -117,7 +117,7 @@ function! go#cmd#Run(bang, ...)
     if !empty(errors) && !a:bang
         call go#list#JumpToFirst()
     endif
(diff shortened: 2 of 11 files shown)


fatih:
Merging this as this is also going to be used by #607.

sasha-s:
This made vim-go  pretty much unusable for me on a laptop (limited horizontal space) with vertical splits.
I am used to having a single wide quickfix window (so I can see the multiple errors without wrapping.
I was also able to navigate to next/prev error (either in same file or not), which was handy.
Is there a way of doing it with location lists?

fatih:
You can change the location of the window. It's up to you, it's not vim-go's fault. Please check out Google or have a look at my dotfiles where I'm modifying it (github.com/fatih/dotfiles). So put this:

```
" This trigger takes advantage of the fact that the quickfix window can be
" easily distinguished by its file-type, qf. The wincmd J command is
" equivalent to the Ctrl+W, Shift+J shortcut telling Vim to move a window to
" the very bottom (see :help :wincmd and :help ^WJ).
 autocmd FileType qf wincmd J
```

So for your case put this into your vimrc:

```
 autocmd FileType qf wincmd L
```

svanharmelen:
I must admit I feel the same way about the move from one quickfix window to multiple (one per buffer) location lists.

Being able to automatically move the lists based on the filetype is nice, but in the case of a location list per buffer, you can very easily end up with a few location lists stacked on top of each other. And if you want to close them with :lclose it means you have to find the buffer it was related to. I had this several times the last week and found it pretty annoying.

All of this was way nicer with the one single quickfix window (which I also automatically moved by the way), so if it could be made configurable (to use either a single quickfix or multiple location lists) that would be really awesome!

Cheers!

fatih:
This PR not only introduced the change to location list, but it also changed how we call an error window. It's pretty easy to change it back to quickfix. But the problem is that quickfix is racy with Neovim integration and not something that fits into the model of NeoVim. For example you can easily execute `:GoBuild` in two windows and edit on a separate third window. Location list makes it more useful here as    each window has it's own list.

svanharmelen:
I must admit that after working with it for a while, I seem to have found my way with it... Still think it's not as clean as using the quickfix window, but I can work with it and I do see the additional benefits you discussed. So I'm good :wink: Thx...

ebfe:
> It's pretty easy to change it back to quickfix.

Any pointers on how to do that?

fatih:
@ebfe It's in `autoload/go/list.vim`. All locallist references needs to be changed to quickfix and that's it.

ebfe:
@faith Thanks. 

Would it be acceptable to have this as an option? While the new behaviour might
be superior, it is somewhat annoying when you are also working with other
languages and have to switch how you navigate build/test errors all the time.

fatih:
In the future maybe @ebfe You can open a separate PR so we don't forget it. But I'm not sure If want to implement based on prior comments.

PR #638, opened 2015-12-01
Update to mention Location lists navigation

owenbutler:
A small section of the readme with some suggestions around navigating through errors in Location Lists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,17 @@ More `<Plug>` mappings can be seen with `:he go-mappings`. Also these are just
 recommendations, you are free to create more advanced mappings or functions
 based on `:he go-commands`.
 
+### Location list navigation
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -156,6 +156,17 @@ More `<Plug>` mappings can be seen with `:he go-mappings`. Also these are just
 recommendations, you are free to create more advanced mappings or functions
 based on `:he go-commands`.
 
+### Location list navigation
+
+vim-go makes use of Location lists for the output of `:GoLint`, `:GoMetaLinter` etc.
```
fatih: We use it for all commands, not only for linters. Changing the sentence to describe that aspect would be more useful.

owenbutler:
Thanks, how does this look?

fatih:
Thanks @owenbutler :+1:

PR #643, opened 2015-12-04
:GoInstallBinaries now installs godoc as well.

smarinov:
Otherwise the built-in `:GoDoc` command doesn't work if you try to bootstrap the necessary Go packages *only* by using `:GoInstallBinaries` from inside Vim.

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -9,6 +9,7 @@ let g:go_loaded_install = 1
 " needed by the user with GoInstallBinaries
 let s:packages = [
             \ "github.com/nsf/gocode",
+            \ "golang.org/x/tools/cmd/godoc",
             \ "github.com/alecthomas/gometalinter", 
             \ "golang.org/x/tools/cmd/goimports",
             \ "github.com/rogpeppe/godef",


bhcleek:
`godoc` is one of the commands that ships with Go and shouldn't need to be installed with `:GoInstallBinaries`.

smarinov:
Sure, I have `go doc` in the console after I installed the [`golang` Debian package](https://packages.debian.org/source/stretch/golang) with its dependencies, but still, when I bootstrap `vim-go` and call `:GoDoc` from within Vim, I get the following error message:

    godoc command not found.
      install with: go get golang.org/x/tools/cmd/godoc

... which basically renders its usage pointless if you don't want to manually create shell aliases. The very idea of `:GoInstallBinaries` should be to prevent such errors and to make `vim-go` self-sufficient, right?

bhcleek:
`godoc` is included in the tarballs in `bin/`. Unlike all the other tools installed by `:GoInstallBinaries`, it's included with the official distributions of Go. I don't have a Debian system readily available. Are you sure it didn't put `godoc` in `dirname $(which go)`?

smarinov:
Absolutely, the only Go-related executables I get after running `ls $(dirname $(which go)) | grep ^go` are `go` and `gofmt`. That's why I am saying that apparently `godoc` isn't necessarily part of the Go distribution itself as I thought previously – which made me do this pull request in the first place.

However, I am new to the Go ecosystem, so I am not sure whether `godoc` and `go doc` are the same thing or not. When I do `go help` I get a suggestion that `doc` is a possible command for `go command [arguments]` and I can _actually do_ `go doc` inside a folder with Go source files and it runs without an error (but no output since I don't know yet what am I supposed to enter in the source files, so that I get some output from it).

MilkChunk:
Install the Debian golang-doc package or remove the Debian packages and use the official installer.

simar7:
On aptitude this package is `golang-go.tools`

smarinov:
@MilkChunk, the package **depends** on `golang-doc`, so I obviously have it installed. I am not sure that a _"correctly installed version of Go"_ includes an executable different from `go` itself. As already said, I have `go doc` that works properly, but I don't have `go-doc` which **Vim-go** is calling for some reason instead of the official way to do things.

However, I found out that even when installing from source, you have to [install additional tools manually from the the go.tools repository](https://golang.org/doc/install/source#tools), so that's why the Debian packages are separated. Thanks @simar7 for mentioning this; but the package `golang-go.tools` is transitional for `golang-golang-x-tools` (for future reference if others have to refer to this issue on Debian-based systems).

PR #649, opened 2015-12-06
Fix empty key when GoRun.

demon386:
This will prevent "Cannot use empty key for Dictionary." error

I encounter this when running GoRun cmd.


fatih:
Hi @demon386 This is [fixed](https://github.com/fatih/vim-go/commit/61c9b49f73603f8a9099cc3655e9c218618bd498) already in #607. I'm going to merge it soon. I'm closing this. Thanks anyway!

PR #671, opened 2015-12-31
Fixed errcheck with abspath

hori-ryota:
Errcheck output errors using relative path format but when failed to check packages using absolute path.

Now, absolute path inputed, vim-go errcheck add unnecessary prefix like
```
/your/gopath/your/gopath/filepath.go
```

Errcheck introduced `abspath` options so fixed by using it.

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -159,21 +159,16 @@ function! go#lint#Errcheck(...) abort
     echon "vim-go: " | echohl Identifier | echon "errcheck analysing ..." | echohl None
     redraw
 
-    let command = bin_path . ' ' . goargs
+    let command = bin_path . ' -abspath ' . goargs
     let out = go#tool#ExecuteInDir(command)
 
     if v:shell_error
-        let errors = []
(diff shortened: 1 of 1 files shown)


fatih:
Hi @hori-ryota Thanks for the PR! Can you please add a test case so I can test it ? I would appreciate it.

hori-ryota:
Where are tests implemented?

Errcheck will failed unbuildable package such as syntax error.

fatih:
@hori-ryota I meant a demo test file you can add so I can test what has changed ?

fatih:
@hori-ryota ping. I've asked for a `demo.go` file which **fails**, but is **fixed** with this PR. It's all so I can test it and reproduce your fix :) Let me know if I can do something here.

hori-ryota:
oh, solly for the delay.

Please test with following demo.go

```go
package main

func main() {
	demo := "demo"
}
```

fatih:
Thanks @hori-ryota. We fixed the issue with errcheck multiple times. I'm going to eventually remove it in the future because we have now :GoMetaLinter and maintaining it is simpler.

PR #673, opened 2016-01-01
Integrate asmfmt

cespare:
This change integrates [asmfmt](https://github.com/klauspost/asmfmt) support for `.s` files, in a similar way to how `:GoFmt` works.

Do take a careful look at this change; I'm not at all a vimscript expert.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ disabled/enabled easily.
 * Share your current code to [play.golang.org](http://play.golang.org) with `:GoPlay`
 * On-the-fly type information about the word under the cursor. Plug it into
   your custom vim function.
+* Go asm formatting on save
 * Tagbar support to show tags of the source code in a sidebar with `gotags`
 * Custom vim text objects such as `a function` or `inner function`
 * All commands support collecting and displaying errors in Vim's location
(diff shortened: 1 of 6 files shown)


fatih:
Thanks @cespare. I'm right now fixing my backlog of issues (FIFO order). I'll look into it.

fatih:
Hi @cespare. This is amazing work. Thank you very much. So looking at it, it seems like we're treating any `*.s` file as asm. For formating itself, I think doing the following would be also sufficient:

```viml
let g:go_fmt_command = "asmfmt"

if get(g:, "go_fmt_autosave", 1)
  autocmd BufWritePre *.go,*.s call go#fmt#Format(-1)
endif
```

Doesn't this work? I'm just curious because theoretically it should work.

cespare:
No, this doesn't work. It breaks formatting of Go files because `asmfmt` doesn't get the right filename, so it tries to format everything as asm.

We could probably get it to work by constructing a tempfile with the same file extension as the current file. However, I'm opposed to using asmfmt for formatting Go code from vim in general:

- If there's an improvement or bugfix to gofmt (or goimports, etc) you have to wait for that change to make it into asmfmt
- If you want to use some other gofmt variant that isn't bundled into asmfmt (it includes goimports and goreturns), you're out of luck

The changes I've made are bigger than what you've suggested, but it also makes it easy to add some better highlighting and indent files for Go asm in the future as well.

fatih:
Got it. I don't have much to say, it's already a great work, and thanks again for the work, much appreciated :+1:

PR #675, opened 2016-01-03
gorename: Output original error if error parsing fails.

### LINKED ISSUE #664: GoRename fails if there is some errors in code
I try to rename some variable in my code when there is some errors in it.

After running ```:GoRename``` command my cursor is moved to first error in code and I cannot even understand what happened. There is any error in status line and Location List with errors is not opened. 
Only after running ```:lopen``` I can view list of errors.

Variable is not renamed.

shazow:
Gorename errors were getting swallowed up if we failed to parse them.

Fixes #664.

### DIFF
--- a/autoload/go/rename.vim
+++ b/autoload/go/rename.vim
@@ -35,6 +35,9 @@ function! go#rename#Rename(bang, ...)
         call go#list#Window(len(errors))
         if !empty(errors) && !a:bang
             call go#list#JumpToFirst()
+        elseif empty(errors)
+            " failed to parse errors, output the original content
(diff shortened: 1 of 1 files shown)


shazow:
There's probably more work left to do for other commands that exhibit a similar bug (:GoTest?) but this should be fairly safe for rename at least.

fatih:
Thanks @shazow. We would do the same thing actually, probably It was silently removed by myself while I refactored the code recently. Thanks for tackling it, much appreciated! :)

shazow:
Sending another PR for GoTest!

PR #676, opened 2016-01-03
GoTest: Output original message on failure return code

shazow:
When using jobcontrol, assume FAILED if return code is >0. Only prints the first line.

This one is a little messier than the GoRename fix. The non-jobcontrol semantics aren't identical, as it prints the entire output rather than just the first line (but it's more consistent with the GoRename fix).

Let me know if you'd like for me to change it to be consistent. The output on unexpected test failures is typically a panic/timeout, so it's quite big.

### DIFF
--- a/autoload/go/jobcontrol.vim
+++ b/autoload/go/jobcontrol.vim
@@ -91,21 +91,9 @@ endfunction
 " references and also displaying errors in the quickfix window collected by
 " on_stderr handler. If there are no errors and a quickfix window is open,
 " it'll be closed.
-function! s:on_exit(job_id, data)
+function! s:on_exit(job_id, exit_status)
   let std_combined = self.stderr + self.stdout
-  if empty(std_combined)
-    call go#list#Clean()
-    call go#list#Window()
-
-    let self.state = "SUCCESS"
-    return
-  endif
-
-  let errors = go#tool#ParseErrors(std_combined)
-  let errors = go#tool#FilterValids(errors)
-
-  if !len(errors)
-    " no errors could be past, just return
+  if a:exit_status == 0
     call go#list#Clean()
     call go#list#Window()
 
@@ -114,6 +102,15 @@ function! s:on_exit(job_id, data)
   endif
 
   let self.state = "FAILED"
+
+  let errors = go#tool#ParseErrors(std_combined)
+  let errors = go#tool#FilterValids(errors)
+
+  if !len(errors)
+    " failed to parse errors, output the original content
+    call go#util#EchoError(std_combined[0])
+    return
+  endif
 
   " if we are still in the same windows show the list
   if self.winnr == winnr()
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -195,6 +195,9 @@ function! go#cmd#Test(bang, compile, ...)
         call go#list#Window(len(errors))
         if !empty(errors) && !a:bang
             call go#list#JumpToFirst()
+        elseif empty(errors)
+            " failed to parse errors, output the original content
+            call go#util#EchoError(out)
         endif
         echon "vim-go: " | echohl ErrorMsg | echon "[test] FAIL" | echohl None
     else


autoload/go/jobcontrol.vim
```diff
@@ -93,7 +93,7 @@ endfunction
 " it'll be closed.
 function! s:on_exit(job_id, data)
   let std_combined = self.stderr + self.stdout
-  if empty(std_combined)
+  if jobwait([a:job_id])[0] > 0
```
shazow: `jobwait(...)` is needed to get the return code of the job. Since it's called on_exit of the respective job, it should never block, so I didn't provide a timeout.

autoload/go/jobcontrol.vim
```diff
@@ -101,20 +101,17 @@ function! s:on_exit(job_id, data)
     return
   endif
 
+  let self.state = "FAILED"
```
shazow: Safe to assume ret>0 == failure?

autoload/go/jobcontrol.vim
fatih: This is a highly critical path as it's used by many applications. As I know relying on exit code might produce wrong results. I don't remember which subcommand or command, but there was a case where I wanted to get the output regardless of job id. Do we need to change it here ?

fatih:
Jobcontrol is only valid for neovim though, did you also test it with Vim? Or is this only valid for Neovim?

autoload/go/jobcontrol.vim
shazow: This should make it equivalent to checking v:shell_error. But yes, if there are commands that specifically avoid checking v:shell_error, then using jobcontrol would yield unexpected FAILURE. Are there any? Only one I can think of off the top of my head is the linter, which refuses to use proper return codes in a silly philosophical protest.

fatih: Ok, valid arguments. Right now we only use it for `build` and `test`. If there is anything wrong with any upstream app, we should fix it instead of.

shazow:
https://github.com/fatih/vim-go/pull/676/files#diff-a3998a862a20bc1665a15c84f401349fR198 is the change for plain-Vim. It's exactly the same change as gorename had. As I mentioned, it dumps the full output unlike the neovim path which dumps just the first line.

I have mixed feelings about that inconsistency, let me know how you feel about it. :)

fatih:
Btw, I've just pulled this and tried it on a very simple app, such as : 

```go
package main

import (
	"fmt"
)

func main() {
	fmt.Println("fatih arslan")
}
```

The output is:

image

I've echoed it and it returns `-1`. There  are two weird things going on here:

1. It seems it doesn't return a list, which contradicts with the help documentation (it says a list of integers)
2. `-1` means

shazow:
Hm what version of neovim are you running? I'm on NVIM 0.1.1.

fatih:
> I've echoed it and it returns -1

Forgot to say which part I've echoed:

```
let a = jobwait([a:job_id])
echo join(a, '')
```

fatih:
> Hm what version of neovim are you running? I'm on NVIM 0.1.1.

I have this too: NVIM 0.1.1 (compiled Dec 28 2015 16:18:38)

shazow:
HMMMMM strange, can't reproduce it on my end. Will dig around, let me know if you stumble on a better way. :)

shazow:
Ah :GoBuild does indeed reproduce, I was trying :GoTest. Hmm.

fatih:
Yeah, I was going to try :GoBuild and :GoTest, couldn't try :GoTest as :GoBuild just failed.

shazow:
Aha, a:data has the return code, so we don't need jobwait. Fixing.

fatih:
Can you rename the variable to `exit_status`? It's ok as `data` for stderr and stdout handlers, but for the exit it would be better. In the future it also would helpful for others :)

shazow:
Fixed! Btw nice way to test a command error,

```
:let g:go_test_timeout="badstring"
:GoTest
```

fatih:
> Fixed! Btw nice way to test a command error,

Great, indeed it works perfect :) Thanks!. Btw can you change the variable name? I think you missed it? If not I can do it later after merging :)  Just let me know please

fatih:
Thanks @shazow :+1:

shazow:
It's changed, I've been force-pushing into the same commit like a gentleman/savage (depending on your practices). :p

shazow:
Happy to help! :custard:

fatih:
Yeah, I had the cache, didn't saw it :facepalm: :)

autoload/go/jobcontrol.vim
```diff
@@ -91,30 +91,27 @@ endfunction
...
   let errors = go#tool#FilterValids(errors)
 
   if !len(errors)
-    " no errors could be past, just return
-    call go#list#Clean()
-    call go#list#Window()
-
-    let self.state = "SUCCESS"
+    " failed to parse errors, output the original content
+    call go#util#EchoError(std_combined[0])
```
dangerhuss: This is causing errors when std_combined is empty.

PR #680, opened 2016-01-04
Add interface highlighting options

StabbyCutyou:
This adds in the ability to have Interface names highlighted like Struct names can be.

Still need to update docs. Open to the suggestion of using a different color, but I didn't have a strong opinion on that.


StabbyCutyou:
Hmmm... I think I messed something up with a rebase I did to fix my author name.

Closing, will re-open

PR #681, opened 2016-01-04
Adding in go_highlight_interfaces option, to provide the same highlig…

StabbyCutyou:
…hting option that structs have

I need to add documentation updates, but this is basically just applying the same approach as structs, but for interfaces. It highlights them the same color as structs - open to suggestions on that, but I had no strong opinion on it.

Tested locally using NeoBundle pointed at my branch.


fatih:
Thanks @StabbyCutyou :+1:

StabbyCutyou:
Thank you!

PR #682, opened 2016-01-06
Use vim-stay for buffer restore if present

### LINKED ISSUE #502: All folds get closed on write
This is obviously highly annoying, having to re-expand whatever fold you need every time you write the buffer.

eapache:
@guns @fatih 

Thanks to @Konfekt for the idea. This fixes #502 if you have https://github.com/kopischke/vim-stay installed.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -55,7 +55,11 @@ let s:got_fmt_error = 0
 "  improvements, patches are welcome :)
 function! go#fmt#Format(withGoimport)
     " save cursor position and many other things
-    let l:curw=winsaveview()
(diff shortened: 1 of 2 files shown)


fvdnabee:
Tested, and this does indeed solve #502 for me.

fatih:
Hi @eapache Thanks for the PR. Even though this solves the problem (albeit it needs to be change a little bit) I'm not keen to depend on vim-stay. It seems like it uses mkview under the hood, and we can implement that functionality into gofmt  ourselves(just like @Konfekt said). I'll going to create a PR with this fix. Thanks for the feedback.

PR #685, opened 2016-01-16
complete: show identifier information after completion is done

### LINKED ISSUE #663: Show function signature when filling its arguments
I like ```:GoInfo``` feature for functions using which I can view function arguments order with their types.

But when typed function name ```someFunction(``` with opening ```(``` I cannot view arguments anymore without returning in Normal-mode and going to function name. 

I would be great to have an option to see ```:GoInfo``` for used function when filling its argument list.

fatih:
This is a new feature that prints the information of the completed identifier in statusline. This is experimental.

closes #663

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -122,9 +122,16 @@ endfunction
 " ============================================================================
 "
 function! s:echo_go_info()
-    let offset = go#complete#gocodeCursor()
-    let result = go#complete#GetInfoFromOffset(offset-1)
-    redraws! | echo "vim-go: " | echohl Function | echon result | echohl None
+    if !exists('v:completed_item') || empty(v:completed_item)
+        return
+    endif
+    let item = v:completed_item
+
+    if !has_key(item, "info")
+        return
+    endif
+
+    redraws! | echo "vim-go: " | echohl Function | echon item.info | echohl None
 endfunction
 
 augroup vim-go
@@ -135,6 +142,8 @@ augroup vim-go
         autocmd CursorHold *.go nested call go#complete#Info()
     endif
 
+    " Echo the identifier information when completion is done. Useful to see
+    " the signature of a function, etc...
     autocmd CompleteDone *.go nested call s:echo_go_info()
 
     " Go code formatting on save
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -119,6 +119,12 @@ endfunction
 
 " Autocommands
 " ============================================================================
+"
+function! s:echo_go_info()
+    let offset = go#complete#gocodeCursor()
+    let result = go#complete#GetInfoFromOffset(offset-1)
+    echo "vim-go: " | echohl Function | echon result | echohl None
+endfunction
 
 augroup vim-go
     autocmd!
@@ -127,6 +133,8 @@ augroup vim-go
     if get(g:, "go_auto_type_info", 0)
         autocmd CursorHold *.go nested call go#complete#Info()
     endif
+
+    autocmd CompleteDone *.go nested call s:echo_go_info()
 
     " code formatting on save
     if get(g:, "go_fmt_autosave", 1)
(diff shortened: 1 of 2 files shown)


kodek:
This is exactly what I'm looking for, but it seems to be a bit glitchy at the moment. I generally use tab to select a match, then I type the opening parenthesis to close the autocomplete list. If I do that, the function signature does not show up. How are you triggering the autocomplete of "" and the opening/closing parentheses?

Also, I had to set noshowmode to hide the overlapping mode status. Powerline shows it anyway.

fatih:
@kodek push a better approach. Now it's much more efficient and should just work out of the box. Can you please test and return me your thoughts ?

kodek:
Not working at all right now :( Let me know if there's anything you'd like me to test.

remeh:
Is it also bound on a command (e.g GoSignature) or is it only on auto-complete ? Anyway, great feature!

fatih:
@remeh it's all automatically. It works perfect for me and actually all I do is echoing the current completion from the cache. There is no magic, but the statusline is a place that many other plugins, sub systems of vim uses. So even though I display it, it might be overwritten meanwhile by something else. So it's tricky :)

plugin/go.vim
```diff
@@ -129,6 +142,10 @@ augroup vim-go
         autocmd CursorHold *.go nested call go#complete#Info()
     endif
 
+    " Echo the identifier information when completion is done. Useful to see
+    " the signature of a function, etc...
+    autocmd CompleteDone *.go nested call s:echo_go_info()
```
Shougo: I think the check of CompleteDone is needed.
`if exists('##CompleteDone')`

fatih: Yeah definitely @Shougo :+1: Should it not with one `#` ? Your example has two `##`. If yes why it is that way ?

Shougo: `##event` is the check of the event is supported.
From `:help exists()`:

```
			##event		autocommand for this event is
					supported.
```

kodek:
I think my issue is related to a YCM conflict. I just realized I might be using the wrong autocomplete engine. Trying to fix, but I'll report back. Sorry!

plugin/go.vim
fatih: Ok, thanks! Fixed now.

gonzaloserrano:
@kodek i'm also using YCM and i must have the info beeing overwritten so if you find the reason please share :+1:

eduncan911:
I can't seem to get this to work.  Vim acts normally:

* no `match # of #` ever shows up while scrolling through the C-x C-o menu options
* `-- INSERT --` remains in the statusline at all times.

On this revision:
```
commit fd5661a1e16a1fd41385d7011877bfa1f0a1353f
Merge: 5f64a57 65ef9bc
Author: Fatih Arslan <ftharsln@gmail.com>
Date:   Thu Mar 3 09:37:40 2016 +0200

    Merge pull request #748 from bhcleek/master
    show output of :GoRun
```

Tried a few random things:

* disabled neocomplete
* disabled omnicomplete's `completeopt` changes, the "Preview" buffer came back.
* had several omnicomplete tweaks (`inoremap`), disabled all.
* saw issues with symlinks and vim; moved my /USB/HDD/go/src/... to `~/go/src/...` (no more symlinks)
* disabling `vim-sensible`

For testing, I closed VIM each time and attempted reopen and edit a known .go file with known local func()s as well as accessing some of the standard library packages (`os`, `fmt`), and some other packages I wrote in my github location.  

Nada.

One thing to note: `:GoDoc` doesn't seem to be able to find anything in my GOPATH.  


**EDIT**: I noticed that if I insert this into my `.vimrc`, that `match # of #` now shows along with the func definition:

`set noshowmode`

I have [vim-sensible](https://github.com/tpope/vim-sensible) setup the defaults for my .vimrc; so, perhaps it has `showmode` or alike.

^- while that has been fixed, I still do not get the auto-completion of the "" when selecting `fmt.Printf()` from the autocomplete.

Issue #384, opened 2015-04-13
Better :GoCoverage support


fatih:
Instead of opening a HTML page via the system default browser, open a Vim scratch buffer and show an annotated view of the buffer. Here is an example script:

https://gist.github.com/guelfey/6253836

And here is a plugin that does it fairly well:

https://github.com/t-yuki/vim-go-coverlay

oryband:
:+1:

fatih:
Closing this in favor of https://github.com/t-yuki/vim-go-coverlay. We could merge this into vim-go if @t-yuki agrees, but it doesn't need to :)

oryband:
wow, this tool is amazing. major productivity boost
![](http://theartmad.com/wp-content/uploads/2015/06/Amazing-Funny-Photos-11.jpg)

t-yuki:
It's welcome to merge coverlay into vim-go.
However I think the quality of code is low: its implementation is a bit tricky and not good code with not-tested well even if it works as a plugin of vim-go.
So I'm afraid to make PR currently. Fork or patches are also welcome!

oryband:
@avivl

PR #686, opened 2016-01-16
Neovim integration for GoCoverage

t-yuki:
Hi, I'm working on nvim integration of my coverage plugin: https://github.com/t-yuki/vim-go-coverlay
However, I've realized that go#cmd#GoCoverage is not integrated with nvim yet and several API changes are required for better integration.

Moreover, I guess that a recent change 579dd89f broke go#cmd#GoTestFunc because go#util#Shelllist also escapes "-run" flag.

So I want to include changes:

* cmd: fix GoTestFunc is not working
* cmd: show testing and test pass message in GoTest on neovim
* jobcontrol: add AddHandler and RemoveHandler API to catch on_exit callback for callers of caller
* cmd: add neovim integration for GoCoverage

However, I used ugly event handler method to track job status by go#cmd#GoTest and go#cmd#GoCoverage function. Another solution is welcome.

refs #607

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -213,11 +213,11 @@ function! go#cmd#Test(bang, compile, ...)
 
     if has('nvim')
         if get(g:, 'go_term_enabled', 0)
-            call go#term#new(a:bang, ["go"] + args)
+            let id = go#term#new(a:bang, ["go"] + args)
         else
-            call go#jobcontrol#Spawn(a:bang, "test", args)
+            let id = go#jobcontrol#Spawn(a:bang, "test", args)
         endif
-        return
+        return id
     endif
 
     call go#cmd#autowrite()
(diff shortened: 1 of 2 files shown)


fatih:
@t-yuki thank you a lot for your contribution! I'm going to look at it asap.

fatih:
@t-yuki the whole handler additions are, in my opinion, not needed. I think just the fix for `:GoTestFunc` would be OK. I've actually fixed it for plain `vim`, but neovim is still alpha and second citizen as support, I think that's the reason I didn't picked it up yet :) Is it possible for you to remove these two:

```
cmd: show testing and test pass message in GoTest on neovim
jobcontrol: add AddHandler and RemoveHandler API to catch on_exit callback for callers of caller
```

It's already complex right now and the message is OK in my opionin. Tough, I'm curious about your thoughts. Let me know what you think.

t-yuki:
Sorry for late reply, ok, I'll resemble PR.
I agree with you that it is too complex.
I'll try to implement job control in my plugin.
If async job is supported on vim officially, let's reconsider it.

t-yuki:
I wrote #695 
However, It doesn't distinguish the job is GoBuild or GoTest because there is no context information in job struct.
To do this, adding completion event handler mechanism or attaching context information or callback handler information to `jobcontrol#Span` method as a argument will be required.
I think adding `callback` function pointer argument to `jobcontrol#Span` is more acceptable.

fatih:
Thanks @t-yuki. I'll look into this once we merge #695

fatih:
Hi @t-yuki do you still want to work on this? If yes could you please again rebase it? I'm more than happy to review it just let me know please! Thanks :)

t-yuki:
Rebased, but not refactored as proposed in the above.
Also I removed showing test pass message because the master shows `SUCCESS` or `FAIL` in jobcontrol.
see: https://github.com/fatih/vim-go/compare/master...t-yuki:nvim-cov-wip#diff-a3998a862a20bc1665a15c84f401349fR185

Well, I'm looking for the way to work with latest vim's `job_start` function instead of nvim.
It's nice if `jobcontrol` can work with `job_start`, not only nvim.

fatih:
@t-yuki just tested it works well but if `g:go_term_enabled=1` is set, it doesn't work. Did you test that case ?

t-yuki:
Not yet.
I just tested and it works except it can't change current window using C+w so I can't return to original buffer, meaningless.
However, I guess it is reproducible in GoTest on master, not originated in this change.

svanharmelen:
@fatih @t-yuki would be awesome if we could land this PR and the related `vim-go-coverlay` code :+1:

fatih:
Thanks @t-yuki a lot for the PR. I really would like to merge `vim-go-coverlay` into this with some modifications. If you are OK I can create a PR with your changes (or you can do it), and we can go over it. Let me know what you think :) You did earlier that you would be OK merging it (https://github.com/fatih/vim-go/issues/384#issuecomment-115608142). I think we can go over the code and make it nicer and integrate with vim-go's codebase (which has many great utilities and base functions). I'm more than happy to work with you, just let me know.

PR #691, opened 2016-01-19
fmt: save folds on close

fatih:
Fixes: #502

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -54,8 +54,8 @@ let s:got_fmt_error = 0
 "  this and have VimL experience, please look at the function for
 "  improvements, patches are welcome :)
 function! go#fmt#Format(withGoimport)
-    " save cursor position and many other things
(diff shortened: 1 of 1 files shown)


eapache:
:+1: works for me with or without `vim-stay` installed

nhooyr:
Works for me as well.

fvdnabee:
Works here also, without vim-stay.

fatih:
Thanks for all the comment. Please use it for couple of days/weeks. This function is fairly critical so I don't want to break something here :)

cassava:
:+1: Works for me. So excited!

osleg:
Works for me as well.... could it be merged?

mattn:
when user doesn't set viewdir, default value of viewdir become $VIM i.e. /usr/share/vim. So mkview! makes error. So you need to give filename for mkview!.

eapache:
Why `mkview` uses `$VIM` and not `$HOME` on windows is beyond me...

Issue #404, opened 2015-04-29
:GoInstall error navigation seems broken


daaku:
Haven't figure out what exactly is happening, but this was working fine until I upgraded yesterday. Steps to reproduce:

1. Start vim in a say `$HOME`
1. Open in go file say `$HOME/go/src/foo/foo.go`
1. `:GoInstall foo/..`
1. If there is a error from the `GoInstall` in `$HOME/go/src/bar/bar.go` you'll see it listed but won't be opened as it used to be.

fatih:
Hi @daaku. Weird thing indeed. The thing is there is no code that changed `:GoInstall` since the previous release. If there is any error, `:GoInstall` calls `cwindow` which opens the quick fix list and shows all errors. It doesn't jump to it though, you need to call `:cc`. In your case, you want to jump to it, just like `:GoBuild` right ?

daaku:
It's getting the absolute paths wrong, `:cc` doesn't work either. `:GoBuild` does in fact get them right.

fatih:
It's so hard to test this without a proper test files. I've tried in my current GOPATH with several other packages, and in all of them it was working. Is there a way you could paste me a self contained go code so I could directly reproduce it ?

fatih:
I also wonder if this is due: https://github.com/fatih/vim-go/commit/0d1e8c945481128ec7e17b51780a85af5046224d But again there is no way I can test it. Pinging @mattn may be he has some idea. I couldn't reproduce it (waiting for your example files)

mattn:
I'll look into it later

fatih:
@daaku can you test with https://github.com/fatih/vim-go/pull/405 please ?

daaku:
I pulled but it still fails for me :/ Here's a repro case:

```sh
# setup a test case
export GOPATH=$PWD/tmp
mkdir -p $GOPATH/src/foo
mkdir -p $GOPATH/src/bar
cat << 'EOF' > $GOPATH/src/foo/foo.go
package foo

import "fmt"

func Do() {
  fmt.Println("hello from foo")
}
EOF
cat << 'EOF' > $GOPATH/src/bar/bar.go
package main

import (
  "fmt"
  "foo"
)

func main() {
  foo.Do()
  fmt.Println("hello from bar")
}
EOF
go install -v ...
$GOPATH/bin/bar

# now break the code so go install fails
cat << 'EOF' > $GOPATH/src/foo/foo.go
package foo

func Do() {
[...]

PR #692, opened 2016-01-19
use make for GoInstall to correctly parse error paths

jellevandenhooff:
I think I ran into a regression of #404 (where `:GoInstall` did not parse the paths in the output correctly.)

I fixed my issue locally with this PR by scavenging some logic from `go#cmd#Build`. However, I don't really know what I'm doing, so if there is a better way to fix this issue, let me know.

Thanks,


fatih:
Thanks @jellevandenhooff for the fix :+1:

PR #694, opened 2016-01-21
Fix mode of non-executable files

nforro:
The files *doc/vim-go.txt* and *ftplugin/go/commands.vim* have executable bit set, probably by mistake. That is causing problems with vim-go package in Fedora.


fatih:
Yeah, probably by mistake, they imho shouldn't be executable. Thanks @nforro for the fix :+1:

PR #700, opened 2016-01-26
support configurable list types

### LINKED ISSUE #696: quickfix vs location list
The most recent release, 1.4, changed lists to use the location list instead of the quickfix list. This is a useful improvement for functions that operate on a single buffer (e.g. GoCallers). But for functions that operate on a package instead of the buffer (e.g GoBuild), I strongly prefer the quickfix list, (e.g. GoBuild). 

I'm not certain, but I think the difference in my mind is that the location list is most appropriate for lists of locations that are the result of a search (e.g. GoImplements), while the quickfix list is appropriate for lists of locations that communicate needed user action (e.g. GoTest). 

I've read your comments on #626. Would you consider a pull request that introduces a new option to allow the quickfix list to be used for functions that generate a list of errors that require action if the new option's default is to preserve the 1.4 behavior?

bhcleek:
* Prefer the quickfix window for commands whose output is considered to
  be errors for more than current buffer. e.g. :GoBuild.

* Prefer the location list window for commands whose output is related
  specifically to the current buffer only (e.g: :GoFmt and
  :GoImplements).

* Rename 'g:go_loclist_height' to 'g:go_list_height'.

* Add a new option, 'g:go_list_type', to configure whether to always use
  the quickfix window, a location list window, or to vary based on the
  preferred window as documented above.

@fatih I'm not sure if you'd want to keep the default as I've coded it. The default here is to vary the list by the command, but that changes the 1.4 behavior, and we'd talked about preserving the 1.4 behavior by default.

I tried to update the relevant documentation where the location list was referenced to be accurate according to the preferred windows that I described above, but please double check my work.

Consider this a start point - I realize you may have lots of feedback; I'm happy to do any rework that you find necessary.

Fixes #696

### DIFF
--- a/autoload/go/list.vim
+++ b/autoload/go/list.vim
@@ -5,15 +5,15 @@ endif
 " Window opens the list with the given height up to 10 lines maximum.
 " Otherwise g:go_loclist_height is used. If no or zero height is given it
 " closes the window
-function! go#list#Window(quickfix, ...)
-    let l:quickfix = go#list#Type(a:quickfix)
+function! go#list#Window(listtype, ...)
+    let l:listtype = go#list#Type(a:listtype)
     " we don't use lwindow to close the location list as we need also the
     " ability to resize the window. So, we are going to use lopen and lclose
     " for a better user experience. If the number of errors in a current
     " location list increases/decreases, cwindow will not resize when a new
     " updated height is passed. lopen in the other hand resizes the screen.
     if !a:0 || a:1 == 0
-        if l:quickfix == 0
+        if l:listtype == "locationlist"
             lclose
         else
             cclose
@@ -31,7 +31,7 @@ function! go#list#Window(quickfix, ...)
         endif
     endif
 
-    if l:quickfix == 0
+    if l:listtype == "locationlist"
       exe 'lopen ' . height
     else
       exe 'copen ' . height
@@ -40,9 +40,9 @@ endfunction
 
 
 " Get returns the current list of items from the location list
-function! go#list#Get(quickfix)
-    let l:quickfix = go#list#Type(a:quickfix)
-    if l:quickfix == 0
+function! go#list#Get(listtype)
+    let l:listtype = go#list#Type(a:listtype)
+    if l:listtype == "locationlist"
         return getloclist(0)
     else
         return getqflist()
@@ -50,9 +50,9 @@ function! go#list#Get(quickfix)
 endfunction
 
 " Populate populate the location list with the given items
-function! go#list#Populate(quickfix, items)
-    let l:quickfix = go#list#Type(a:quickfix)
-    if l:quickfix == 0
+function! go#list#Populate(listtype, items)
+    let l:listtype = go#list#Type(a:listtype)
+    if l:listtype == "locationlist"
         call setloclist(0, a:items, 'r')
     else
         call setqflist(a:items, 'r')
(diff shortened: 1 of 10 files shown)


fatih:
Thanks @bhcleek for the PR. I'm aware of this just hadn't the time to look at in more detail. This is a needed PR as it's really annoying to me too lately. So I want to use this to test how it works with quickfix + neovim. I'll let you know and also add comments.

bhcleek:
Thank you for the update @fatih. I'm really looking forward to getting this merged; I've been using it since I submitted it and am pleased with the lists.

fatih:
@bhcleek could you rebase it again on top of master? Sorry for the inconvenience but seems like it's not mergable anymore due recent changes.

bhcleek:
@fatih I've rebased on top of master.

fatih:
Can you rename all `a:quickfix` variables to `a:ltype` (or `a:listype`) ? It's confusing as it's not clear what it is.

autoload/go/list.vim
```diff
@@ -22,50 +31,96 @@ function! go#list#Window(...)
...
+function! go#list#Clean(quickfix)
+    let l:quickfix = go#list#Type(a:quickfix)
+    if l:quickfix == 0
+        lex []
+    else
+        cex []
+    endif
 endfunction
+
+function! go#list#Type(quickfix)
```
fatih: I'm not sure if we need this. It's yet another abstraction and seems like we don't need it. Can you remove it? Anywhere else just let use strings. Like:

```
if a:listype == "quickfix"
endif
```

This is more readable and we don't need to manually parse what `0` or `1` means.

fatih:
@bhcleek so as a final note, remove the `Type()` function and pass strings, such as "quickfix" or "locationlist" to each appropriate function.

bhcleek:
There are two different times that resolution of the preferred window and `g:go_list_type` needs to happen; in the `go#list` functions themselves to decide which type of window to manipulate, and also in the callers of those functions, to decide which kind of command to run (e.g. `lmake` vs `make`). To satisfy that need, removing `go#list#Type()`, will require that the code that it contains be sprinkled in lots of places, because of the need to resolve the preferred window according to how the user has set `g:go_list_type`.  With that in mind, do you still want `go#list#Type()` removed?

bhcleek:
@fatih I refactored to pass strings instead of integers. I'll wait to make the change to remove `go#list#Type()` until I get confirmation that you want it removed in light of my previous comment. Perhaps it would be better to rename it so its utility is clearer?

fatih:
@bhcleek sorry but I still didn't grasp the importance of `go#list#Type()`. It retusn the passed string back to you. You have that information already? For example right now this is how the code is written:


```viml
    let l:listtype = go#list#Type(a:listtype)
    if l:listtype == "locationlist"
        return getloclist(0)
    else
        return getqflist()
    endif
```

But you can also write this as:

```viml
    if a:listtype == "locationlist"
        return getloclist(0)
    else
        return getqflist()
    endif
```

So what's the difference ?

bhcleek:
`go#list#Type()` doesn't just return what was passed to it. It returns the list type based on `g:go_list_type`. It returns what was passed in (e.g. `a:listtype`) only if `g:go_list_type` is not set to always use the quickfix or location list window.

Maybe this will help: https://github.com/bhcleek/vim-go/blob/master/doc/vim-go.txt#L1010

fatih:
Thanks @bhcleek, now it makes sense :+1:  There is nothing that keeps me to merge this, I've already pulled and working with. Let me use this way 1,2 day and I'll merge it (so I can catch any errors meanwhile as I'm heavily using vim-go while developing our projects in my full time job). Thanks again :+1:

fatih:
Thanks @bhcleek a lot for your contribution. This is very useful and most needed feature we needed. If you see any errors or any improvements please feel free to open an issue or send a PR to discuss further.

bhcleek:
:+1:

bhcleek:
I'm happy to help @fatih. If I find any problems, I'll submit a PR to get them fixed. FWIW, I've been using this for two weeks successfully.

:+1:

svanharmelen:
Really like this one :+1: Thx!

PR #709, opened 2016-02-03
Indent code in Go HTML templates

jinzhu:
Before when write go code in templates, it is not indented correctly,  for example:

```
{{if true}}
hello world
{{end}}
```

I think this should be indented to:

```
{{if true}}
  hello world
{{end}}
```

This pull request did the job, fixed the indention for `if/else`, `range`, `with`, `block`, `define`

### DIFF
--- a/indent/gohtmltmpl.vim
+++ b/indent/gohtmltmpl.vim
@@ -3,3 +3,42 @@ if exists("b:did_indent")
 endif
 
 runtime! indent/html.vim
+
+" Indent Golang HTML templates
+setlocal indentexpr=GetGoHTMLTmplIndent(v:lnum)
+setlocal indentkeys+==else,=end
+
+" Only define the function once.
+if exists("*GetGoHTMLTmplIndent")
+  finish
+endif
+
+function! GetGoHTMLTmplIndent(lnum)
+  " Get HTML indent
+  if exists('*HtmlIndent')
+    let ind = HtmlIndent()
+  else
+    let ind = HtmlIndentGet(a:lnum)
+  endif
+
+  " The value of a single shift-width
+  if exists('*shiftwidth')
+    let sw = shiftwidth()
+  else
+    let sw = &sw
+  endif
+
+  " If need to indent based on last line
+  let last_line = getline(a:lnum-1)
+  if last_line =~ '^\s*{{\s*\%(if\|else\|range\|with\|define\|block\).*}}'
+    let ind += sw
+  endif
+
+  " End of FuncMap block
+  let current_line = getline(a:lnum)
+  if current_line =~ '^\s*{{\s*\%(else\|end\).*}}'
+    let ind -= sw
+  endif
+
+  return ind
+endfunction


fatih:
Thanks @jinzhu for your contribution. @cespare do you mind have a look at it ? I'm not using templates recently.

jinzhu:
I have made a test template to try to cover all cases, hope it could be helpful.

```
<div class="qor-page__body">
  {{render "hello1"}}
  {{render "hello2"}}

  {{if true}}
    <p>
      hello world
      {{if hello}}
        hello world
      {{else}}
        hello world
      {{end}}
    </p>
  {{else}}
  {{end}}

  {{render "hello3"}}
</div>
```

cespare:
I can take a look in the next few days.

fatih:
Thanks @cespare :+1:

cespare:
This change doesn't work for me at all, because the `HtmlIndent` function is not defined.

fatih:
@jinzhu any comments on that ?

jinzhu:
Hi @cespare 

This method is defined in vim, file `indent/html.vim`, which is required from here https://github.com/fatih/vim-go/blob/master/indent/gohtmltmpl.vim#L5

image

jinzhu:
From vim's source code. https://github.com/vim/vim/blob/089af18d1fd0961ff504ee72db0156bbfe509cdf/runtime/indent/html.vim#L947

cc @fatih

fatih:
Maybe we should add a clause to check if this exists or not by checking the version and patch level. An example how to do it: http://vi.stackexchange.com/a/2467/2791

cespare:
BTW I'm using a fairly recent vim (7.4 something, can't check the exact version at the moment).

I suspect that calling this function is the wrong approach in general. When I look at indent files for other templating languages, they don't do it this way. Example: https://github.com/mitsuhiko/vim-jinja/blob/master/indent/htmljinja.vim

jinzhu:
I just pushed another commit, fixed it doesn't work on vim 7.3,  I have tested new fix with below three versions, all works for me. @cespare could you test it works for you or not?

```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Feb 10 2016 14:45:20)
MacOS X (unix) version
Included patches: 1-1257
```

```
NVIM v0.1.1-145-g074d3dc (compiled Jan  6 2016 09:09:15)
Build type: Dev
```

```
VIM - Vi IMproved 7.3 (2010 Aug 15, compiled Oct 27 2015 16:22:14)
Compiled by root@apple.com
Normal version without GUI.  Features included (+) or not (-):
```

I think vim-jinja is just using another solution to make things works, from github search results count, it is less used that current solution. I haven't found much documents about this when writing the code, so just pick up a simper solution. ;)

cespare:
Works for me.

LGTM.

indent/gohtmltmpl.vim
```diff
@@ -3,3 +3,42 @@ if exists("b:did_indent")
...
+  if exists('*shiftwidth')
+    let sw = shiftwidth()
+  else
+    let sw = &sw
+  endif
+
+  " If need to indent based on last line
+  let last_line = getline(a:lnum-1)
+  if last_line =~ '^\s*{{\s*\%(if\|else\|range\|with\|define\|block\).*}}'
+    let ind = ind + sw
```
fatih: `let ind .= sw`

indent/gohtmltmpl.vim
```diff
@@ -3,3 +3,42 @@ if exists("b:did_indent")
...
+  endif
+
+  " End of FuncMap block
+  let current_line = getline(a:lnum)
+  if current_line =~ '^\s*{{\s*\%(else\|end\).*}}'
+    let ind = ind - sw
```
fatih: `let ind -= sw`

indent/gohtmltmpl.vim
```diff
@@ -3,3 +3,42 @@ if exists("b:did_indent")
...
+  endif
+
+  " The value of a single shift-width
+  if exists('*shiftwidth')
+    let sw = shiftwidth()
+  else
+    let sw = &sw
+  endif
+
+  " If need to indent based on last line
```
fatih: What do you mean here exactly?

jinzhu: Indention usually based on last line, for below example, the second line need to be indented as last line has a `if` condition

```go
{{if true}}
  hell world
```

fatih:
@jinzhu this looks good. Can you squash it into one commit if possible?

indent/gohtmltmpl.vim
fatih: Thanks :+1:

jinzhu:
@fatih sure, just did this ;)

fatih:
Thanks @jinzhu for the contribution :+1:

PR #713, opened 2016-02-05
complete: fix ycm

fatih:
Seems like YCM is setting it's own omnifunc and also started to use
gocode itself recently. This makes vim-go very slugish and also breaks
it. I really hate it as YCM tries to do a lot of things (such as
GoToDefinition). Now it's not compatible anymore with vim-go because it
tries to do the completion itself instead of letting the application do
it.

### DIFF
--- a/ftplugin/go.vim
+++ b/ftplugin/go.vim
@@ -20,8 +20,12 @@ setlocal noexpandtab
 
 compiler go
 
-" Set gocode completion
-setlocal omnifunc=go#complete#Complete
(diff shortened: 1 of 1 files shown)


casimir:
Hi, innocent and optimistic user here. I use vim-go and YCM from the very beginning, and I'm fairly happy with their association. Beside YCM override the `omnifunc` on first completion I can't see the problem here. Would you mind explain how the problem manifests for the final user?

PR #721, opened 2016-02-11
Remember fmt failure on a per-buffer basis

### LINKED ISSUE #720: Regression in closing location list
The commit 8ef677a3625219758a63e1b6826c189a480b4edb introduced a regression where sometimes the location list is not closed even though it should be.

I'm not sure of all the conditions in which this can happen, but here's one way to repro it:

1. Open a valid .go file in vim. This is split A.
2. Open a new split (split B), and open a different .go file in split B.
3. In split B, type some syntax error into the Go code.
4. `:w` to save the buffer in split B. Now the location list should open in split B because gofmt sees a syntax error.
5. Now switch back to split A and save that buffer (which should not have a syntax error). The location list should still be open in split B.
6. Switch back to split B, fix the syntax error, and save. Now the location list is still open with the old error, even though it is fixed.

At this point, the only way to get rid of the location list is either to close it manually, or else to make a new syntax error, save, fix it, and save again.

I found the commit that caused the problem via bisecting.

cespare:
Fixes #720.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -43,8 +43,6 @@ if !exists("g:go_fmt_experimental")
     let g:go_fmt_experimental = 0
 endif
 
-let s:got_fmt_error = 0
-
 "  we have those problems : 
 "  http://stackoverflow.com/questions/12741977/prevent-vim-from-updating-its-undo-tree
 "  http://stackoverflow.com/questions/18532692/golang-formatter-and-vim-how-to-destroy-history-record?rq=1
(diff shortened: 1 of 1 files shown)


cespare:
I'm happy to rebase if necessary if you merge #700 first.

fatih:
Awesome. Yeah let me merge first #700. If there is any conflict you can rebase.

fatih:
@cespare seems like you need to rebase.

cespare:
Done :)

fatih:
Thanks @cespare :+1:  :)

PR #727, opened 2016-02-17
fallback for mkview

mattn:
related issue #691


fatih:
Good catch @mattn. I knew there would be a problem with this. I'm still not satisfied with using mkview, I believe it makes it slow but I need to benchmark the code.

PR #730, opened 2016-02-17
Allow GoRun to accept file arguments in neovim

nhooyr:
For some reason, we forgot about this feature in neovim. In vim you can
run `:GoRun main.go` so that other files in the same directory do not
interfere. E.g. if I have the main function defined in another file, a
bunch of random c files etc.


fatih:
Thanks @nhooyr :+1:

PR #735, opened 2016-02-22
add support for goimports -srcdir in Format

mhamrle:
"goimports -srcdir" fixes problems with vendoring.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -100,6 +100,17 @@ function! go#fmt#Format(withGoimport)
     if a:withGoimport  != 1 
         let command  = command . g:go_fmt_options
     endif
+    if fmt_command == "goimports"
+        let out = system("goimports --help")
+        if out !~ "-srcdir"
+            echohl WarningMsg
+            echomsg "vim-go: goimports does not support srcdir."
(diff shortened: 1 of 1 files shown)


wuman:
I can confirm that this PR works well.

fatih:
I was going to add this as this is broken if the user uses an old version of goimports. We need somehow warn the user to update the binary.

fatih:
I've tested it and it works well. But if you don't update your `goimports` binary it breaks vim-go. But I think it's ok to move on. This is going to be merged to master and it's not the stale version, so it's ok. Thanks @mhamrle 

Also let me put this commit here which introduces this new feature: https://github.com/golang/tools/commit/bf084ef7580ee99a5efa3086138c942aca4aefd4

mhamrle:
I added check if `goimports --help` contains `-srcdir` argument. If not the warning message is displayed and the `srcdir` arg is not used.

fatih:
@mhamrle I've thought about it but I think it would slow down. How did you do the check? Can you open another PR so we can talk about it ?

fatih:
@mhamrle we can speed it up with an intermediate cache value. So we only check once, if it's available we set `goimports_updated =1` and don't check it again till the user restarts vim.

PR #743, opened 2016-02-26
Redraw! after running Go program from Vim

gypsydave5:
Hi - hope you don't mind the pull request (even though it's really small!) - I've just been playing around with the plugin tonight (having a great time using it and learning Go), but I'd noticed that my vim session was coming up black after I'd run the (tiny) Go program I was writing.

I've thrown a `redraw!` at the end of the run command and it seems to fix it for me locally - thought I'd make you a small pull request before I went to bed.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -137,6 +137,7 @@ function! go#cmd#Run(bang, ...)
 
     let $GOPATH = old_gopath
     let &makeprg = default_makeprg
+    redraw!
 endfunction
 
 " Install installs the package by simple calling 'go install'. If any argument


nerakhon:
I tested this and it solves the problem in my case.

nerakhon:
Actually I was thinking about this and it seems to me a better result would be if after the GoRun the output was waiting for keypress like if running via ! go run %, but that's probably not a subject matter of this bug and an enhancement.

gypsydave5:
I was thinking about that too @nerakhon - came to the conclusion just now that it would be nice if we could redirect the output of run into a scratch buffer... but like you say, that's more of a different feature.

bhcleek:
@nerakhon I think your suggestion about behaving just like `!go run %` is exactly right; that's what vim-go used to do before I messed it up in #700. #748 should restore the previous behavior of showing the output and prompting the user to hit enter before coming back to the buffer.

fatih:
I think this is now fixed with https://github.com/fatih/vim-go/pull/748 fixes via @bhcleek . Can you test it please @gypsydave5 ? We don't need to call redraw anymore as the underlying fix is resolved.

gypsydave5:
Ah nice one! Definite fix - thanks :+1: :fireworks:

Issue #732, opened 2016-02-19
Gofmt-on-save closes folds if foldlevel is >0


cespare:
This is another side effect of ff56a315974f752ab3424ae750548a3a17277f9a.

I have `set foldlevel=99` in my vimrc so files open with folds open. After ff56a315974f752ab3424ae750548a3a17277f9a, saving causes my folds to be closed. Repro steps:

1. Open a .go file
2. Fold a section by selecting a visual block and then `zf`
3. Immediately expand the fold (e.g. `za`)
4. Save -> fold is closed

The problem happens if I set `foldlevel` to anything greater than 0 in my vimrc.

cespare:
The symptom is actually identical to #502, which ff56a315974f752ab3424ae750548a3a17277f9a was solving :P

Issue #728, opened 2016-02-17
Saving sometimes changes scroll position


cespare:
Some very recent change (ff56a315974f752ab3424ae750548a3a17277f9a?) causes the buffer to be scrolled far up or down after a save (gofmt). I don't yet have good repro instructions (it only happens some of the time) so I will update this bug more later.

fatih:
Right, that change is too big and honestly I didn't like that much. Maybe we should add it behind a setting like `go_fmt_save_folds`, which is by default disabled. I think the formatting stuff is the most problematic piece of code in vim-go. And it's due Vim's architecture and not having a set of primitives for these kind of stuff. Let me know if you find something.

cespare:
After updating, I can no longer reproduce this bug. Perhaps ba96e1f1fed4317313bb4852293dc1cc5b005a6b fixed it? But when trying to get to the bottom of it, I was unable to find a prior commit at which the problem reproduced either.

Maybe it was some bad state of the vim-go repo on this computer. Or maybe it's just hard to reproduce reliably and it can still happen. I'll reopen if I see it again.

cespare:
Argh, it just happened again. I'll have to spend more time trying to figure out how to reproduce it. I wrote code for a couple of hours without it happening, but yesterday it was happening on almost every save.

cespare:
I still don't have exact repro steps, but I'm increasingly confident that ff56a315974f752ab3424ae750548a3a17277f9a caused the issue.

It's hard to track down the problem, because it's tied to some on-disk state. For example, here's the steps I took in my most recent attempt:

* I was able to reproduce the problem at tip by by adding some new code to my test file and saving.
* I checked out a commit from a few weeks ago -> the problem went away.
* Tried bisecting the commits, but the problem didn't come back.
* Even when I went back to tip, the problem had stopped happening.

I see that there's a new `views/` directory in my `.vim`, which is presumably caused by the `mkview!` calls in ff56a315974f752ab3424ae750548a3a17277f9a. I think the issue is likely to be caused by some particular state of the view file, due to the fact that the behavior carries across vim restarts.

Once I can get the problem to start happening again, I'm going to see if saving/restoring the view file when it's in a bad state can reliably reproduce the problem.

cespare:
Update: I still don't have repro steps, but the problem is not due to the on-disk state: it's the editor process itself that's in a bad state.

[...]

PR #745, opened 2016-02-29
Put mkview/loadview changes behind go_fmt_experimental

cespare:
For #728 and #732.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -52,13 +52,25 @@ endif
 "  this and have VimL experience, please look at the function for
 "  improvements, patches are welcome :)
 function! go#fmt#Format(withGoimport)
-    " save cursor position, folds and many other things
-    let l:curw = {}
-    try
-        mkview!
-    catch
+    if g:go_fmt_experimental == 1
+        " Using winsaveview to save/restore cursor state has the problem of
+        " closing folds on save:
+        "   https://github.com/fatih/vim-go/issues/502
+        " One fix is to use mkview instead. Unfortunately, this sometimes causes
+        " other bad side effects:
+        "   https://github.com/fatih/vim-go/issues/728
+        " and still closes all folds if foldlevel>0:
+        "   https://github.com/fatih/vim-go/issues/732
+        let l:curw = {}
+        try
+            mkview!
+        catch
+            let l:curw=winsaveview()
+        endtry
+    else
+        " Save cursor position and many other things.
         let l:curw=winsaveview()
-    endtry
+    endif
 
     " Write current unsaved buffer to a temp file
     let l:tmpname = tempname()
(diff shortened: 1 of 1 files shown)


autoload/go/fmt.vim
```diff
@@ -52,13 +52,25 @@ endif
...
+        "   https://github.com/fatih/vim-go/issues/728
+        " and still closes all folds if foldlevel>0:
+        "   https://github.com/fatih/vim-go/issues/732
+        let l:curw = {}
+        try
+            mkview!
+        catch
+            let l:curw=winsaveview()
+        endtry
+    else
```
fatih: I think what we want to do is using `winsaveview()` in normal mode and `mkview`. Otherwise we lose cursor position and also doesn't align with the old version.

cespare: I'm sorry, I don't understand.

When you say "normal mode", are you talking about vim's normal mode, or `g:go_fmt_experimental == 0`?

fatih: Sorry, I mean the latter. Where `g:go_fmt_experimental` is disabled, set to `0`.

cespare: If `g:go_fmt_experimental == 1`, the code is exactly the same as master; if `g:go_fmt_experimental == 0`, that code path is exactly the same as before the mkview changes.

fatih: Right, but the original version was like this: https://github.com/fatih/vim-go/pull/691/files So we were saving the cursor position with `winsaveview`. This change is now putting it behind a flag. 

There is already a flag as you know, so we need to put `mkview` behind this: https://github.com/fatih/vim-go/blob/6c1cc914c35735768e54fc4ab0be900a9d62797b/autoload/go/fmt.vim#L64

So final state should be something like:

```
if experimental == 0 
  winsaveview()
else
  wundo...
  mkview...
endif

cespare: @fatih Sorry, I've read your message several times and I don't see the problem. What's wrong with the `else` branch?

You wrote

> This change is now putting it behind a flag.

but it's not behind a flag; it's in the `else` branch where `g:go_fmt_experimental != 1`.

fatih: @cespare yeah, it's all my fault due reading it wrong and missing the else condition (put facepalm here). Sorry for nagging here :)

cespare: @fatih It's possible you're missing that `winsaveview` is called in both the fallback case (for #727) and in the `g:go_fmt_experimental == 0` case.

cespare: No worries! It's a confusing amount of branching at this point

fatih: @cespare It's midnight here, probably I shouldn't review PR's at this time :) Sorry again didn't want to make feel you bad, it's all my fault. False interpretation.

PR #746, opened 2016-03-01
fmt: use shellescape instead of fnameescape

eliasgs:
The system command would break when a path like '/test/path (0)/ok' was
escaped with fnameescape. The shellescape function seems to handle
these cases better.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -127,7 +127,10 @@ function! go#fmt#Format(withGoimport)
         endif
 
         if exists('b:goimports_vendor_compatible') && b:goimports_vendor_compatible
-            let command  = command . '-srcdir ' . fnameescape(expand("%:p:h"))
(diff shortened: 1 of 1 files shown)


fatih:
Thanks for the fix @eliasgs. To use this we also should disable `shellescape`. We have a lot of pleace like that. One way of doing it changing your piece of code to this:

```
let ssl_save = &shellslash
set noshellslash
let command  = command . '-srcdir ' . shellescape(expand("%:p:h"))
let &shellslash = ssl_save
```

Let me know once you changed your PR to include these changes  and rebased it, so we can merge it :)

fatih:
Ping @eliasgs

eliasgs:
Hi @fatih,

Sorry about this late reply - I will incorporate and test the `noshellslash` as you outlined and update the pull request now:)

Thanks!

eliasgs:
@faith: the pull request is updated.

Issue #739, opened 2016-02-24
GoRun messes up terminal


nerakhon:
### Expected behavior

When issueing :GoRun I expect the go to output to standard output "bellow vim" and return to editor seeing the edited file as before executing this command.

### Actual behavior

The window has no content except cursor and this situation can be resolved by either restarting vim or issuing write (doesn't always help) or new tab commands

### Steps to reproduce:

Please create a reproducible case of your problem. Re produce it 
with a minimal `vimrc` with all plugins disabled and only `vim-go`
enabled:

1.

vim main.go

2.

```
package main

import "fmt"

func main() {
        fmt.Printf("Hello, world.\n")
}
```
3.
:GoRun

### Configuration

OS X 10.11

$GOPATH=~/.gobin

#### `vimrc` you used to reproduce:

```
set nocompatible              " be iMproved, required

call plug#begin('~/.vim/plugged')

Plug 'fatih/vim-go'

call plug#end()

filetype plugin indent on
```

#### vim version: 

version 7.4.1401 with Lua

#### vim-go version:

latest from github as of today

#### go version:

go version go1.5.2 darwin/amd64

fshiori:
I have the similar question on  MacVim.
Can't see any output below vim when execute GoRun.

makedonian:
I'm having the same issue with terminal vim. This didn't used to happen before. Running `:!go run %` still works as expected

nerakhon:
@makedonian I second that, same result with running go run via shell pipe.

makedonian:
There is a pull request here https://github.com/fatih/vim-go/pull/743 that the author says it should fix the issue (still doesn't work for me), maybe it's something as simple as that @fatih?

andrey-tpt:
Tested pull request #743 - works just fine for setup mentioned by @nerakhon (with the only difference as go1.6 vs go1.5.2)

gypsydave5:
Yeah, `redraw!` is a handy command to run when Vim comes back all messed up; just run it as a Vim command as a stopgap :smile:

makedonian:
@gypsydave5 my issue is that before, when you ran :GoRun it would show the output of the program until the next keypress. This is what's happening right now:

![test](https://cloud.githubusercontent.com/assets/471787/13479678/e6f8c7d2-e095-11e5-88e0-e22668d38fba.gif)

I'll try to go back few versions to see where it broke

bhcleek:
[...]

PR #748, opened 2016-03-03
show output of :GoRun

bhcleek:
Fixes regression introduced in #700, as documented on #739 and #743.


fatih:
Thanks @bhcleek, much appreciated. I'm busy lately but this is indeed the fix.

pjaleeuwen:
Thank you so much, this was bugging the hell outa me. ^^

Issue #667, opened 2015-12-23
Add support for Vim tag stack to GoDef?


gavrie:
I'm a longtime happy user of vim-go, but one thing bugs me: When jumping to a definition with `:GoDef`, I often want to look around and then jump back to the original location. While it's possible to use `CTRL-O` for that, this often requires multiple jumps.

When programming in C, I use `ctags` and Vim allows going to a tag definition with the `CTRL-]` mapping, and then jump back with `CTRL-T`. This keeps a tag stack that can be viewed with `:tags`, so it is easy to see the flow of the code.

It would be awesome if this were possible in vim-go: Allow going to a definition with `CTRL-]` and adding the location to the tag stack so that all the rest just works.

fatih:
Hi @gavrie This looks like a nice improvement we could do. I'm looking into it. Thanks for the feedback :)

alaska:
From my initial research, modifying the tag stack is a tricky business.  You might be better off making your own jump stack and remapping `C-T`, `C-]`, and creating your own `:Gotags` to list them. For a reference on how this might be done, check out https://github.com/vim-scripts/VIntSearch/blob/master/autoload/VIntSearch.vim#L258

gavrie:
@alaska That sounds just great!

fatih:
Yeah, my initial research didn't lead me to any important direction. There are no public ip's to create tag stack unfortunately. I'm closing this as I don't have the time to do any kind of hacks to support it. Thanks @gavrie for opening it though.

alaska:
I'll take a look and submit a PR if I can get anything done in a reasonable time.

spsholleman:
+1 for the request

fatih:
This is now in master. Though I'm still going to work on this for the `guru` branch and the final commands might change. Fyi.

gavrie:
Very cool! Thanks @alaska and @fatih. This will make the user experience much better.

jakebasile:
Awesome feature, but I wanted to point out that just in case anyone ends up here because their `<C-t>` maps stop working with error message `godef: tag stack empty`, you can disable this feature with `let g:go_def_mapping_enabled=0`

I have NERDTree mapped to `<C-t>` and [this line](https://github.com/fatih/vim-go/commit/80f6cf8fa443b259af8f90762ec51fadde57873e#diff-0ad1676dd1d31e7de174d292cadf0cecR34) overrode that map.

[...]

PR #756, opened 2016-03-10
:GoDef custom Tag Stack support (#667)

alaska:
This PR adds a godef-driven jump stack to vim-go along with associated bindings and commands that are roughly analogous to the way vim's built in tags support works.  `:help :GoDef` and `:help godef-stack` for docs.

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -104,12 +104,6 @@ function! s:godefJump(out, mode)
         return
     endif
 
-    " Don't jump if we're in a modified buffer
-    if getbufvar(bufnr('%'), "&mod")
-        call go#util#EchoError("No write since last change")
-        return
-    endif
-
     let parts = split(a:out[0], ':')
 
     " parts[0] contains filename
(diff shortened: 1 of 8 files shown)


fatih:
First of all thanks for this great feature @alaska. I really like it and thinks it's awesome. Before I review the code I've tested the functionality and noted the following bugs and things that needs to be improved:

### 1
When I use the following mapping (`au FileType go nmap <Leader>v <Plug>(go-def-vertical)`),  I get this error:

image

This doesn't happen when I use `gd`

### 2

When I type `:GoDefPop` (without any argument) I get this error:

image

But according to the help text it should be equivalent to `GoDefPop 1` which is not the case. Note that `:GoDefPop 1` works as indented.

### 3
When I type `:GoDefJump` and use it without any arguments (just hit enter) I get this error:

image

When I type a given number it works as intended

### 4
When I type `:GoDefJump` and press a non available number (say 3 for a list of two entries), it gives an error like:

image

It should instead close the window and return an error with our utility function [go#util#EchoError](https://github.com/fatih/vim-go/blob/master/autoload/go/util.vim#L95)

### 5
When I type `:GoDefPop 300` (a very high number, just to show an invalid number), it gives the following: 

image

This should be again more user friendly, saying something like `vim-go: invalid stack number for godef. Please execute :GoDefJump to see the valid entries`

autoload/go/def.vim
```diff
@@ -84,17 +93,29 @@ function! s:godefJump(out, mode)
...
+
+	" Echo the godef error if we had one.
+	if location =~ 'godef: '
+		let gderr=substitute(location, go#util#LineEnding() . '$', '', '')
+		echom gderr
+		" Don't jump if we're in a modified buffer
+	elseif getbufvar(bufnr('%'), "&mod")
+        echohl ErrorMsg
+		echo "No write since last change"
+        echohl None
```
fatih: We should use our own error message functions here: https://github.com/fatih/vim-go/blob/master/autoload/go/util.vim#L95 

Also is it safe to return here if instead of continuing further?

autoload/go/def.vim
```diff
@@ -114,11 +135,122 @@ function! s:godefJump(out, mode)
...
 	end
-	let &errorformat = old_errorformat
+endfunction
+
+function! go#def#StackPrint()
+    if len(w:go_stack) == 0
+        echohl ErrorMsg
+        echo "godef stack empty"
+        echohl None
+        return
```
fatih: Let's use our helper error function here

autoload/go/def.vim
```diff
@@ -114,11 +135,122 @@ function! s:godefJump(out, mode)
...
+        " Remove anything newer than the current position, just like basic
+        " vim tag support
+        if w:go_stack_level == 0
+            let w:go_stack = []
+        else
+            let w:go_stack = w:go_stack[0:w:go_stack_level-1]
+        endif
+
+		" increment the stack counter
+		let w:go_stack_level = w:go_stack_level + 1
```
fatih: Seems like the indentation is wrong

fatih: Also please use `let w:go_stack_level += 1`

autoload/go/def.vim
```diff
@@ -114,11 +135,122 @@ function! s:godefJump(out, mode)
...
+        echon entry["line"] . " col " . entry["col"]
+        echohl None
+        echon "|"
+        echon entry["ident"] . "\n"
+        let i += 1
+    endwhile
+    if w:go_stack_level == i
+        echo ">"
+    endif
+endfunction
```
fatih: I think this is the wrong approach. Check out https://github.com/fatih/vim-go/blob/master/autoload/go/ui.vim#L4 how we create a custom buffer with a custom filetype and just append the content into it. It's more scalable and maintainable.

autoload/go/def.vim
```diff
@@ -114,11 +135,122 @@ function! s:godefJump(out, mode)
...
+
+function! go#def#StackPop(numPop)
```
fatih: variable names in vim should be in form of `num_pop`

autoload/go/def.vim
```diff
@@ -114,11 +135,122 @@ function! s:godefJump(out, mode)
...
+        " jump
+        call s:goToFileLocation(target["file"], target["line"], target["col"])
+    else
+        echohl ErrorMsg
+        echo "invalid godef stack location"
+        echohl None
+    endif
+endfunction
+
+function! s:goToFileLocation(...)
```
fatih: The function signature should be explicit about the given arguments. Like: `s:goToFileLocation(a:file, a:line, a:col)`

doc/vim-go.txt
```diff
@@ -14,15 +14,16 @@
...
-  4. Mappings.....................................|go-mappings|
-  5. Text Objects.................................|go-text-objects|
-  6. Functions....................................|go-functions|
-  7. Settings.....................................|go-settings|
-  8. Troubleshooting..............................|go-troubleshooting|
-  9. Credits......................................|go-credits|
+  1.  Intro........................................|go-intro|
+  2.  Install......................................|go-install|
+  3.  Commands.....................................|go-commands|
+  4.  GoDef Stack..................................|godef-stack|
```
fatih: We should not have a separate section for this. This should be under `Commands` just below the current `:GoDef` commands so people can follow it.

alaska:
Great comments.  I'll get them resolved.  The one thing I want to mention is that, for your number 5, that's how the default :tags support behaves when it bottoms out.  It goes to the first level and issues a message.  Even the error message is identical (minus the E:## bit).  If you want to change the message to "invalid location", then I'm assuming you also don't want it to jump?  That's fine if this is the case, but I want to be clear if it's going to deviate from the default.  In the meantime, I'll be fixing everything else :)

fatih:
@alaska no we shouldn't jump to an invalid and also not assume it's going to be the last item. Just it should be a more user friendly error.

alaska:
In that case, I think I'll replace it with the same functionality as jump and be done with it

autoload/go/def.vim
alaska: I've fixed the error, and I'll exchange the echom for the util version as well. Returning is unnecessary, though, unless I've made a huge mistake, as all of the work is done in the else clause.  If you'd like it to be more go-y, I'd be happy to do early returns.  I'd probably put the modified check first in that case.

autoload/go/def.vim
alaska: Fixed the go_stack_level.  Would I be okay to do a :retab as a final step to fix the indentation?  What's the format you're using?

autoload/go/def.vim
fatih: Please return early whenever you can, it makes the code simpler and easier to maintain.

alaska:
How does this look so far?  The UI integration took the longest and required the deepest modifications, but the result is much cleaner.  Obviously, it needs some more testing, and I will continue to push changes if I find any issues.  Also, I'm not terribly keen on having the current line and cursor highlighted in the non-interactive mode, and would like to change that to behave more light a pager.  Thoughts/feedback?

autoload/go/def.vim
```diff
@@ -15,7 +15,9 @@ endf
 " modified and improved version of vim-godef
 function! go#def#Jump(...)
 	if !len(a:000)
-		let arg = "-o=" . go#util#OffsetCursor()
+		" gives us the offset of the word, basically the position of the word under
+		" the cursor
+		let arg = s:getOffset()
```
fatih: This is wrong, you should use `go#util#OffsetCursor()`

fatih: I think this comes from a wrong merge. Can you just rebase and squash your commits into one commit? That would fix these kind of issues. I can then re-check your PR, as right now I can't use it.

alaska: Thanks.  I think that's from 0b86d11.  I'll open another PR.

alaska:
Closing this in favor of the merged pr.

PR #758, opened 2016-03-10
alternate.vim: handle empty buffer names and general improvements

nhooyr:
Also throughout the code I see many raw calls to print error messages. We should move the rest of the code to use use `go#util#EchoError()`.


fatih:
Thanks @nhooyr :+1:

PR #759, opened 2016-03-10
goinfo improvements

nhooyr:
`Info()` will not echo `PANIC PANIC PANIC` if it's called from the
autocommand.

@peterhellberg what do you think?

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -153,9 +153,12 @@ function! go#complete#GetInfo()
     return go#complete#GetInfoFromOffset(offset)
 endfunction
 
-function! go#complete#Info()
+function! go#complete#Info(auto)
+    " auto is true if we were called by g:go_auto_type_info's autocmd
     let result = go#complete#GetInfo()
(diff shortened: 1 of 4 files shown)


autoload/go/complete.vim
```diff
@@ -153,9 +153,9 @@ function! go#complete#GetInfo()
     return go#complete#GetInfoFromOffset(offset)
 endfunction
 
-function! go#complete#Info()
+function! go#complete#Info(auto)
     let result = go#complete#GetInfo()
-    if !empty(result)
+    if !empty(result) && (!a:auto && result ==# "PANIC PANIC PANIC")
```
fatih: Can you please add a comment here what `auto` means and what `Panic Panic Panic` means? (I know what it is, but we have to much logic here and if someone else reads the code there is zero information)

nhooyr: that if condition didn't even work properly. I updated it now.

fatih:
Thanks @nhooyr

PR #762, opened 2016-03-16
refactored duplicate offset functions

### LINKED ISSUE #719: Remove and refactor duplicate offset functions
Right now there are several places in vim-go that parses the byte offset of the position under the cursor. The ones I could see are below. And what interesting is that each of them has it's own way of implementing it:

https://github.com/fatih/vim-go/blob/master/autoload/go/complete.vim#L90
https://github.com/fatih/vim-go/blob/master/autoload/go/rename.vim#L67
https://github.com/fatih/vim-go/blob/master/autoload/go/oracle.vim#L65
https://github.com/fatih/vim-go/blob/master/autoload/go/def.vim#L67

Let us create a single function under `util.vim` with the function signature: `go#util#OffSet()`, remove all others and use this new function.

nhooyr:
Closes #719

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -96,11 +96,12 @@ fu! s:gocodeAutocomplete()
     return result
 endf
 
-function! go#complete#GetInfoFromOffset(offset)
+function! go#complete#GetInfo()
+    let offset = go#util#OffsetCursor()+1
     let filename = s:gocodeCurrentBuffer()
     let result = s:gocodeCommand('autocomplete',
                 \ [s:gocodeCurrentBufferOpt(filename), '-f=godit'],
-                \ [expand('%:p'), a:offset])
+                \ [expand('%:p'), offset])
     call delete(filename)
 
     " first line is: Charcount,,NumberOfCandidates, i.e: 8,,1
@@ -134,11 +135,6 @@ function! go#complete#GetInfoFromOffset(offset)
     endif
 
     return ""
-endfunction
-
-function! go#complete#GetInfo()
-    let offset = go#util#OffsetCursor()+1
-    return go#complete#GetInfoFromOffset(offset)
 endfunction
 
 function! go#complete#Info(auto)
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -87,23 +87,11 @@ fu! s:gocodeCurrentBufferOpt(filename)
     return '-in=' . a:filename
 endf
 
-fu! go#complete#gocodeCursor()
-    if &encoding != 'utf-8'
-        let sep = &l:fileformat == 'dos' ? "\r\n" : "\n"
-        let c = col('.')
-        let buf = line('.') == 1 ? "" : (join(getline(1, line('.')-1), sep) . sep)
-        let buf .= c == 1 ? "" : getline('.')[:c-2]
-        return printf('%d', len(iconv(buf, &encoding, "utf-8")))
-    endif
-
-    return printf('%d', line2byte(line('.')) + (col('.')-2))
-endf
-
 fu! s:gocodeAutocomplete()
     let filename = s:gocodeCurrentBuffer()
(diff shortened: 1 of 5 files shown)


autoload/go/complete.vim
```diff
@@ -148,11 +137,6 @@ function! go#complete#GetInfoFromOffset(offset)
     return ""
 endfunction
 
-function! go#complete#GetInfo()
-    let offset = go#complete#gocodeCursor()+1
-    return go#complete#GetInfoFromOffset(offset)
-endfunction
```
fatih: This is a public function that people use, for example to plugin into statusline or use for their own functions.

fatih:
Hi @nhooyr, did you also checked out which implementation is the correct one? Or are they really all the same?

autoload/go/def.vim
```diff
@@ -65,18 +65,7 @@ endfunction
...
-	else
-		let c = pos[1]
-		let buf = line('.') == 1 ? "" : (join(getline(1, pos[0] - 1), go#util#LineEnding()) . go#util#LineEnding())
-		let buf .= c == 1 ? "" : getline(pos[0])[:c-2]
-		let offs = len(iconv(buf, &encoding, "utf-8"))
-	endif
-
-	let argOff = "-o=" . offs
-	return argOff
+	return "-o=" . go#util#OffsetCursor()
```
fatih: I think we should remove `s:getOffset()` and change `def.vim` to make it that it uses this directly if possible.

autoload/go/util.vim
```diff
@@ -85,6 +85,21 @@ function! go#util#Shelllist(arglist, ...)
...
 
+" Returns the byte offset for line a:l and column a:c
+function! go#util#Offset(l, c)
+    if &encoding != 'utf-8'
+        let buf = a:l == 1 ? '' : (join(getline(1, a:l-1), "\n") . "\n")
+        let buf .= a:c == 1 ? '' : getline('.')[:a:c-2]
+        return len(iconv(buf, &encoding, 'utf-8'))
+    endif
+    return line2byte(a:l) + (a:c-2)
+endfunction
```
fatih: I think the newline ending are different for fileformat dos. For example in `complete.vim`'s offset implementation there was a line:

```
 let sep = &l:fileformat == 'dos' ? "\r\n" : "\n"	
```

that would check for that. Can you please go over the offset functions and also add these kind of edge cases into this funciton ?

autoload/go/util.vim
```diff
@@ -85,6 +85,21 @@ function! go#util#Shelllist(arglist, ...)
     endtry
 endfunction
 
+" Returns the byte offset for line a:l and column a:c
+function! go#util#Offset(l, c)
```
fatih: just rename the arguments to `line` and `col` and change the doc to reflect it: `Returns the byte offset for the given line and column`

nhooyr:
@fatih all are the exact same as far as I can tell.

autoload/go/complete.vim
nhooyr: The function is still there, I renamed the top one.

autoload/go/util.vim
nhooyr: I forgot, my bad. We have a function in `util.vim` just for this.

autoload/go/def.vim
nhooyr: will do.

nhooyr:
updated based on feedback.

fatih:
Thanks @nhooyr I need to test this myself, If everything works as intended, I'll merge it.

nhooyr:
@fatih ok. here are the differences, the ones in rename.vim and oracle.vim were the exact same. They were copies of the `go#complete#gocodeCursor` without the special case newline line  and instead took the length/column as arguments. The last one in def.vim is the exact same as well but instead of using `line()` or `col()` it uses `getpos()`

fatih:
Thanks @nhooyr , great refactor, much appreciated :+1:

PR #776, opened 2016-03-25
Add custom godef jump stack (fixes #667)

### LINKED ISSUE #667: Add support for Vim tag stack to GoDef?

alaska:
Unified and updated PR

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -29,13 +29,17 @@ function! go#def#Jump(...)
 	let $GOPATH = go#path#Detect()
 
 	let fname = fnamemodify(expand("%"), ':p:gs?\\?/?')
-	let command = bin_path . " -f=" . shellescape(fname) . " -i " . shellescape(arg)
+	let command = bin_path . " -t -f=" . shellescape(fname) . " -i " . shellescape(arg)
 
 	" get output of godef
 	let out = s:system(command, join(getbufline(bufnr('%'), 1, '$'), go#util#LineEnding()))
 
+	" First line is <file>:<line>:<col>
+	" Second line is <identifier><space><type>
+	let godefout=split(out, go#util#LineEnding())
+
 	" jump to it
-	call s:godefJump(out, "")
+	call s:godefJump(godefout, "")
 	let $GOPATH = old_gopath
 endfunction
 
@@ -52,12 +56,17 @@ function! go#def#JumpMode(mode)
 	let $GOPATH = go#path#Detect()
 
 	let fname = fnamemodify(expand("%"), ':p:gs?\\?/?')
-	let command = bin_path . " -f=" . shellescape(fname) . " -i " . shellescape(arg)
+	let command = bin_path . " -t -f=" . shellescape(fname) . " -i " . shellescape(arg)
 
 	" get output of godef
 	let out = s:system(command, join(getbufline(bufnr('%'), 1, '$'), go#util#LineEnding()))
 
-	call s:godefJump(out, a:mode)
+	" First line is <file>:<line>:<col>
+	" Second line is <identifier><space><type>
+	let godefout=split(out, go#util#LineEnding())
+
+	" jump to it
+	call s:godefJump(godefout, a:mode)
 	let $GOPATH = old_gopath
 endfunction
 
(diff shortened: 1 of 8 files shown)


alaska:
How does this look so far? The UI integration took the longest and required the deepest modifications, but the result is much cleaner. Obviously, it needs some more testing, and I will continue to push changes if I find any issues. Also, I'm not terribly keen on having the current line and cursor highlighted in the non-interactive mode, and would like to change that to behave more light a pager. Thoughts/feedback?

fatih:
Hi @alaska 

I was absent on the weekend. I'll look into this soon!

autoload/go/def.vim
```diff
@@ -71,41 +80,191 @@ function! s:godefJump(out, mode)
...
+        if bufloaded(fileName) == 0
+            tab split
+        endif
+    elseif a:mode  == "split"
+        split
+    elseif a:mode == "vsplit"
+        vsplit
+    else
+        " We didn't get a mode, so we start the process of pushing to the
+        " jumpstack
```
fatih: So does it mean when I open a split view, we'll not going to use the jumpstack ?

autoload/go/ui.vim
```diff
@@ -23,37 +28,60 @@ function! go#ui#OpenWindow(title, content)
     else
         exe 'resize ' . len(a:content)
     endif
```
fatih: We should resize the window to the length of the content. I remember making it here to defaulting to `10`. But I think we can for now remove that limitation and make it in the future optional. So these lines:

```
    " Keep minimum height to 10, if there is more just increase it that it
    " occupies all results
    let buffer_height = 10
    if len(a:content) < buffer_height
        exe 'resize ' . buffer_height
    else
        exe 'resize ' . len(a:content)
    endif
```

should be changed to:

```
    " Resize window to content lenght
    exe 'resize ' . len(a:content)
```

doc/vim-go.txt
```diff
@@ -211,11 +211,67 @@ COMMANDS                                                          *go-commands*
...
+
+    Jumps to a given location in the jump stack, retaining all other entries.
+    If no argument is given, it will print out an interactive list of all items
+    in the stack, from which you can select one, if you choose.
+
+    For details on the format of this list, see |:GoDefStack|.
+
+    Jumps to non-existent entries will print an informative message, but are
+    otherwise a noop.  As with |:GoDefPop|, any :GoDef command after a jump
+    will replace the current entry and remove all newer entries.
```
fatih: I think we can unify `:GoDefJump` and `:GoDefStack`. No need to pollute the command namespace. They are essentially the same, but one does jump and the another not. I suggest we remove `:GoDefStack` completely. If someone wants to see the stack they can use `:GoDefJump`. Thoughts ?

autoload/go/def.vim
alaska: That's correct. I was trying to avoid inconsistent behavior, but I have just thought of a better way to do this that will push the location AFTER the jump, so that each new window gets its own stack that starts from the first location visited.

doc/vim-go.txt
alaska: I agree.  They can always `:q`, `q`, `<ESC>` or select the current entry.  What about the cursor? I would like to turn it off so that it is more like a menu. Thoughts on that?

alaska:
All requested changes have been made. In this version, the jumpstack starts new in new splits/tabs with one entry, which is the location you left.  Do you think this is the proper behavior, or should the stack be empty?

fatih:
Hi @alaska 

This seems to be working well. I'm using it now. Probably going to merge it today/tomorrowy and then will do some changes to make it compatible with the new `guru` changes. Thanks for all the work, this all looks great :+1:

alaska:
Thanks!  I'm still on the fence about whether the stack should be empty in a new split/tab or not.  I'm leaning towards not, since the location you jumped from would still be in the window you left, and popping would then leave you with two splits on the same buffer, but I can see how some folks might find it useful.

fatih:
About that. I think we need a command to clear the stack. Let's do it manual. Basically we should be able to set the entry point ourselves. Best way is doing it like:

* `:GoDefJump` should be renamed to `:GoDefStack`
* `:GoDefStackClear` is a new command that clears the stack, with an appropriate mapping

The reason is that split or tab is not a boundary. For example I do a lot `ctrl+w v` which opens a new split for the same file. However I still would like to use the previous stack for that file. It's a little bit trickt. To solve tricky problems, we should just provide primitives (such as `:GoDefStackClear`) and then let the user do the rest.

fatih:
There is a weird error. Suppose you called `:GoDef` 3-4 times, after that if you call `:GoDefJump` and you press the `upper arrow` case, it tries to insert the character `A`. I'm not sure if it's my settings or not, but looking into it.

fatih:
@alaska I'm merging this and going to open another PR inside the guru pr to fix some of the style changes and co. That would make it easier for us. Thanks a lot again for the idea and working on this :+1:

PR #780, opened 2016-03-31
display test target for go#cmd#Test

rxwen:
Show test target in vim status bar, so that we have a clear idea of what's being tested.
screen shot 2016-03-31 at 2 52 25 pm

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -208,7 +208,7 @@ function! go#cmd#Test(bang, compile, ...)
     if args[1] == "-run"
         let target = args[2]
     else
-        let target = expand("%:t")
+        let target = expand("%:h")
     endif
 
     if a:compile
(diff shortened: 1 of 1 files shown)


fatih:
Hi @rxwen 

I'm not sure we need this. You run just `:GoTest` which in that case runs for all packages. What's the gain we get this from here?

rxwen:
Hi @fatih,

The situation is I often need to run a single test case within vim, so that I can verify what I'm working now very quickly. I've setup a keymap to trigger :GoTestFunc.
With the target being shown, I'm sure that the case just run is exactly the one I intended to run.

I'm not sure if other people work in similar way. But I think it doesn't hurt to show more information.

rxwen:
@fatih 

I checked the code carefully. It seems my patch isn't correct, regardless of if this feature is useful or not.
I'll close it.

PR #782, opened 2016-04-02
Godoc improvement

### LINKED ISSUE #332: GoDoc doesn't properly find documentation
I have vim-go fully installed and my `$GOPATH` contains everything needed for a full build of my package. It does reference a few different directories though.

When I try to run `:GoDoc` on anything other than standard package items I receive:

```
2015/02/24 15:07:24 open /usr/local/go/src/db.Db: no such file or directory
```

Where `db.Db` is the package of the item I was hovering over.

Additionally, when running `:GoDocBrowser` I get referred to `https://godoc.org/db.Db#Select` which of course does not actually exist. Did I miss something that will give GoDoc the information I need? Is there a way to refer it to a running instance of `GoDoc -http`

Let me know if I can provided more helpful information about my setup. Thank you

fatih:
Fixes #332 

This uses @zmb3's new excellent tool called `gogetdoc`. It solves most of our problem here. Just call `:GoDoc` and it should show the documentation for the identifier under the cursor. Hit `esc` or `enter` to close the window easily.


jzacsh:
Nice, works wonderfully! Only thing is the new doc window is strangely small compared to the (50%?) height it used to have. I think this is intentional, as it seems full docs aren't listed anymore. Is this possible to turn on? (ie: open doc on a method, show that method _in_ context on its package's full doc, as before?)

zmb3:
No, gogetdoc doesnt work this way.  It gets the doc right from the source code, where it isn't grouped by package.

fatih:
@jzacsh unfortunately it breaks the old way of showing the doc. The other way could be getting the package name and then pass it to `go doc`. This might be another solution.

fatih:
Also there is to explore the docs: https://github.com/garyburd/go-explorer 

I think I have to re evaluate how we're going to want to explort go documentation in the future. Sometimes just getting a look at the current identifiers documentation is what we want. But sometimes we want to have a full disclosure to all documentation, to get a feeling what the remaining exported identifiers are, how the package can be used etc...

So we have two behavior of how we want to explore documentation:

1. Checking/showing an identifier 
2. Exploring the documentation of a package

Based on this I think we should have the following commands:

1. `:GoDoc`: the current approach which will show a single documentation of the underlying identifier
2. `:GoDocExplore`: this will be used to explore a package and see the full documentation (so basically https://github.com/garyburd/go-explorer)
3. `:GoDocBrowser`: we have this already, but it's flawed and doesn't work as intended. This opens the documentation in godoc.org. We also should re-evaluate this for private packages and and co.

Any feedback is welcome, I'm happy to discuss this. 

/cc @garyburd @zmb3

jzacsh:
interesting, thanks for expanding @fatih!

I think I only unconsciously expect full-doc view because I'm _used_ to a `C` file where Ctrl+K opens the full man page for me.

To clarify: no.2 the thing that currently happens (_when_ an identifier is easily found), right? Except that we actually scroll/jump down to the method's doc *inside* the full package doc? This is a nice mixture of both package-doc and method-doc.

fatih:
> To clarify: no.2 the thing that currently happens (when an identifier is easily found), right? Except that we actually scroll/jump down to the method's doc inside the full package doc? This is a nice mixture of both package-doc and method-doc.

This is the old approach, and will be replaced by the new approach. It's just for showing the documentation, like `go doc`. If you want to explore it' you have to use `:GoDocExplore` (which is not implemented and is the functionality provided via go-explorer).

PR #785, opened 2016-04-03
Coverage overlay in buffers

t-yuki:
As suggested by #686 , I've merged `vim-go-coverlay` repository simply and modified several conflicts.

I'd ask these points to merge:

* Should we merge command mappings?
* Should we squash it?
* It includes testsuite using VimFlavor.
 * Should we remove it or use another test tool?
* It includes several ugly hacks to implement overlay.
 * Colorize lines using `matchadd`
* Known Problem
 * No way to customize colors (requested in https://github.com/t-yuki/vim-go-coverlay/issues/4 )
 * It doesn't redraw all open splits so non-active splits will not be colored until it gets the focus (reported in https://github.com/t-yuki/vim-go-coverlay/pull/3 )


ftplugin/go/coverlay.vim
```diff
@@ -0,0 +1,14 @@
...
+
+" Some handy plug mappings
+nnoremap <silent> <Plug>(go-coverlay) :<C-u>call go#coverlay#Coverlay(!g:go_jump_to_error)<CR>
+nnoremap <silent> <Plug>(go-clearlay) :<C-u>call go#coverlay#Clearlay()<CR>
+
+" coverlay
+command! -nargs=* -bang GoCoverlay call go#coverlay#Coverlay(<bang>0, <f-args>)
+command! -nargs=* GoClearlay call go#coverlay#Clearlay(<f-args>)
+
+" vim:ts=4:sw=4:et
```
fatih: Some notes:

1. Mappings should be moved to `mappings.vim`
2. Commands should be moved to `commands.vim`
3. Commands should be renamed to:
  * `:GoCoverage` -> `:GoCoverageBrowser`
  * `:GoCoverlay` -> `:GoCoverage`
  * `:GoClearlay` -> `:GoCoverageClear`

Our current command will not open the browser anymore, instead it will use this current implementation. However if people want they can still access them via `:GoCoverageBrowser`

fatih:
> Should we merge command mappings?

Yes, I've added notes. We're going to rename them and this implementation will replace the current command

> Should we squash it?

I would prefer to squash it

> It includes testsuite using VimFlavor. Should we remove it or use another test tool?

Right now we don't have any kinds of tests. But this is very high on my list soon. I have to make a research and then implement the tests. For now I think we can keep these tests and once we decide on the tests, I'll can change them.

> It includes several ugly hacks to implement overlay. Colorize lines using matchadd

That's ok. But I was thinking of changing it. We should be able to highlight the whole line if possible. I'll look into it now.

>  No way to customize colors (requested in t-yuki/vim-go-coverlay#4 )
It's ok, I don't want people to change it. It's an implementation deteail and people should not change it. The same applies for `go tool cover -html` mode. There is green and red, and you can't change the color.

> It doesn't redraw all open splits so non-active splits will not be colored until it gets the focus (reported in t-yuki/vim-go-coverlay#3 )

We can check this out, right now I'm not sure how this will affect but we probably can find a way

--

Excellent work btw, thank you very much including this :) I'm using coverage extensively, so I think this is a huge addition!

autoload/go/coverlay.vim
```diff
@@ -0,0 +1,174 @@
+if !exists("g:go_gopath")
+    let g:go_gopath = $GOPATH
+endif
+
+augroup plugin-go-coverlay
+    autocmd!
+    autocmd BufEnter,BufWinEnter,BufFilePost * call go#coverlay#draw()
+    autocmd BufWinLeave * call go#coverlay#clear()
+augroup END
```
fatih: This should be moved to our autogroup section in  file `plugin/go.vim`

autoload/go/coverlay.vim
```diff
@@ -0,0 +1,174 @@
...
+    let ret = {}
+    let ret.file = substitute(l, mx, '\1', '')
+    let ret.startline  = substitute(l, mx, '\2', '')
+    let ret.startcol = substitute(l, mx, '\3', '')
+    let ret.endline = substitute(l, mx, '\4', '')
+    let ret.endcol = substitute(l, mx, '\5', '')
+    let ret.numstmt = substitute(l, mx, '\6', '')
+    let ret.cnt = substitute(l, mx, '\7', '')
+    return ret
+endfunction
```
fatih: You can simplify it like:

```
function! go#coverlay#parsegocoverline(line)
    " file:startline.col,endline.col numstmt count
    let mx = '\([^:]\+\):\(\d\+\)\.\(\d\+\),\(\d\+\)\.\(\d\+\)\s\(\d\+\)\s\(\d\+\)'
    let tokens = matchlist(a:line, mx)
    let ret = {}
    let ret.file = tokens[1]
    let ret.startline  = tokens[2]
    let ret.startcol = tokens[3]
    let ret.endline = tokens[4]
    let ret.endcol = tokens[5]
    let ret.numstmt = tokens[6]
    let ret.cnt = tokens[7]
    return ret
endfunction
```

fatih:
Alright I couldn't stop myself. I've made it faster by using `matchaddpos` and changed the way we display  coverage. Checkout the gif:

screen recording 2016-04-03 at 11 11 pm

I'll merge your changes and then apply my patch (as I know what to do already). Let me finalize the latest bits on my part. Thanks again for the work, works really cool ;)

fatih:
@t-yuki I've opened https://github.com/fatih/vim-go/pull/786 with your commits squashed and included and my changes on top of it. I'm going to merge it. Thanks a lot for your work again! Please let me know if there is anything I can do.

t-yuki:
Thanks!

PR #786, opened 2016-04-03
Coverage overlay integration (vim-go-coverlay)

fatih:
This PR includes #785, mainly @t-yuki's excellent https://github.com/t-yuki/vim-go-coverlay plugin with additional improvements and integrations into the vim-go code base. Changes are:

* `:GoCoverage` does highlights the current source code instead of opening a browser. When called again it clears the highlighting.
* `:CoCoverageBrowser` opens the html annotated coverage result (previous `:GoCoverage`

Thanks for @t-yuki  excellent initial work!

Here is demo:

gocoverage-2


gonzaloserrano:
Nice work! 

I was thinking in a couple of UX suggestions:
- have a "toggle" command
- instead of changing the lines colors, change its "gutter" similar to how does it https://github.com/airblade/vim-gitgutter

Thanks anyway ;-)

fatih:
Hi @gonzaloserrano 

> have a "toggle" command

Yeap, I'm going to remove `:GoCoverageClear` and `:GoCoverage` will act as a toggle. I'm always trying to use a feature for a couple of days/weeks to get a sense how it should work and today I realized is that it should be indeed a toogle

>  instead of changing the lines colors, change its "gutter" similar to how does it

This is hard to do as I have to hack the sidebar. This is exactly the same as the HTML representation and I think it fits it better. So there will be only one opinionated way. 

Thanks for the feedback!

gonzaloserrano:
Sounds cool anyway, thanks a lot! :+1:

VimFlavor
```diff
@@ -0,0 +1 @@
+flavor 'fatih/vim-go', '~> 1.5'
```
t-yuki: This line should be removed since it means `This test depends on external plugin vim-go`

fatih: Good call. Btw I've changed the source code a lot and renamed the internal functions too. They will not work and probably will be broken. But I'm going to work on tests soon.

fatih:
I'm merging this. Tested it extensively and should be working very well. I've disabled supporting multiple files as it's not scalable. I prefer we go simple, let first be good at supporting the current buffer, later we can use the cached result for other buffers in the future. Please open an issue if you see bugs! Thanks.

PR #787, opened 2016-04-05
Add go_highlight_format_strings option for printf-style format strings

cespare:
I'd like to disable this feature, so I added an option.


fatih:
LGTM, thanks @cespare

PR #794, opened 2016-04-11
GoTestFunc: support testable examples

cuzelac:
Support godoc example tests ([described here](https://golang.org/pkg/testing/#hdr-Examples)) in `go#cmd#TestFunc()`


fatih:
Thanks @cuzelac :+1:

PR #795, opened 2016-04-13
Should be FileType

mattn:
```
$ vim foo.go
```
vim-go dosen't work for godef when opening file like above.


fatih:
This fails if you open an `:GoDef` in a new vertical split. I'll change it completely by removing the variable initialization here and moving it to `def.vim`, making it script only.

mattn:
okay, please.

PR #799, opened 2016-04-15
Handle windows paths

mattn:
When the line from godef is `C:\foo\bar\baz.go:123: descriptions`, `split(out, ":")` break the paths on windows.

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -36,7 +36,11 @@ endfunction
 function! s:jump_to_declaration(out, mode)
 	" strip line ending
 	let out = split(a:out, go#util#LineEnding())[0]
-	let parts = split(out, ':')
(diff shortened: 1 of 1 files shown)


autoload/go/def.vim
```diff
@@ -36,7 +36,11 @@ endfunction
 function! s:jump_to_declaration(out, mode)
 	" strip line ending
 	let out = split(a:out, go#util#LineEnding())[0]
-	let parts = split(out, ':')
+	if has('win32') || has('win64')
```
fatih: Can you use `if go#util#IsWin()` ?

fatih:
Thanks @mattn, hopefully with JSON support coming to Vim, it'll prevent problems like this.

mattn:
done

PR #801, opened 2016-04-15
use go#util#system(), go#util#shell_errors() instead of system(), v:shell_error

mattn:
errors from guru is ignored because when using vimproc. maybe having another issues with vimproc.


autoload/go/util.vim
```diff
@@ -37,6 +37,46 @@ function! go#util#IsWin()
...
+
+    return g:go#use_vimproc
+endfunction
+
+if s:has_vimproc()
+	let s:vim_system = get(g:, 'gocomplete#system_function', 'vimproc#system2')
+    let s:vim_shell_error = get(g:, 'gocomplete#shell_error_function', 'vimproc#get_last_status')
+else
+	let s:vim_system = get(g:, 'gocomplete#system_function', 'system')
+    let s:vim_shell_error = ''
```
mattn: storange indentation, will fix.

autoload/go/complete.vim
```diff
@@ -19,25 +19,6 @@ fu! s:gocodeCurrentBuffer()
...
-    endif
-    return call(s:vim_shell_error, [])
-endf
-
-fu! s:system(str, ...)
-    return call(s:vim_system, [a:str] + a:000)
-endf
-
 fu! s:gocodeShellescape(arg)
     if go#vimproc#has_vimproc()
```
fatih: This needs to be fixed as well, completion is broken with this.

mattn:
Fixed. But I need to check more.

mattn:
Do you prefer `go#util#System` or `go#util#Shellerror` ?

autoload/go/util.vim
```diff
@@ -37,6 +37,46 @@ function! go#util#IsWin()
...
+    let s:vim_system = get(g:, 'gocomplete#system_function', 'vimproc#system2')
+    let s:vim_shell_error = get(g:, 'gocomplete#shell_error_function', 'vimproc#get_last_status')
+else
+    let s:vim_system = get(g:, 'gocomplete#system_function', 'system')
+    let s:vim_shell_error = ''
+endif
+
+function! go#util#system(str, ...)
+    return call(s:vim_system, [a:str] + a:000)
+endfunction
```
fatih: Just had time due my new son :) So I just had a new idea and curious about your thought. What if we add a function that  returns both the output and shell error. A list that contains the following `[out, shell_error]` ? This way we could write code like this:

```
let [out, shell_error] = go#util#Run("command ...")
if shell_error != 0 
    return -1
endif

// continue with out
```

This would make it more safer to use and would emulate the way we use it in Go. Let me know what you think. We can do it in another PR btw :)

fatih:
> Do you prefer go#util#System or go#util#Shellerror ? 

Checkout my other comment. For this question `go#util#System` is good 👍

fatih:
Thanks @mattn 👍

PR #802, opened 2016-04-15
Clarify that mappings are enabled by default

sethvargo:
This clarifies in the README that none of the mappings are enabled by default. I completely glanced over the "here are some examples" and assumed these mappings were enabled by default. I was wrong.

Great work on the plugin! Loving it so far :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,10 +105,11 @@ After that just open the help page to see all commands:
 
     :help vim-go
 
-## Mappings
+## Example Mappings
 
 vim-go has several `<Plug>` mappings which can be used to create custom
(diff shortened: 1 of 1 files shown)


fatih:
Hi @sethvargo,

Yeah, I never add any mappings as they can interfere with user commands. The only default mappings are the ones that are already defined in Vim but not usable for vim-go. That enables us to override those mappings. Here are the ones that are default (note that these also can be disabled if wished :):

https://github.com/fatih/vim-go/blob/master/ftplugin/go.vim#L26-L55

Thanks for change

Issue #807, opened 2016-04-20
Feature Request: gotests support


itsjamie:
It would be great to be able to run :GoGenerateTest which would build the basic scaffolding for table-driven tests.

You can find the tool here: https://github.com/cweill/gotests

Not knowing what's possible with vim as far as determining context, the way I would imagine it working:
* When inside a function in normal mode, a :GoGenerateTest would generate tests for that single function
* When inside visual mode, a :GoGenerateTest would generate tests for any functions inside the selection block.

If it's not possible, or it's hard to determine context, then `GoGenerateTestFunc` to match with `GoTestFunc` would be great.

Thank you for the awesome tool. It's the first time I've used vim for development.

fatih:
Closing per my comment here: https://github.com/fatih/vim-go/pull/808#issuecomment-215553108 Thanks again for the feedback.

PR #808, opened 2016-04-21
Generate table tests using gotests tool via :GoTableTests

bradleyfalzon:
Extends vim-go to use https://github.com/cweill/gotests to generate table tests https://github.com/golang/go/wiki/TableDrivenTests as discussed in issue #807 (see also cweill/gotests#20)

Only supports generating table tests on the function under the cursor, i.e. does not add support for generating table tests for all selected functions when using visual mode.

Does not support refreshing buffers or tabs if the corresponding _test.go file is already open.

`:GoInstallBinaries` does not display the path `github.com/cweill/gotests/...` correctly when installing, which is the path gotests project expects you use to install it. Instead, this change references the path to the binary `github.com/cweill/gotests/gotests`. This should not be a problem and tested OK, but may need to be reconsidered if issues do arise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ disabled/enabled easily.
 * Integrated with the Neovim terminal, launch `:GoRun` and other go commands
   in their own new terminal. (beta)
 * Alternate between implementation and test code with `:GoAlternate`
+* Generate table tests using [gotests](https://github.com/cweill/gotests) with `:GoTableTests`
 
 ## Install
 
@@ -276,8 +277,8 @@ just a way to support vim-go's ongoing development directly. Thanks!
 ## Credits
 
 * Go Authors for official vim plugins
-* Gocode, Godef, Golint, Guru, Goimports, Gotags, Errcheck projects and
-  authors of those projects.
+* Gocode, Godef, Golint, Guru, Goimports, Gotags, Errcheck, Gotests projects
+  and authors of those projects.
 * Other vim-plugins, thanks for inspiration (vim-golang, go.vim, vim-gocode,
   vim-godef)
 * [Contributors](https://github.com/fatih/vim-go/graphs/contributors) of vim-go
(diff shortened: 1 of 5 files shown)


bradleyfalzon:
Reopening buffers/tabs may require existing buffer to be saved first, gotests executed and then buffer refresh contents, further complications with tabs may ensue. However, this should be done at a later stage. Note, gotests can write changes to _test.go file or stdout, currently gotest writes to the file itself.

bradleyfalzon:
Note wasn't able to get v:shell_error to ever be true when there was an error gotests does set non-zero exit codes on error. Am I checking this wrong?

bradleyfalzon:
Note, I completely expect feedback on this before a merge, as I'm OK if this isn't yet release ready.

itsjamie:
This is the work for #807. Thank you @bradleyfalzon!

fatih:
Hi @bradleyfalzon 

Thanks for the work you put. I've went over the tool `gotests` first and thinking not to merge it. Let me explain why. 

First of all, the reason is there are tons of other tools, such as stringer that generates code for you. It's just too much for vim-go, because in the end I'll be the guy to maintain this. There are many of other stuff that breaks and needs maintenance, and honestly something like this is just too much for vim-go. 

Second this is something that is not widely used and only a certain percentage of users would use. It's not something like :GoCoverage or co, where it's a part of everyones everyday's workflow. It produces a test for a certain kind (table tests) and that too can be implemented in various ways too.

I know you put some work, but I have to neglect it. I was going to write this under https://github.com/fatih/vim-go/issues/807, but you started it already before I even could write something. 

I hope it's ok for you, I appreciate your work and I hope you don't get me wrong.

Thanks

bradleyfalzon:
Those are valid reasons fatih, and this doesn't preclude creating its own plugin.

PR #821, opened 2016-04-27
pass -tags flag to guru

sillyousu:
`function! go#def#Jump` should pass tags flags to guru.

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -10,8 +10,14 @@ function! go#def#Jump(mode)
 	let old_gopath = $GOPATH
 	let $GOPATH = go#path#Detect()
 
+	let flags = ""
+	if exists('g:go_guru_tags')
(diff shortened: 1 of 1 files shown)


autoload/go/def.vim
```diff
@@ -10,8 +10,14 @@ function! go#def#Jump(mode)
 	let old_gopath = $GOPATH
 	let $GOPATH = go#path#Detect()
 
+	let flags = ""
+	if exists('g:go_guru_tags')
+		let tags = get(g:, 'go_guru_tags')
+		let flags .= printf(" -tags %s", tags)
```
fatih: This should use `=` as we don't append to the string.

fatih:
Thanks @sillyousu for the fix. There is a comment. Once you fix it it's good to go.

sillyousu:
ping @fatih . Fixed.

fatih:
Thanks @sillyousu

PR #823, opened 2016-04-27
don't default to binaries in the PATH

luan:
I asked this on #188 but since that's a closed PR I feel like it's less likely to receive attention so I'm creating a new issue instead.

I have to start this with the question: is there any particular reason for append instead of prepend? This is causing issues if people have some other binary in their path with a colliding name, because appending makes our gobin the last lookup path rather than the first.

I'm getting several reports of people hitting this issue because they had tools compiled on 1.6 and were hanging out in their `$GOPATH/bin` and now that we're using `g:go_bin_path` that is not getting updated, short term fix for us is just removing those stale binaries but seems like a prepend to the `$PATH` would be appropriate.

I made this as a PR because this may be the right solution. I didn't think through all the edge cases here because I wanted to just get opinions on this first. It seems like surprising behavior to me that when I configure `g:go_bin_path` the `$PATH` still overrides it.

Thoughts?

### DIFF
--- a/autoload/go/path.vim
+++ b/autoload/go/path.vim
@@ -138,14 +138,16 @@ endfunction
 function! go#path#CheckBinPath(binpath)
     " remove whitespaces if user applied something like 'goimports   '
     let binpath = substitute(a:binpath, '^\s*\(.\{-}\)\s*$', '\1', '')
+    " save off original path
+    let old_path = $PATH
 
     " check if we have an appropriate bin_path
     let go_bin_path = go#path#BinPath()
-
-    " append our GOBIN and GOPATH paths and be sure they can be found there...
-    " let us search in our GOBIN and GOPATH paths
-    let old_path = $PATH
-    let $PATH = go_bin_path . go#util#PathListSep() . $PATH
+    if empty(go_bin_path)
+        " append our GOBIN and GOPATH paths and be sure they can be found there...
+        " let us search in our GOBIN and GOPATH paths
+        let $PATH = go_bin_path . go#util#PathListSep() . $PATH
+    endif
 
     " if it's in PATH just return it
     if executable(binpath)
@@ -155,13 +157,6 @@ function! go#path#CheckBinPath(binpath)
 
     " just get the basename
     let basename = fnamemodify(binpath, ":t")
-
-    if empty(go_bin_path)
-        let $PATH = old_path
-        echo "vim-go: could not find '" . basename . "'. Run :GoInstallBinaries to fix it."
-        return ""
-    endif
-
     if !executable(basename)
         echo "vim-go: could not find '" . basename . "'. Run :GoInstallBinaries to fix it."
         " restore back!
--- a/autoload/go/path.vim
+++ b/autoload/go/path.vim
@@ -143,7 +143,7 @@ function! go#path#CheckBinPath(binpath)
 
     " check if we have an appropriate bin_path
     let go_bin_path = go#path#BinPath()
-    if empty(go_bin_path)
+    if !empty(go_bin_path)
         " append our GOBIN and GOPATH paths and be sure they can be found there...
         " let us search in our GOBIN and GOPATH paths
         let $PATH = go_bin_path . go#util#PathListSep() . $PATH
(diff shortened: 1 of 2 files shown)


autoload/go/path.vim
```diff
@@ -139,11 +139,6 @@ function! go#path#CheckBinPath(binpath)
     " remove whitespaces if user applied something like 'goimports   '
     let binpath = substitute(a:binpath, '^\s*\(.\{-}\)\s*$', '\1', '')
 
-    " if it's in PATH just return it
-    if executable(binpath) 
-        return binpath
-    endif
-
```
fatih: Any reason you removed this? This shouldn't be removed.

fatih:
Hi @luan 

The only reason is that PATH always comes first. Other than this there is no other significant reason. You have a valid concenr and  I'm ok changing the order of the binary paths.

autoload/go/path.vim
luan: Yeah, maybe this isn't the best way to do this, but, without that it always uses what's present in `$PATH`, regardless of the ordering changes. Maybe we should move the `let $PATH` line up above this block?

luan:
@fatih take a look at the updated changes, I added the temporary `$PATH` first and use `exepath` to resolve it.

autoload/go/path.vim
fatih: I think it should use it from PATH. Can you please change it back? PATH is universal and it's ok to read it from there. It also makes the logic clear.

luan: Did you see my updated changes?

autoload/go/path.vim
```diff
@@ -139,9 +139,15 @@ function! go#path#CheckBinPath(binpath)
     " remove whitespaces if user applied something like 'goimports   '
     let binpath = substitute(a:binpath, '^\s*\(.\{-}\)\s*$', '\1', '')
 
+    " append our GOBIN and GOPATH paths and be sure they can be found there...
+    " let us search in our GOBIN and GOPATH paths
+    let old_path = $PATH
+    let $PATH = go_bin_path . go#util#PathListSep() . $PATH
```
fatih: This is wrong imho. This variable needs to be fetched first below. I think you need to revisit all the logic here?

fatih:
@luan Can you please check the logic. The variable `go_bin_path` needs to be first fetched. You need to change it as well.

autoload/go/path.vim
luan: Yes, it is, sorry. I just realized I had my rtp set to something else when I was doing this. pushing a fish now.

luan:
@fatih yeah sorry. I moved those lines up. What do you mean I need to change it though?

autoload/go/path.vim
```diff
@@ -139,26 +139,29 @@ function! go#path#CheckBinPath(binpath)
     " remove whitespaces if user applied something like 'goimports   '
     let binpath = substitute(a:binpath, '^\s*\(.\{-}\)\s*$', '\1', '')
 
+    " check if we have an appropriate bin_path
+    let go_bin_path = go#path#BinPath()
```
fatih: Can you please handle the case when this is empty (it can be). You should handle it here, not below.

luan:
@fatih handled that case by not changing the `PATH` when `go_bin_path` isn't set, it no longer short circuits on that case and checks for `executable(basename)`. There is a number of ways this could be done and this is the one that looked cleaner to me. Let me know what you think.

autoload/go/path.vim
```diff
@@ -138,14 +138,16 @@ endfunction
...
-
-    " append our GOBIN and GOPATH paths and be sure they can be found there...
-    " let us search in our GOBIN and GOPATH paths
-    let old_path = $PATH
-    let $PATH = go_bin_path . go#util#PathListSep() . $PATH
+    if empty(go_bin_path)
+        " append our GOBIN and GOPATH paths and be sure they can be found there...
+        " let us search in our GOBIN and GOPATH paths
+        let $PATH = go_bin_path . go#util#PathListSep() . $PATH
+    endif
```
fatih: Ok I didn't quite get this. If `go_bin_path` is empty, this means we're going to prepend an empty string to PATH. Why do we prepend an empty string ? We shouldn't do that :)

luan:
@fatih yes... Fixed that.

fatih:
Thanks @luan 👍

Issue #814, opened 2016-04-25
Option to auto-open location list if tests fail


lorin:
If I run `:GoTest`,  and there are test failures, to see them I need to open the location list (`:lopen`). I would love if the location list opened automatically if any tests failed. An option to enable this would be great.

fatih:
Hi @lorin It opens automatically. Are you sure it doesn't ?

lorin:
I did some digging to track this down. The problem seems to only manifest on neovim (not officially supported, I know...), and only when [syntastic](https://github.com/scrooloose/syntastic) is loaded. 

Here's a minimal .nvimrc file that reproduces the issue on my machine:

```
call plug#begin('~/.vim/plugged')
" plugins
Plug 'fatih/vim-go'
Plug 'scrooloose/syntastic'
call plug#end()
```

fatih:
I'm also using `neovim` and I think this is due `syntastic` (which I don't use btw). Closing this as this is not a vim-go issue. Thanks!

lcd047:
> I'm also using `neovim` and I think this is due `syntastic`

Nope, the problem is that both `vim-go` and `syntastic` are trying to manage the same loclist.  As I pointed out [elsewhere](https://github.com/scrooloose/syntastic/issues/1650), this is not going to work, and the solution is for the OP to configure `vim-go` to use quickfix lists instead of loclists.

> (which I don't use btw)

Which is great: you can blame everything on syntastic, and since I'm not using `vim-go`, I can blame it back on `vim-go`.  The perfect circle. :)

fatih:
@lcd047 I'm not trying to blame or start a tension. Please talk for yourself. Here is a gif that shows how vim-go behaves:

![screen recording 2016-05-02 at 09 14 am](https://cloud.githubusercontent.com/assets/438920/14948194/5a5bcd8e-1046-11e6-96a0-7c8bbdeeb52f.gif)

And here is the code: https://github.com/fatih/vim-go/blob/master/autoload/go/cmd.vim#L63-L69 I even have an explicit settings just for jumping to the first error, called `go_jump_to_error`.

As you see it jumps to the first error. So vim-go works out of the box. Also we have extensive detailed settings and you can change the type via `let g:go_list_type = "quickfix"` easily as well.

How is that, vim-go is the one that is broken? Every time there is a problem with vim-go, it's always due syntastic. Heck I even have a dedicated section just for it: https://github.com/fatih/vim-go#using-with-syntastic

[...]

PR #834, opened 2016-05-02
Document using quickfix list for syntastic

lorin:
Add a note in the readme about using quickfix lists to work around an issue when
using syntastic.

For details, see #814 and https://github.com/scrooloose/syntastic/issues/1650


fatih:
Thanks @lorin, much appreciated 👍  /cc @lcd047

PR #838, opened 2016-05-07
add nnoremap <buffer> <silent> <C-w><C-]> :<C-u>call go#def#Jump(spli…

roxma:
vim-go has already map the `<C-]>` key, and shoud also map the `<C-w><C-]>` key, which is opening tag in a split window in vim.

### DIFF
--- a/ftplugin/go.vim
+++ b/ftplugin/go.vim
@@ -33,6 +33,7 @@ if get(g:, "go_def_mapping_enabled", 1)
     " useful again for Go source code
     nnoremap <buffer> <silent> gd :GoDef<cr>
     nnoremap <buffer> <silent> <C-]> :GoDef<cr>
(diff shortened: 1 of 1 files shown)


fatih:
Hi @roxma 

This seems like a good improvement. According to the doc:

```
CTRL-W ]					*CTRL-W_]* *CTRL-W_CTRL-]*
CTRL-W CTRL-]	Split current window in two.  Use identifier under cursor as a
		tag and jump to it in the new upper window.
		In Visual mode uses the Visually selected text as a tag.
		Make new window N high.
```

It has the shorcuts: `<C-W>]` and `<C-W><C-]>`.  It seems like you only added `<C-W><C-]>`. Would you mind to change the code so it also covers  `<C-W>]`  ? Thanks

roxma:
Hi @fatih 
`<C-W>]` is added now

fatih:
Thanks @roxma 👍

PR #846, opened 2016-05-10
impl: add vim-go-impl

### LINKED ISSUE #122: Add code generation of interface implementations
This is an idea. Executing a vim-go command like:

```vim
:GoImplement io.Writer
```

generates a snippet like:

```go
type T struct{}

func (t T) Write(p []byte) (n int, err error) {
	return 0, nil
}
```

The zero value of return variables are optional.

fatih:
This adds the plugin https://github.com/rhysd/vim-go-impl with the
following changes/improvements:

* `:GoImpl` can be now called without any arguments, in this case it
  assumes the identifier under the cursor to be the receiver type
* `:GoImpl io.Writer` is a valid usage now, in this case it assumes the
  identifier under the cursor is the receiver type
* Removed the restriction for completion that there needs to be at least
  three words. This is neede because `:GoImpl io.Writer` is now a valid
  expression.
* We use now `go doc` for the auto completion. This allows for example
  the following usage as well `:GoImpl json.`, in this case you can now
  complete to json.Unmarshaller, whereas `godoc` is not capable of doing
  it.
* There is a newline prependend before we put the generated method stubs
* Refactored to use internal functions of vim-go
* Moved to `autoload` and load it lazily
* Added docs to `docs/vim-go.txt`

Closes #122

Demo:
screen recording 2016-05-10 at 12 32 pm

/cc @rhysd

### DIFF
--- a/autoload/go/impl.vim
+++ b/autoload/go/impl.vim
@@ -1,0 +1,124 @@
+function! go#impl#Impl(...)
+    let binpath = go#path#CheckBinPath('impl')
+    if empty(binpath)
+        return
+    endif
+
+    let recv = ""
+    let iface = ""
+
+    if a:0 == 0
+        " user didn't passed anything,  just called ':GoImpl'
+        let receiveType = expand("<cword>")
+        let recv = printf("%s *%s", tolower(receiveType)[0], receiveType)
+        let iface = input("vim-go: interface to be generated: ")
+        redraw!
+        if empty(iface)
+            call go#util#EchoError('usage: interface type is not provided')
+            return
+        endif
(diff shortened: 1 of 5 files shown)


rhysd:
LGTM :+1:  Thank you for the integration ❤️

fatih:
@rhysd this is now merged. Would you like to add a note to https://github.com/rhysd/vim-go-impl to forward people to vim-go? Thanks to you too for the initial plugin!

rhysd:
@fatih I added a note to an introduction in README.

fatih:
Thanks @rhysd 👍

PR #854, opened 2016-05-13
add goField

hhsnopek:
Adds the ability to differentiate `go.fields` & `go.methods()`.

## Matches
Parses out `bar`
### ==
`foo.bar`
`foo.bar =`
`foo.bar:`

### !=
`foo.bar()`

## note
I'm not sure what the default highlight should link to, so as of now it's `Type`, tho this should be change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,7 @@ To change it:
 ```vim
 let g:go_highlight_functions = 1
 let g:go_highlight_methods = 1
+let g:go_highlight_fields = 1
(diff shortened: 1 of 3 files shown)


fatih:
@hhsnopek thanks for the feature. Can you also update the docs please.

fatih:
Ping @hhsnopek

hhsnopek:
added and rebased @fatih 😄

fatih:
Thanks @hhsnopek 👍

PR #861, opened 2016-05-18
Add non-standard workspaces to package path list

DeepAnchor:
Allows guru tools to work with gb, godep, etc

### DIFF
--- a/autoload/go/package.vim
+++ b/autoload/go/package.vim
@@ -46,7 +46,7 @@ function! go#package#Paths()
         let dirs += [s:goroot]
     endif
 
-    let workspaces = split($GOPATH, go#util#PathListSep())
+    let workspaces = split(go#path#Detect(), go#util#PathListSep())
     if workspaces != []
         let dirs += workspaces
     endif


fatih:
Hi @DeepAnchor 

Thanks for your contribution. Can you please give a detail on how `guru` uses this and what it improves?

DeepAnchor:
Sure. Guru tools such as `implements`, `describe`, `callees`, `callers`, etc., rely on being able to find package import paths to perform its analysis. Currently, `go#package#Paths()` only scans  $GOROOT and the default $GOPATH to find these import paths, causing guru commands to fail when projects are located elsewhere with the message "current directory is not inside of a valid GOPATH".

For example when using [`gb`](https://github.com/constabulary/gb), projects could be located in an arbitrary location outside of $GOPATH and vim-go will fail to recognize them as valid workspaces, causing all guru commands to fail. With this patch, the current workspace will be detected at runtime using `go#path#Detect()` instead of `$GOPATH` which will allow guru to find the proper import paths.

fatih:
Thanks for the clarification @DeepAnchor. I've asked because the function is being used a lot every where else and we still don't have tests (I hope I'll start them soon) to see the global affect I need to be careful while picking stuff.

DeepAnchor:
No problem, better safe than sorry!

Issue #839, opened 2016-05-07
Executing :GoFmt or :GoImports does not work on Windows


vrischmann:
### Actual behavior

Just updated vim-go on my windows machine, and looks like I can't execute any commands (I didn't test everything, just GoFmt and GoImports).

This is the error I get:

    Error detected while processing function go#fmt#Format..go#path#CheckBinPath:
    line   17:
    E117: Unknown function: exepath
    E15: Invalid expression: exepath(binpath)

I confirmed that it works before commit [62f8aa38e3686db79019d75117d35e905a74a6c5](https://github.com/fatih/vim-go/commit/62f8aa38e3686db79019d75117d35e905a74a6c5)    (which introduced exepath I think ?).

### Expected behavior

The command should not fail.

### Steps to reproduce:

1. Create a file with this:

    package main

    func main() {
        fmt.Println("foobar")
    }

2. Run `:GoImports`

### Configuration

Vim 7.4 on Windows 10.

fatih:
Thanks @vrischmann 👍 . Fixed it back.

PR #864, opened 2016-05-20
Properly use exepath for CheckBinPath

luan:
#823 added this the wrong way, it was reverted back on 9f0cf00f6e9a437b738959f629ca29658334c6b5 but still not achieving the right functionality.
 This uses it properly now, also adding it to how we lookup goimports for srcdir support.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -87,7 +87,7 @@ function! go#fmt#Format(withGoimport)
 
     " get the command first so we can test it
     let fmt_command = g:go_fmt_command
-    if a:withGoimport  == 1 
+    if a:withGoimport  == 1
         let fmt_command  = g:go_goimports_bin
     endif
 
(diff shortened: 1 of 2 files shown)


luan:
I realized that #839 had issues with this, it's because they were running an older patch of vim 7.4 that doesn't have `exepath`. I added a guard for that now.

autoload/go/fmt.vim
```diff
@@ -115,7 +115,8 @@ function! go#fmt#Format(withGoimport)
 
     if fmt_command == "goimports"
         if !exists('b:goimports_vendor_compatible')
-            let out = go#util#System("goimports --help")
+            let binpath = go#path#CheckBinPath("goimports")
+            let out = go#util#System(binpath . " --help")
```
fatih: Please remove this as we check above already.

autoload/go/path.vim
```diff
@@ -151,6 +151,9 @@ function! go#path#CheckBinPath(binpath)
 
     " if it's in PATH just return it
     if executable(binpath)
+        if v:version == 704 && has('patch235')
+            let binpath = exepath(binpath)
+        endif
```
fatih: Why do we need to return the full path, I didn't understand your need for this?

luan: This allows us to call the binary from anywhere even on places we are not changing `$PATH`, like on `go#fmt#Format` where we're checking for `goimports` support for `srcdir` or on `go#def#Jump` where we call `guru`. They both fail if you have `goimports` or `guru` in your `$GOBIN` but not in your `$PATH`.

autoload/go/fmt.vim
luan: Pushed a fix for this.

luan:
@fatih thoughts?

emgee:
fwiw, I manually applied the patch and it fixed things for me.

fatih:
Thanks @luan 👍

PR #869, opened 2016-05-22
Increase compatibility with tcsh

tsne:
Hi,

some of the `:Go*` commands report errors when tcsh is configured in Vim. This pull request aims to resolve these errors.
I tested the commands with tcsh and sh configured, and here it works fine. Maybe someone else can double check if everything still works smoothly with her/his shell.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -196,7 +196,7 @@ function! go#cmd#Test(bang, compile, ...)
 
     if a:0
         " expand all wildcards(i.e: '%' to the current file name)
-        let goargs = map(copy(a:000), "expand(v:val)")
+        let goargs = map(copy(a:000), "shellescape(expand(v:val))")
 
         call extend(args, goargs, 1)
     else
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -196,7 +196,8 @@ function! go#cmd#Test(bang, compile, ...)
 
     if a:0
         " expand all wildcards(i.e: '%' to the current file name)
-        let goargs = map(copy(a:000), "shellescape(expand(v:val))")
+        let goargs = map(copy(a:000), "expand(v:val)")
+        let goargs = go#util#Shelllist(goargs, 1)
 
         call extend(args, goargs, 1)
     else
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -13,7 +13,7 @@ function! go#tool#Deps()
     if go#util#IsWin()
         let format = '{{range $f := .Deps}}{{$f}}{{printf \"\n\"}}{{end}}'
     else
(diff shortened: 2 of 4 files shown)


autoload/go/cmd.vim
```diff
@@ -196,7 +196,7 @@ function! go#cmd#Test(bang, compile, ...)
 
     if a:0
         " expand all wildcards(i.e: '%' to the current file name)
-        let goargs = map(copy(a:000), "expand(v:val)")
+        let goargs = map(copy(a:000), "shellescape(expand(v:val))")
```
fatih: Can you just add another line below, so it should be like:

```
let goargs = map(copy(a:000), "expand(v:val)")
let goargs = go#util#Shelllist(goargs, 1)
```

autoload/go/tool.vim
```diff
@@ -1,38 +1,41 @@
 function! go#tool#Files()
     if go#util#IsWin()
-        let command = 'go list -f "{{range $f := .GoFiles}}{{$.Dir}}\{{$f}}{{printf \"\n\"}}{{end}}{{range $f := .CgoFiles}}{{$.Dir}}\{{$f}}{{printf \"\n\"}}{{end}}"'
+        let format = '{{range $f := .GoFiles}}{{$.Dir}}\{{$f}}{{printf \"\n\"}}{{end}}{{range $f := .CgoFiles}}{{$.Dir}}\{{$f}}{{printf \"\n\"}}{{end}}'
     else
-        let command = "go list -f '{{range $f := .GoFiles}}{{$.Dir}}/{{$f}}{{printf \"\\n\"}}{{end}}{{range $f := .CgoFiles}}{{$.Dir}}/{{$f}}{{printf \"\\n\"}}{{end}}'"
+        let format = "{{range $f := .GoFiles}}{{$.Dir}}/{{$f}}{{printf \"\\n\"}}{{end}}{{range $f := .CgoFiles}}{{$.Dir}}/{{$f}}{{printf \"\\n\"}}{{end}}"
     endif
+    let command = 'go list -f '.shellescape(format)
```
fatih: There is already a lot of escaping above, do you think this is necessary?

fatih:
Thanks @tsne for the contribution. My macbook is broken and in service, using a four year old one. Just had time to look at it.

autoload/go/tool.vim
tsne: Unfortunately the escaping is necessary. The escaping above only escapes for the string, but we need to escape for the shell, too.

tsne:
Hi @fatih.

Thanks for reviewing. I fixed the issue with the extra line for `Shelllist` and some escaping issues. Unfortunately we can't get rid of the `shellescape` calls here.

fatih:
Thanks @tsne 👍

PR #878, opened 2016-06-01
Added GoImports mapping

edganiukov:
Hi, 
In this PR I have added mapping for GoImports, to make GoImports assignable to any key combination.


fatih:
Thanks @edganiukov 👍

PR #883, opened 2016-06-03
make go get -u configurable

macb:
Not all GOPATHs are made equal and some don't work well with `go get -u`. This allows users to disable the `-u` for `go get`'s.

cc @si74

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ disabled/enabled easily.
 
 ## Install
 
-Master branch is supposed to be a development branch. So stuff here can break and change. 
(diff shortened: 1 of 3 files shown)


si74:
:+1:

fatih:
Thanks @macb for the contribution. Can you also please add it to our [docs](https://github.com/fatih/vim-go/blob/master/doc/vim-go.txt) as well?

macb:
updated @fatih, that is some fancy doc file.

fatih:
Thanks @macb 👍

PR #888, opened 2016-06-06
def: add go_def_mode setting, we now support godef too

### LINKED ISSUE #887: add support for GoDef call instead of go guru
### Actual behavior
Go Guru is used for "GoDef" or "gd" to jump to definition by default.

### Expected behavior
Have config parameter to use "godef", which is faster in most circumstances.

### Steps to reproduce:

open a go project
put cursor on a function
type ":GoDef"
wait for go guru to do its thing....

Let me know if you need any more info!
Running both in parallel and getting the fastest result would also be awesome!
a variant like `GoDefGuru` for using one or the other would also be awesomer!

fatih:
Closes #887

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -2,202 +2,227 @@ let s:go_stack = []
 let s:go_stack_level = 0
 
 function! go#def#Jump(mode)
-	let bin_path = go#path#CheckBinPath("guru")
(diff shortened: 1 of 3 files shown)


fatih:
Lot's of changes due re-indentation of the file unfortunately, but I had to do it.

PR #894, opened 2016-06-08
open GodocView if argument is provided for :GoDoc

### LINKED ISSUE #893: :GoDoc flag doesn't work
### Actual behavior

As far as I remember, `:GoDoc flag` works well. But says `vim-go: gogetdoc: no documentation found`

### Expected behavior

No use gogetdoc.

### Steps to reproduce:

Please create a reproducible case of your problem. Re produce it 
with a minimal `vimrc` with all plugins disabled and only `vim-go`
enabled:

1. open some go file
2. `:GoDoc flag`

### Configuration

On windows, But same on linux

* vim version: latest
* vim-go version: latest
* go version: latest

mattn:
fixes #893

### DIFF
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -77,17 +77,27 @@ function! go#doc#OpenBrowser(...)
 endfunction
 
 function! go#doc#Open(newmode, mode, ...)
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @mattn 👍

PR #896, opened 2016-06-08
Integration w/Syntastic

MartinBrugnara:
Suggest to explicitly set Syntastic options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,13 @@ let g:go_term_enabled = 1
 ```
 
 ### Using with Syntastic
+Get all the features:
+```vim
+let g:syntastic_go_checkers = ['go', 'golint', 'govet', 'errcheck']
+let g:syntastic_mode_map = { 'mode': 'active', 'active_filetypes': ['go'] }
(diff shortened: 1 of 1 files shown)


fatih:
Hi @MartinBrugnara 

We have already this in our Readme https://github.com/fatih/vim-go#using-with-syntastic. Seems like you missed it :) Thanks!

MartinBrugnara:
Thank you for you fast response, but what we got now is for when _"...Vim will start lagging..."_, and I was not addressing this issue.
I just installed you great plugin, but Syntastic did not 'checked' my files by default (expected behavior).
I found out that I've to explicitly set the lines I propose in this PR.

P.S. 
It could be that I just messed up with my .vimrc but its unlikely since I'd not set any _global_ Syntastic options.

NVIM v0.1.5-393-gd966018
Build type: Release

mrap:
Agreed. The current language suggests a "fix" rather than a sample configuration when you want to use Syntastic. Why not clarify yo?

PR #908, opened 2016-06-22
Adding go-test compiler

nyarly:
The go-test compiler runs `go test -short ./...` and formats the output appropriately.

I'm not 100% convinced this is the best way to handle this (i.e. writing out a shell script to wrap go test to filter the output) but I was trying to avoid adding shell code to the repository, and I couldn't see how to handle `go test`'s output (i.e. "FAIL" after the messages) with errorformat alone.

### DIFF
--- a/compiler/go-test.vim
+++ b/compiler/go-test.vim
@@ -1,0 +1,67 @@
+" Copyright 2013 The Go Authors. All rights reserved.
+" Use of this source code is governed by a BSD-style
+" license that can be found in the LICENSE file.
+"
+" compiler/go.vim: Vim compiler file for Go.
+
+if exists("current_compiler")
+  finish
+endif
+
+runtime! compiler/go.vim
+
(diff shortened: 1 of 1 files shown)


fatih:
Hi @nyarly 

What is this solving? We have already `:GoBuild`, `:GoTest` and `:GoTestCompile`.

nyarly:
I was setting this up as a compiler to work more closely with established Vim features. 

At least in my experience, the output of this compiler is nicer than the GoTest output - e.g. the QF list links to the test files when there are fails

fatih:
@nyarly currently `:GoTest` also opens a QF window and you can jump between the files. I'm not sure what this fixes? I definitely don't want to include any shell script, as this plugin is used on several other platforms as well (linux, darwin, windows, etc..). If you have any issue with the original behaviour please try to fix the original `:GoTest`, which is well tested and used by many users already for a long time. Thanks.

PR #914, opened 2016-06-24
guru: fix slash direction for windows users

### LINKED ISSUE #872: Windows GVim :GoCallers Problem
Hi @fatih,

I came up with another path backslash problem and a toy/raw solution for :GoCallers command.

```
diff --git a/autoload/go/package.vim b/autoload/go/package.vim
index 4f75613..f577da4 100644
--- a/autoload/go/package.vim
+++ b/autoload/go/package.vim
@@ -59,7 +59,7 @@ function! go#package#ImportPath(arg)
     let dirs = go#package#Paths()
 
     for dir in dirs
-        if len(dir) && match(path, dir) == 0
+        if len(dir) && matchstr(escape(path, '\/'), escape(dir, '\/')) == 0
             let workspace = dir
         endif
     endfor
@@ -68,8 +68,8 @@ function! go#package#ImportPath(arg)
         return -1
     endif
 
-    let srcdir = substitute(workspace . '/src/', '//', '/', '')
-    return substitute(path, srcdir, '', '')
+    let srcdir = substitute(workspace . '\src\', '//', '/', '')
+    return path[len(srcdir):]
 endfunction
 
 function! go#package#FromPath(arg)
```

I wish it will help you to solve the issue in a more elegant way.

Best,

fatih:
Closes #872

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -7,7 +7,7 @@ func! s:RunGuru(mode, format, selected, needs_scope) range abort
     return {'err': "bin path not found"}
   endif
 
-  let filename = expand('%:p')
+  let filename = fnamemodify(expand('%:p'), ':p:gs?\\?/?')
   let dirname = expand('%:p:h')
   let pkg = go#package#ImportPath(dirname)
 


bhcleek:
This shouldn't be necessary anymore, because I made the same change in #933, was incorporated into #936.

fatih:
Right @bhcleek. Thanks for the heads up.

PR #915, opened 2016-06-25
style: use two spaces for all files

fatih:
I had enough of dealing with various styles in our VimL files. We had a
mix of tabs and spaces over all files. Some were using tabs, some were
using 4 spaces.

From now on the official vim-go style will be the following settings:

```
shiftwidth=2
tabstop=2
expandtab
```

The modeline for this setting is:

```
" vim: sw=2 sw=2 et
```

I've added the modeline for all files and reformatted all the files
again. Some might not like the two space setting, but that's ok. The
official Vim runtime files are usually like this, and the vim style
guide from Google also uses two spaces for indents (no tabs).

If someone wants to set a autocmd, this can be used as well:

```
autocmd BufNewFile,BufRead *.vim setlocal expandtab shiftwidth=2 tabstop=2
```

### DIFF
--- a/autoload/ctrlp/decls.vim
+++ b/autoload/ctrlp/decls.vim
@@ -1,26 +1,26 @@
 let s:go_decls_var = {
-			\  'init':   'ctrlp#decls#init()',
-			\  'exit':   'ctrlp#decls#exit()',
-			\  'enter':  'ctrlp#decls#enter()',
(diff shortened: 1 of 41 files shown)


fatih:
To see the diff with whitespace ignored checkout: https://github.com/fatih/vim-go/pull/915/files?w=1

PR #920, opened 2016-06-30
Fixed typo causing GoDocBrowser to fail

### LINKED ISSUE #919: Typo: calling echoWarning instead of EchoWarning causes :GoDocBrowser to fail if godoc is missing
https://github.com/fatih/vim-go/blob/1a9cc786751947441f05b92f380540dee388feff/autoload/go/doc.vim#L22

GoDocBrowser call fails because of this.

dobegor:
Fixes #919.


fatih:
Thanks @dobegor 👍

PR #922, opened 2016-06-30
Restore original files permissions

### LINKED ISSUE #921: Original file permissions are lost on gofmt
### Actual behavior

When editing a go source file, the file permissions can be changed.

### Expected behavior

The file permissions should not be changed

### Steps to reproduce:

1. set umask to 0022
2. touch test.go
3. chmod 666 test.go
4. vi test.go
5. save and quit
6. file permissions are now 644

joegrasse:
This fixes #921.

### DIFF
--- a/autoload/go/asmfmt.vim
+++ b/autoload/go/asmfmt.vim
@@ -41,7 +41,11 @@ function! go#asmfmt#Format()
 
     " Replace the current file with the temp file; then reload the buffer.
     let old_fileformat = &fileformat
+    " save old file permissions
+    let original_fperm = getfperm(expand('%'))
     call rename(l:tmpname, expand('%'))
+    " restore old file permissions
+    call setfperm(expand('%'), original_fperm)
     silent edit!
     let &fileformat = old_fileformat
     let &syntax = &syntax
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -149,7 +149,11 @@ function! go#fmt#Format(withGoimport)
 
     " Replace current file with temp file, then reload buffer
     let old_fileformat = &fileformat
+    " save old file permissions
+    let original_fperm = getfperm(expand('%'))
     call rename(l:tmpname, expand('%'))
+    " restore old file permissions
+    call setfperm(expand('%'), original_fperm)
     silent edit!
     let &fileformat = old_fileformat
     let &syntax = &syntax


autoload/go/asmfmt.vim
```diff
@@ -41,7 +41,11 @@ function! go#asmfmt#Format()
 
     " Replace the current file with the temp file; then reload the buffer.
     let old_fileformat = &fileformat
+    " save old file permissions
+    let original_fperm = getfperm(expand('%'))
```
fatih: According to the doc this also returns `-` if the user does not have the given permission. We should test if the variable is `-` below before we set it to prevent unwanted errors.

fatih:
Thanks @joegrasse. If you add the check in `setfperm()` before overriding the permission we're good to go.

joegrasse:
Actually I believe we don't want the check for ```-```. In my scenario in #921, the original file would have returned ```-rw-rw-rw-```. After editing the file, with my umask set to ```0022```, it would become ```-rw-r--r--```. With my change above it stays ```-rw-rw-rw-``` after editing. So a ```-``` is actually a valid scenario.

fatih:
The problem is that it should not set the permissions when `getfperm` returns a `-`. It's an edge case we don't want. So you should add an if clause just before `setfperm` to avoid this.

joegrasse:
I sorry, I guess I am not following you. Are you thinking there is the case where it would return only ```-```? Even if the file is chmoded as ```000```, it would still return ```---------```. What the doc is saying is, if the file doesn't has a specific permission it returns ```-```. So if the file doesn't have the group writable permission it would give a ```-``` in that permission slot instead of ```w```.

gabrielsjoberg:
Passing "-" back into `setfperm()` turns the relevant permission off, which is the expected behavior if the original file also did not have that permission.  Calling something like the following would essentially be a NOOP:

`setfperm(file, getfperm(file))`

However, in the reported bug, there are two separate files: the original one with the intended permissions and a temp file that has different (wrong) permissions due to the user's umask.  When the temp file is moved over the original file, the temp file's permissions overwrite the original, intended permissions.  This could lead to problems such as a world-writable file suddenly becoming read-only to everyone but the current user (or worse, it could happen the other way around).

This PR reapplies the permissions from the original file to the temp file exactly as they were.  When the temp file is moved into place, it will have the same permissions as it did before calling `fmt()`, with no new permissions added and no previous permissions lost.  Other than verifying that `getfperm()` does not return an empty string, no further checks or modifications of the perm string are necessary.

joegrasse:
Do the last couple of comments clear your questions up for you, or do you still have concerns?

fatih:
Thanks @joegrasse

PR #924, opened 2016-06-30
fix highlight type keyword

tsne:
If `g:go_highlight_types` is set to 0, the `type` keyword is not highlighted. This pull request fixes this issue.


fatih:
Thanks @tsne 👍

PR #925, opened 2016-07-04
Changed/Fixed go-to-def using tabs

cfsalguero:
If you are using tabs, go-to-def not always switches to the correct tab if the file is already open in a different tab.
I've simplified the switching by using Vim's 'drop' command instead of manually checking for the correct tab.
This was tested on Neovim as well.


fatih:
Thanks @cfsalguero 

I'm busy lately but I'll look into it and will reform you. Just a side not that you have used two different accounts for pushing the commits. Please open the PR in a single commit and be sure you only use a single account. Thanks

PR #927, opened 2016-07-05
snippet: make json field tag expansion smarter

fatih:
Now the `json` snippet expansion is much more smarter. It does the followings:

* The placeholder is pre populated according to the first word in the line
* It automatically converts it to either `snakecase` or `camelcase`

Demo:

screen recording 2016-07-05 at 04 15 am


andrewstuart:
This is awesome. :-)

PR #928, opened 2016-07-05
Changed/Fixed go-to-def using tabs

cfsalguero:
If you are using tabs, go-to-def not always switches to the correct tab
if the file is already open in a different tab.
I've simplified the switching by using Vim's 'drop' command instead of
manually checking for the correct tab.
This was tested on Neovim as well.

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -100,18 +100,20 @@ function! s:jump_to_declaration(out, mode)
 
   " jump to existing buffer if, 1. we have enabled it, 2. the buffer is loaded
   " and 3. there is buffer window number we switch to
-  if a:mode == "tab"
-    exec 'tab drop '.filename
-  elseif get(g:, 'go_def_reuse_buffer', 0) && bufloaded(filename) != 0 && bufwinnr(filename) != -1
-    exec 'drop '.filename
+  let cmd='edit '
+  if get(g:, 'go_def_reuse_buffer', 0) && bufloaded(filename) != 0 && bufwinnr(filename) != -1
+    " jumpt to existing buffer if it exists
+    execute bufwinnr(filename) . 'wincmd w'
+  elseif a:mode == "tab"
+    let cmd='tab drop '
   elseif a:mode == "split"
-    split 
-    exec 'edit '.filename
+    split
   elseif a:mode == "vsplit"
     vsplit
-    exec 'edit '.filename
   endif
 
+  " open the file and jump to line and column
+  exec cmd.filename
   call cursor(line, col)
 
   " also align the line to middle of the view
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -100,22 +100,18 @@ function! s:jump_to_declaration(out, mode)
 
   " jump to existing buffer if, 1. we have enabled it, 2. the buffer is loaded
   " and 3. there is buffer window number we switch to
-  if get(g:, 'go_def_reuse_buffer', 0) && bufloaded(filename) != 0 && bufwinnr(filename) != -1
-    " jumpt to existing buffer if it exists
-    execute bufwinnr(filename) . 'wincmd w'
-  elseif a:mode == "tab"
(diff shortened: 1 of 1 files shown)


bhcleek:
I've found two issues with this PR.

First, if the current buffer is unmodified, and you try to jump using `gd` to an identifier that's in another file, the buffer doesn't change. The cursor, though, is placed on the line of the current buffer on which the identifier is defined in the buffer that wasn't loaded.

Second, if the current buffer is modified, and you try to jump using `gd` to an identifier that's in another file, an error occurs, and the cursor doesn't jump as expected.

percona-csalguero:
@bhcleek I've made a new fix.
I wasn't using buffers before (I always use tabs). Could you check this new commit please?
Thanks.

bhcleek:
That fixes the first issue. The second issue, though, still isn't resolved. However, I don't think it's necessarily the job of this PR to resolve it, since the problem already exists on master. If this gets merged, I'll submit a PR to resolve the second issue.

fatih:
Hi @cfsalguero 

I've tried to test this for the following file (put my cursor on top of `Println`:

```go
package main

import "fmt"

func main() {
	fmt.Println("vim-go")
}
```

but I'm receiving this error:

```
E319: Sorry, the command is not available in this version: tab drop /usr/local/go/src/fmt/print.go
```

I'm using HEAD of Vim, currently it's 7.4 Patch 2085 with Huge version. These are the options that are enabled as well:

```
+acl             +cmdline_hist    -ebcdic          -gettext         +localmap        -mouse_sysmouse  +postscript      +startuptime     +timers          +windows
+arabic          +cmdline_info    +emacs_tags      -hangul_input    +lua             +mouse_urxvt     +printer         +statusline      +title           +writebackup
+autocmd         +comments        +eval            +iconv           +menu            +mouse_xterm     +profile         -sun_workshop    -toolbar         -X11
-balloon_eval    +conceal         +ex_extra        +insert_expand   +mksession       +multi_byte      +python          +syntax          +user_commands   -xfontset
-browse          +cryptv          +extra_search    +job             +modify_fname    +multi_lang      -python3         +tag_binary      +vertsplit       -xim
++builtin_terms  +cscope          +farsi           +jumplist        +mouse           -mzscheme        +quickfix        +tag_old_static  +virtualedit     -xsmp
+byte_offset     +cursorbind      +file_in_path    +keymap          -mouseshape      +netbeans_intg   +reltime         -tag_any_white   +visual          -xterm_clipboard
+channel         +cursorshape     +find_in_path    +langmap         +mouse_dec       +num64           +rightleft       -tcl             +visualextra     -xterm_save
+cindent         +dialog_con      +float           +libcall         -mouse_gpm       +packages        +ruby            +termguicolors   +viminfo         -xpm
-clientserver    +diff            +folding         +linebreak       -mouse_jsbterm   +path_extra      +scrollbind      +terminfo        +vreplace
+clipboard       +digraphs        -footer          +lispindent      +mouse_netterm   +perl            +signs           +termresponse    +wildignore
+cmdline_compl   -dnd             +fork()          +listcmds        +mouse_sgr       +persistent_undo +smartindent     +textobjects     +wildmenu
```

Seems like this only exists for certain versions. If yes what is it?

fatih:
Oh seems like this is only valid if it's compiled with `gui`:

```
{only available when compiled with a GUI}
```

Carpetsmoker:
This feature uses `+clientserver`, which is implemented using the X API...

At any rate, there is a reason I stopped using everything that depends on this feature: there is no timeout when waiting for a process to respond.
For example, let's say I open `a.go` in one Vim instance, hit `<C-z>` to do something else, and then open `a.go in another Vim process. The second Vim process will now hang indefinitely because it's waiting for the first Vim process to respond...

I don't know if many people use `<C-z>` like this − I often use it anyway...

fatih:
I'm closing this as it's not compatible without guarding it and can't be merged as well. Please reopen a clean PR with the latest changes. Thanks.

PR #930, opened 2016-07-05
small syntax fixes

datanoise:
Fixed issues reported by @svanharmelen in comments of pull request #917.
 
- do not highlight method receiver if g:go_highlight_types is not set
- fixed highlighting of anonymous functions

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -286,19 +286,22 @@ if g:go_highlight_operators != 0
   syn match goOperator /:=\|||\|<-\|++\|--/
   " match ...
   syn match goOperator /\.\.\./
+
+  hi def link     goPointerOperator   Operator
 endif
 hi def link     goOperator          Operator
 
 " Functions;
 if g:go_highlight_functions != 0
   syn match goDeclaration       /\<func\>/ nextgroup=goReceiver,goFunction skipwhite skipnl
-  syn match goReceiver /([^),]\+)/ contained nextgroup=goFunction contains=goReceiverType skipwhite skipnl
-  syn match goReceiverType /\(\s\|*\)\w\+)/hs=s+1,he=e-1 contained
-  syn match goFunction /\w\+/ contained
+  syn match goReceiver          /(\(\w\|[ *]\)\+)/ contained nextgroup=goFunction contains=goReceiverVar skipwhite skipnl
+  syn match goReceiverVar       /\w\+/ nextgroup=goPointerOperator,goReceiverType skipwhite skipnl contained
+  syn match goPointerOperator   /\*/ nextgroup=goReceiverType contained skipwhite skipnl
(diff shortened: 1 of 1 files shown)


svanharmelen:
FYI:

image

fatih:
@datanoise would you mind to check the comment from @svanharmelen?

I’m on travel this weekend, and next week I’ll be in Gophercon. I’m preparing myself for the vim-go session. So the issues are on hold till next weekend. Thanks for all the feedback and fixes!

datanoise:
This two examples are quite different. My changes added highlighting of the method's receiver types. For example, in the following `Status` method

```go
func (reqCtx *requestContext) Status() int {
```

`requestContext` is the type of the receiver.

In this example, we have a plain go function `NewRequest` (not a method) 

```go
func NewRequest(ctx *Options) RequestContext {
```
and `ctx *Options` are parameters to this function. I don't think go syntax ever provided highlighting for function parameters. Something that we should consider adding in the future I guess.

I'm traveling until the next week. If someone can look at this problem and send a PR, it would be nice.

svanharmelen:
@datanoise the go syntax highlighting from before #917 did highlight this properly, so IMHO it's "bug" introduced in #917 

I reverted `vim-go` to a commit before #917 for the time being, guess my eyes are overly sensitive to changes :wink:

datanoise:
@svanharmelen maybe I misunderstand the bug report. Which part is highlighted properly before #917?

Are we talking about Option type in 

```go
func NewRequest(ctx *Options) RequestContext {
```

svanharmelen:
Yes, have a look at how it highlights the pointer sign before #917:

image

datanoise:
I'm maybe missing something, but in this particular case the pointer sign is highlighted properly even without this pull request or #917 (assuming you have `let g:go_highlight_operators = 1`). The syntax name of the pointer sign is detected as `goOperator`. My patches should not touch the syntax of function parameters.

svanharmelen:
Hmm... I guess I am the one missing something 😞  I forwarded this from the slack channel, without verifying it myself... I'll point your conclusions out to the guy posting this. If he want to follow up or explain it in more detail himself he should. Otherwise I guess I have to apologize myself for causing this noise....

magicalbanana:
Wait, so it shouldn't be highlighted?

svanharmelen:
It should... And it does (at least with this patch) in both cases (receiver and func param).

magicalbanana:
Still nothing, perhaps it's because of the colorscheme I use [gruvbox](https://github.com/morhetz/gruvbox)

datanoise:
gruvbix doesn't highlight Operator syntax. Try something like :hi! link Operator String

nhooyr:
@fatih please merge this!

fatih:
Thanks @datanoise 👍

Issue #199, opened 2014-10-12
GoRename doesn't work for file not opened from current directory


tbeati:
If i do :
$ vi main/main.go
and try to use GoRename i get an error : vim-go: Error: no such file: main.go.

If i do :
$vi main.go
directly from the directory it works fine.

PR #933, opened 2016-07-12
implement :GoSameIds

bhcleek:
Add :GoSameIds to highlight all identifiers in the current buffer that
are equivalent to the identifier under the cursor.

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -7,7 +7,13 @@ func! s:RunGuru(mode, format, selected, needs_scope) range abort
     return {'err': "bin path not found"}
   endif
 
-  let filename = expand('%:p')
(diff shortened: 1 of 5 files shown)


fatih:
Seems like there is a bug if you open a new go file and if `auto_sameids` is enabled. It just fails. I need to look more on why it's like that.

fatih:
Alright found the issue. But there are other issues as well. This is a great feature, just it needs some tweaking. For example `CursorMove` is not suitable. It's slows down Vim (and the document also says it's an expensive operation). I think better is `CursorHold` with a `set updatetime=100`. We can set it ourselves. And probably even better is to use a caching mechanism, but I'm not sure how we could cache according to the byte offset, because say `#197` and `#198` might refer to the same identifier but not `#199`

PR #936, opened 2016-07-13
implement :GoSameIds

fatih:
Carried https://github.com/fatih/vim-go/pull/933 over to here with additional improvements and fixes.

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -558,6 +558,11 @@ CTRL-t
     Highlights all identifiers that are equivalent to the identifier under the
     cursor.
 
+                                                          *:GoSameIdsClear*
+:GoSameIdsClear
+
+    Clears all SameIds highlights from a |:GoSameIds| call.
+
                                                               *:GoMetaLinter*
 :GoMetaLinter [path]
 
@@ -927,16 +932,16 @@ with |:GoPlay|. By default it's enabled. >
 Use this option to show the type info (|:GoInfo|) for the word under the cursor
 automatically. Whenever the cursor changes the type info will be updated.
 By default it's disabled.
-
+>
   let g:go_auto_type_info = 0
-
+<
                                                       *'g:go_auto_sameids'*
 
-use this option to highlight all uses of the identifier under the cursor
-(:GoSameIds) automatically. By defult it's disabled.
-
+Use this option to highlight all uses of the identifier under the cursor
+(:GoSameIds) automatically. By default it's disabled.
+>
   let g:go_auto_sameids = 0
-
+<
                                                       *'g:go_jump_to_error'*
 
 Use this option to enable/disable passing the bang attribute to the mappings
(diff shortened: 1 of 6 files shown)


fatih:
@bhcleek I've carried your changes over here with the fixes and improvements I made. Thanks for all this again, I'm still testing it and think to merge it after the GopherCon workshop as I believe this should be thoroughly tested for edge cases.

doc/vim-go.txt
```diff
@@ -927,16 +927,21 @@ with |:GoPlay|. By default it's enabled. >
...
-use this option to highlight all uses of the identifier under the cursor
-(:GoSameIds) automatically. By defult it's disabled.
-
+Use this option to highlight all uses of the identifier under the cursor
+(:GoSameIds) automatically. Please set |updatetime| by adding it to your vimrc
+to a value that is suitable for you. Example:
+>
+  set updatetime=100
+<
+By defult it's disabled.
```
bhcleek: defult -> default

bhcleek:
I don't think you need to use CursorHold. The docs for CursorMoved say "don't do anything that is slow", which is different than "it's an expensive operation". FWIW, I talked to Alan Donovan, and he confirmed that `guru what` is going to be very fast (a few milliseconds). Coupled with his input and actual use, I chose to use CursorMoved over CursorHold intentionally. In my testing it was definitely very performant; I didn't perceive any lag when navigating through some large files. And there are valid reasons that someone may not want vim-go to mess with their updatetime value.

fatih:
@bhcleek try to create a variable called `foooooooooooo` and then start from `f` till the end by moving your cursor. You'll see that there is lag and slowness with `CursorMoved`, whereas there is none with `CursorHold` as it doesn't count movements as you don't need to calculate the sameID query while you're moving.

I'm still using and testing it. Usually it takes time until I merge stuff like this. It's good because you'll find edge cases and fix them.

bhcleek:
Yeah, I do see some lag there. Here's another alternative solution that you may want to consider: http://vim.wikia.com/wiki/Prevent_frequent_commands_from_slowing_things_down.

I'm totally onboard with taking the time to make sure it's tested and the edge cases are properly worked out. Please let me know if you need me to do anything else; I'm happy to help polish this up.

bhcleek:
FWIW, the reason I'd like to find a solution that uses CursorMoved instead of CursorHold, is because I really like the highlighting as I'm moving through the file instead of having to wait for a pause after movement.

bhcleek:
I've been playing with this some more, and I don't see any lag when using CursorMoved, even when editing a large file that has unsaved changes. Could the lag you're seeing be resolved by moving `call go#guru#ClearSameIds` to after `call go#guru#What` instead of before?

fatih:
@bhcleek I've moved back to `CursorMoved`, but also added explicit highlighting. Thanks again for your work.

bhcleek:
Thanks @fatih, and thank you for helping to polish this up.

I enjoyed meeting you at GopherCon. Sorry I couldn't make your session on the hack day.

fatih:
You're welcome @bhcleek, me as well! 

I've mentioned you in the session as well and gave you a shoot out to your name :) I hope people remember you. And thanks again for your work, much appreciated.

bhcleek:
+1

PR #937, opened 2016-07-13
add yaml tags completion

dangogh:
Unsure if there is more to it than this,  but make yaml tag completion work as json does.


fatih:
That's it thanks. Eventually we can get rid of all this maybe and put the `json`, `yaml` automatically just according for the given name. But that can wait :) Thanks @dangogh 👍

PR #938, opened 2016-07-15
reduce flicker when using go_auto_sameids

bhcleek:
Only clear the highlighted identifiers after running Guru to reduce the
flicker time.

It's a very minor improvement, and probably barely noticeable on some machines, but on a large modified buffer with a slow disk drive it may be more significant.


fatih:
I've noticed this improvement, looks good thanks @bhcleek 👍

PR #939, opened 2016-07-15
allow auto features to be disabled without restarting Vim

bhcleek:
I find sometimes want to change the settings of vim-go that control whether or not some actions happen in response to events (e.g. formatting, linting, etc.). This change allows the settings to be changed at runtime instead of having to restart Vim to pick up the changes.

There were a couple of `autocmd` entries where I removed the `nested` argument, because it didn't seem to be necessary.


fatih:
Thanks @bhcleek 👍

PR #940, opened 2016-07-15
allow goSameId highlight group to be overridden

bhcleek:
Move the modification of the goSameId highlight group from
autoload/go/guru.vim to a proper definition in syntax/go.vim so that
users can customize it.


fatih:
Thanks @bhcleek 👍

PR #942, opened 2016-07-15
Neovim supports JSON, :GoSameIds now works.

### LINKED ISSUE #941: :GoSameIds doesn't work in Neovim: GoWhat is not supported in Neovim (though JSON support is already there)
We can't use `:GoSameIds` in Neovim due to this error.

dobegor:
Fixes #941.
![screenshot] (http://i.imgur.com/4gOMJsL.png)

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -271,10 +271,11 @@ function! go#guru#Referrers(selected)
 endfunction
 
 function! go#guru#What(selected)
-  " json_encode() and friends are introduced with this patch
-  " https://groups.google.com/d/msg/vim_dev/vLupTNhQhZ8/cDGIk0JEDgAJ
-  if !has('patch-7.4.1304')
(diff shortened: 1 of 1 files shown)


fatih:
Can you please add a check  that checks whether the `json_decode` exists? Just remove the Vim and NeoVim specific checks all together. I think it would be more useful:


```viml
  " json_encode() and friends are introduced with this patch (7.4.1304)
  " vim: https://groups.google.com/d/msg/vim_dev/vLupTNhQhZ8/cDGIk0JEDgAJ
  " nvim: https://github.com/neovim/neovim/pull/4131		
  if !exists("*json_decode")
    return {'err': "GoWhat is not supported due old version of Vim/NeoVim"}
  endif
```

dobegor:
@fatih thanks for the note. Added a commit to this PR.

fatih:
Thanks @dobegor 👍

PR #943, opened 2016-07-15
Add ToggleSameIds

idawes:
Auto sameIds is cool, but it doesn't behave nicely in large files, where scolling the screen causes the cursor to move and therefore change the identifier that's being matched. This change allows me to quickly turn on and off same-id highlighting, and have the highlighting stay in effect as I scroll through the file.

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -346,4 +346,12 @@ function! go#guru#ClearSameIds()
   endfor
 endfunction
 
+function! go#guru#ToggleSameIds(selected)
+  if len(getmatches()) != 0 
+    call go#guru#ClearSameIds()
+  else
+    call go#guru#SameIds(a:selected)
+  endif
+endfunction
+
 " vim: sw=2 ts=2 et
--- a/ftplugin/go/commands.vim
+++ b/ftplugin/go/commands.vim
@@ -14,6 +14,7 @@ command! -range=% GoReferrers call go#guru#Referrers(<count>)
 command! -nargs=? GoGuruTags call go#guru#Tags(<f-args>)
 command! -range=% GoSameIds call go#guru#SameIds(<count>)
 command! -range=0 GoSameIdsClear call go#guru#ClearSameIds()
+command! -range=% GoToggleSameIds call go#guru#ToggleSameIds(<count>)
 
 " -- tool
 command! -nargs=0 GoFiles echo go#tool#Files()
--- a/ftplugin/go/mappings.vim
+++ b/ftplugin/go/mappings.vim
@@ -44,6 +44,7 @@ xnoremap <silent> <Plug>(go-freevars) :<C-u>call go#guru#Freevars(0)<CR>
 nnoremap <silent> <Plug>(go-channelpeers) :<C-u>call go#guru#ChannelPeers(-1)<CR>
 nnoremap <silent> <Plug>(go-referrers) :<C-u>call go#guru#Referrers(-1)<CR>
 nnoremap <silent> <Plug>(go-sameids) :<C-u>call go#guru#SameIds(-1)<CR>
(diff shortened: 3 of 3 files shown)


bhcleek:
@idawes I'm interested in addressing the "doesn't behave nicely with large files" that you describe. How big of a file are you working with when you see problems?

idawes:
Perhaps this is something that I've got misconfigured. I'm using neovim in the terminal, and when I scroll the file using the mouse, the cursor stays on its current line until it hits the top (or bottom) of the screen, then moves down (or up) x lines (the mouse scroll distance) to stay in the visible portion of the file. When this happens, auto_sameids causes the highlighted selection to change. This isn't particularly surprising behaviour, given that it is just highlighting instances of whatever identifier is under the current cursor position. It is however undesirable, because I want to be able to see instances of the original highlighted identifier throughout the entire file, not just within one screen. 

After submitting this PR, I've actually found that it works better for me to just bind <leader>gs to :GoSameIds, and add :GoSameIdsClear to my C-l (clear highlight) binding. That way I can highlight all instances of an identifier throughout a file, then move to another identifier and highlight those without first having to toggle off the highlighting.

bhcleek:
That makes sense, and I think your expectation is natural. #939 may point a way forward for you. With #939, if there were a function that would toggle `g:go_auto_sameids`, you could add a mapping that would call that function before you start scrolling. Would that meet your needs?

idawes:
That sounds like another way to address the issue for sure. I think I'll likely stick with turning off auto sameids and manually triggering sameids on and off though, as that currently fits with how I think about the functionality. 

I don't mind if you close this pr.

fatih:
I'm also experiencing the same as @idawes. And it's ok because the fix is not possible with current 7.4. I'm using it also for the past week so I have some ideas. My idea to fix it and going forward with vim 8.0 is the following (@bhcleek also shares the same idea as I see)

* Change `:GoSameIds` behaviour. When called automatic highlighting of identifiers is enabled. 
* Calling `:GoSameIds` again will stop automatic highlighting.
* We still have `g:go_auto_sameids`, without the commands

With vim 8.0 we're going implement a Job based call. Right now the implementation in `emacs` has a `200ms` latency, so `:GoWhat` is only called after 200ms when the cursor is stopped. But in our case it's not like that. And honestly I think it should have some caching or blocking mechanism.

What do you think? Does the above makes sense ?

idawes:
That's essentially what I was doing with this PR... I the toggle command bound to <leader>gs, so typing that once turned on the highlighting, and typing it again turned it off. I found that it was annoying if I had one thing highlighted and wanted to highlight something else. I had to type it 2 times, once to turn it off and once to start highlighting the new thing. 

Now I just have <leader>gs bound to GoSameIds, and I call GoSameIdsClear with Ctrl-l. This means that I can turn highlighting on, then move to a new identifier and highlight that, and when I want to clear the highlighting it's the same as clearing the highlighting for a search. 

I find this is working really well for me, so I'd prefer if GoSameIds didn't change behaviour. If you want the behaviour you're talking about @fatih, I'd suggest merging this PR and binding to that instead.

fatih:
This PR doesn't automatically enables it. It just highlights for the current cursor. When you move your cursor it doesn't do anything. I the approach I've described is useful and makes sense going forward.

idawes:
Oh right... sorry... missed that you were looking to change it to toggle auto-highlighting. If you do change the behaviour of GoSameIds, would you consider adding another function to trigger highlighting of the current identifier without enabling auto-highlighting?

bhcleek:
@fatih are you suggesting that `GoSameIds` be changed to turn on automatic highlighting, or just that `GoSameIds` should toggle highlighting of the identifier under the cursor? I tend to agree with @idawes - `GoSameIds` shouldn't `g:go_auto_sameids`. Now that #939 is merged, though, I think it makes sense to add a specific `GoToggleAuto....` for each of the auto features.

PR #944, opened 2016-07-16
use guru's -modified flag

bhcleek:
Pass the -modified flag and the buffer contents to guru on stdin when
the buffer has been modified instead of having to manage a temp file
manually.

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -116,7 +116,7 @@ function! s:jump_to_declaration(out, mode)
   " jump to existing buffer if, 1. we have enabled it, 2. the buffer is loaded
   " and 3. there is buffer window number we switch to
   if get(g:, 'go_def_reuse_buffer', 0) && bufloaded(filename) != 0 && bufwinnr(filename) != -1
-    " jump to existing buffer if it exists
+    " jumpt to existing buffer if it exists
     execute bufwinnr(filename) . 'wincmd w'
   elseif a:mode == "tab"
     let &switchbuf = "usetab"
@@ -127,8 +127,6 @@ function! s:jump_to_declaration(out, mode)
     split
   elseif a:mode == "vsplit"
     vsplit
-  elseif &modified
-    split
   endif
 
   " open the file and jump to line and column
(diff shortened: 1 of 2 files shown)


fatih:
Just tested locally, but get this error when I jump when my file is modified:

image 2016-07-16 at 8 42 26 pm

bhcleek:
I see the same behavior on the latest release, too, though; it wasn't introduced in this PR. The problem is due to https://github.com/fatih/vim-go/blob/595df5118585d931fad648aeef5e766094941f49/autoload/go/def.vim#L118. I think that `drop` should be used instead of edit, but I'm not sure. I started to fix it, but then noticed #928, which fixes the problem. However, #928 has a different problem that I'll describe there.

bhcleek:
@fatih In light of your discovery that `:drop` is only available when compiled with a GUI (#928), I went ahead and resolved jumping to a definition when the buffer is modified by using `:split` when no other option has been specified.

fatih:
If you try to jump to the same file we get this error:
image

bhcleek:
Yes, that error does happen, but it also happens on master. At this point, it's clear that the problem is not this PR, but rather master. I've reverted my last commit, and may submit a change to master that resolves the problems with jumping to definitions from modified buffer as a separate PR.

bhcleek:
I duplicated what you were seeing and thought it was a separable task, so I reverted `	bc000e0`. All the problems you see with modified buffers also exist on master, but there's other problems on master with modified buffers that the guru changes here address, so I ended up fixing the problems associated with jumping to a definitions from a modified buffer, regardless of whether the definition is in the same buffer or a different buffer.

autoload/go/def.vim
```diff
@@ -98,24 +113,29 @@ function! s:jump_to_declaration(out, mode)
...
-  " and 3. there is buffer window number we switch to
-  if get(g:, 'go_def_reuse_buffer', 0) && bufloaded(filename) != 0 && bufwinnr(filename) != -1
-    " jumpt to existing buffer if it exists
-    execute bufwinnr(filename) . 'wincmd w'
-  elseif a:mode == "tab"
-    let &switchbuf = "usetab"
-    if bufloaded(filename) == 0
-      tab split
+  let l:fname = fnamemodify(expand("%"), ':p:gs?\\?/?')
+  if filename != l:fname
```
fatih: We can early return here, but I'll do that with a follow up commit. Thanks

fatih:
Thanks @bhcleek, great addition 👍

PR #945, opened 2016-07-17
add commands to toggle settings that control auto behavior

bhcleek:
Add commands to toggle settings that control automatic behaviors. All the new commands begin with `GoToggle`, and are suffixed with the name of the setting, but using PascalCase in place of snake_case.

Also moved one function so that all functions called by events and functions that control settings are in the same relative order as the respective events.

### DIFF
--- a/autoload/go/asmfmt.vim
+++ b/autoload/go/asmfmt.vim
@@ -51,4 +51,15 @@ function! go#asmfmt#Format()
   call winrestview(l:curw)
 endfunction
 
+function! go#asmfmt#ToggleAsmFmtAutoSave()
+  if get(g:, "go_asmfmt_autosave", 1)
+    let g:go_asmfmt_autosave = 0
+    call go#util#EchoProgress("auto asmfmt disabled")
+    return
+  end
+
+  let g:go_asmfmt_autosave = 1
+  call go#util#EchoProgress("auto asmfmt enabled")
+endfunction
+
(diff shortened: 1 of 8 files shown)


bhcleek:
@fatih I think these new functions (e.g. s:toggle_auto_type_info()) should be moved to autoload/go/util.vim. At the time I submitted this, I didn't really know how autoloading worked. Do you want me to move them?

fatih:
Thanks @bhcleek for the work. I've wrote my comments below. Yes they should be moved to `autoload/go` but under different files to make it maintainable.

* Can you change the names and append the `Toggle` to the end? All our commands are like that so you can easily find them. 
* Can you add `enabled` and `disabled` echo messages. Check out `guru.vim` how I've added for sameids. I think each function should have a custom logic. So they should be in the associate files

`go_auto_sameids` -> `guru.vim` (this exists already under `go#guru#AutoSameIds`, so don't touch it)

These should be moved:

`go_fmt_autosave` -> `fmt.vim`
`go_asmfmt_autosave` -> `asmfmt.vim`
`go_metalinter_autosave` -> `lint.vim`
`go_template_autocreate` -> `template.vim`

* Can you please move the commands to `ftplugin/go/commands.vim`

bhcleek:
PTAL

bhcleek:
PTAL

fatih:
Thanks @bhcleek 👍

PR #946, opened 2016-07-18
use CursorHold instead of CursorMove for go_auto_sameids

bhcleek:


fatih:
I still think this is the way to do it :) What was the main reason you closed it?

bhcleek:
Yeah, I think it's the right way to do it, but I know you had another PR in the works, and I though this would just be a distraction.

bhcleek:
@fatih I re-opened since we both agree it's the right way to do it and a larger change that includes this hasn't been proposed yet.

fatih:
Thanks @bhcleek I'm going over the `:GoSameIds` PR's now. Maybe I can combine all of yours (with your commits) and merge them with couple of improvements I want to make.

bhcleek:
Yeah, they'll all merge together cleanly. Doing an octopus merge on them works great.

PR #948, opened 2016-07-18
guru: add :GoWhicherrs

krishicks:
Hi,

Alan Donovan at Gophercon 2016 showed that `guru whicherrs` is a thing, which I noticed was missing from vim-go. This is my attempt at adding it.

Cheers,

KH

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -176,6 +176,18 @@ function! go#guru#Tags(...)
   endif
 endfunction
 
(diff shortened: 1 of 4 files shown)


fatih:
Thanks @krishicks 👍 

There are couple improvements we need to make (such as displaying and handling the case where no errs are available and it results in empty output). I'll push those myself.

krishicks:
Thanks!

PR #949, opened 2016-07-19
Replace 'he=e-1' with '\ze'

ArkBriar:
Replace 'he=e-1' with '\ze' to avoid conflicts with other highlight plugins (e.g. luochen1990/rainbow).


fatih:
Thanks @ArkBriar

PR #952, opened 2016-07-20
Fixes the 'setfperm' issue

magicalbanana:
So after updating I found this issue where when I saved I would get an error of:

`setfperm` unknown function. This PR should fix that issue.

Thanks @fatih for letting me fix this. :D


fatih:
Thanks @magicalbanana 👍 ;)

PR #954, opened 2016-07-21
util: add go#util#env function

### LINKED ISSUE #901: Reinstall binaries when a GOROOT change is detected
When GOROOT changes, for example if you simply upgrade Go with Homebrew, all binaries compiled with the old compiler will end up with a stale GOROOT. This is confusingly breaks tools like gocode.

Maybe vim-go could (suggest to?) rerun InstallBinaries when a GOROOT change is detected.

See also: https://blog.filippo.io/stale-goroot-and-gorebuild/

fatih:
Add new function to return cached version of Go environment variables.
With that set GOROOT everytime we execute gocode. This will prevent
gocode using the internal embedded GOROOT and will use the GOROOT from
the latest installed Go version.

Fixes #901

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -32,7 +32,9 @@ function! s:gocodeCommand(cmd, preargs, args)
   " we might hit cache problems, as gocode doesn't handle well different
   " GOPATHS: https://github.com/nsf/gocode/issues/239
   let old_gopath = $GOPATH
+  let old_goroot = $GOROOT
   let $GOPATH = go#path#Detect()
+  let $GOROOT = go#util#env("goroot")
 
   let socket_type = get(g:, 'go_gocode_socket_type', s:sock_type)
   let cmd = printf('%s -sock %s %s %s %s', 
@@ -45,6 +47,8 @@ function! s:gocodeCommand(cmd, preargs, args)
 
   let result = go#util#System(cmd)
   let $GOPATH = old_gopath
+  let $GOROOT = old_goroot
+
   if go#util#ShellError() != 0
     return "[\"0\", []]"
   else
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -121,7 +121,9 @@ endfunction
 
 function! go#tool#ExecuteInDir(cmd) abort
   let old_gopath = $GOPATH
+  let old_goroot = $GOROOT
   let $GOPATH = go#path#Detect()
+  let $GOROOT = go#util#env("goroot")
 
   let cd = exists('*haslocaldir') && haslocaldir() ? 'lcd ' : 'cd '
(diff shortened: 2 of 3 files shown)


ashb:
I had to add this hunk to get it working for :GoRename too:

```diff
diff --git a/autoload/go/tool.vim b/autoload/go/tool.vim
index 487b933..fbbe12b 100644
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -121,7 +121,9 @@ endfunction
 
 function! go#tool#ExecuteInDir(cmd) abort
   let old_gopath = $GOPATH
+  let old_goroot = $GOROOT
   let $GOPATH = go#path#Detect()
+  let $GOROOT = go#util#env("goroot")
 
   let cd = exists('*haslocaldir') && haslocaldir() ? 'lcd ' : 'cd '
   let dir = getcwd()
@@ -133,6 +135,7 @@ function! go#tool#ExecuteInDir(cmd) abort
   endtry
 
   let $GOPATH = old_gopath
+  let $GOROOT = old_goroot
   return out
 endfunction
 
```

Are there any other paths that might need it to?

fatih:
I'm not sure why `:GoDoc` or `:GoRename` needs them. Do they also use compiled GOROOT information?

ashb:
I guess so - they seem to have problems if there's a syntax error. Trying to use `^]` I just got this error trying to jump to the code for `json.Marshal`

```
vim-go: guru: cannot find package "encoding/json" in any of:
        /Users/ash/.homebrew/Cellar/go/1.6.2/libexec/src/encoding/json (from $GOROOT)
        /Users/ash/code/go/src/encoding/json (from $GOPATH)
```

However addingthe fix-up into autoload/go/guru.vim didn't seem to help in this case. So maybe this won't work for all tools :(

(My test is I last ran :GoUpdateBinaries with go 1.6.2 and this morning I upgraded to go 1.6.3)

fatih:
@filosottile can you try this PR and see if this fixes #901 ?

fatih:
Another ping @ashb @filosottile

ashb:
I think it makes sense to include the patch form https://github.com/fatih/vim-go/pull/954#issuecomment-234195675 too - it helps for more of the tools, even if not every single one.

fatih:
@ashb good call, I've forget it. I'll try to add them.

fatih:
@ashb sorry for the late response. But I've included your patch as well. Does this look good now? Can you try it?

ashb:
This is an improvment and fixes some of the tools, if not all of them. Lets merge it.

(guru doesn't work because it doesn't call `go#tool#ExecuteInDir` but uses `go#uitl#System`. Probably worth fixing guru in another PR. Either to use ExecuteInDir or to save/restore GOROOT itself.)

PR #957, opened 2016-07-22
fmt: pass full name to -srcdir for upcoming new features

### LINKED ISSUE #956: goimports -srcdir will soon also accept a full filename
As discussed in https://github.com/dominikh/go-mode.el/issues/146 for emacs,

As of https://golang.org/cl/25140, `goimports -srcdir=....` will accept a full filename of the file being edited, instead of just its directory. (But directory will still work as it does today)

You'll probably want to update vim-go to pass the full name, which will enable some new features in goimports. (analyzing other files in your package's directory for better recommendations, and being aware of global variables)

fatih:
Closes #956


bradfitz:
LGTM. Thanks!

PR #958, opened 2016-07-22
go#fmt#Format fails on new unsaved files

datanoise:
make sure that we pass a valid parameter to setfperm function.


fatih:
Thanks @datanoise

PR #964, opened 2016-07-26
Add :GoAvailImports to search importable packages with CtrlP

buztard:
The :GoAvailImports command allows searching and importing available
packages using CtrlP.

### DIFF
--- a/autoload/ctrlp/imports.vim
+++ b/autoload/ctrlp/imports.vim
@@ -1,0 +1,82 @@
+let s:go_imports_var = {
+      \  'init':   'ctrlp#imports#init()',
+      \  'exit':   'ctrlp#imports#exit()',
+      \  'enter':  'ctrlp#imports#enter()',
+      \  'accept': 'ctrlp#imports#accept',
+      \  'lname':  'go imports',
+      \  'sname':  'imports',
+      \  'type':   'tabs',
+      \}
+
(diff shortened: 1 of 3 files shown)


buztard:
I still don't like the name of the command. Maybe `:GoPackages` might be better?

autoload/ctrlp/imports.vim
```diff
@@ -0,0 +1,82 @@
...
+  if empty(bin_path)
+    return
+  endif
+
+  " get a list of all available imports including package docs
+  let command = printf("%s list -f \"{{.ImportPath}}\t{{.Doc}}\" ...", bin_path)
+
+  let out = go#util#System(command)
+  if go#util#ShellError() != 0
+    call go#util#EchoError(out)
```
fatih: This is not good. For example this doesn't work for me because in my GOPATH there are files that don't get compiled. But because the command above calls `...` it searches all packages under my GOPATH. It fails. And when you echo all these, my whole Vim screen get black. I think we should come up with a faster and better solution. Did you think to implement it as a part of `motion`? https://github.com/fatih/motion 

I dont want to be look picky but I've explicitly said https://github.com/fatih/vim-go/blob/master/.github/CONTRIBUTING.md to open an issue before opening PR's like these. I like the idea but it's something we need to talk before :)

fatih:
Closing this as there is no response for a long time. We can always tackle it later. Thanks.

PR #966, opened 2016-07-27
Escape all template path special characters

thiagocaiubi:
Add `fnameescape()` to handle any special characters within a template path.

Example: usage of `hello_word.go` template with a space within path is broken.
screen shot 2016-07-27 at 9 31 45 am


fatih:
Thanks @thiagocaiubi 👍

PR #967, opened 2016-07-27
Reset shell before running a command in go#util#System

Carpetsmoker:
This makes sure the commands still work when using a non-Bourne shell such as
fish or tcsh.

:help vim-doc recommends:

        set shell='/bin/sh'

But this is not a great solution. This will mean that `:sh` will now invoke
`/bin/sh`, instead of `tcsh` or `fish`. It also means that things such as shell
aliases for `:!` will no longer work.

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -92,11 +92,11 @@ else
 endif
 
 function! go#util#System(str, ...)
-    let l:shell = &shell
-    let &shell = '/bin/sh'
-    let l:output = call(s:vim_system, [a:str] + a:000)
-    let &shell = l:shell
-    return l:output
+  let l:shell = &shell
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @Carpetsmoker. This sounds like good improvement.

Carpetsmoker:
Thanks @fatih

Maybe it's also a good idea to remove the part in the "TROUBLESHOOTING" section of the help file? This should no longer apply...

fatih:
@Carpetsmoker definitely. If you see something that needs to be changed/fixed feel free to open a PR. I'm usually fully loaded with PR's, fixes, my work, my personal time, etc.. So I have limited time. I'm more than happy for contributions 👍

PR #969, opened 2016-07-28
tool: use double quotes in PackageName for Windows compatibility

### LINKED ISSUE #959: quotes problem with the template
### Actual behavior

I'm on Windows.
I have a directory called clog, with a clog.go file containing `package clog`.

Then when I open another file foobar.go, vim-go auto creates the package line, but it's wrong: it creates _package 'clog'_.

### Expected behavior

vim-go should write the correct package line.

### Steps to reproduce:

Problem only appears on Windows, on my OS X machine it's fine.

 1. create a package foobar.
 2. create a new Go file in the same directory.

### Configuration

* gvim for windows 7.4
* last vim-go commit as of right now (9754508b38414a7544360bd6f2b33c9016517400).
* go tip

Also, I investigated the code and it looks like in https://github.com/fatih/vim-go/blob/master/autoload/go/tool.vim#L47 if I just remove the two single quote in the command it works.

Can't test if that change would work on Unixes too.

vrischmann:
Tested on Mac OS X using bash/zsh as Vim's shell, Linux with bash/zsh as Vim's shell and on Windows 10.

Fixes #959.

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -44,7 +44,7 @@ function! go#tool#Imports()
 endfunction
 
 function! go#tool#PackageName()
-  let command = "go list -f '{{.Name}}'"
+  let command = "go list -f \"{{.Name}}\""
   let out = go#tool#ExecuteInDir(command)
   if go#util#ShellError() != 0
       return -1


fatih:
Thanks @vrischmann 👍

PR #972, opened 2016-07-30
add SimpleAutoComplPop to the list

roxma:
I've been using SimpleAutoComplPop  for go programming for a while.

I believe it can be a simple alternative for realtime auto completion with vim-go.


fatih:
Ok I've tried it and it really works. However just a suggestion you should really make your `README` more simpler and readable. There is too much information and nobody cares about those honestly. You should just provide simple functions for every language and a way to let people define their own triggers. I'm adding this, let's see how people will like this :) But for your information, I'm going to remove most of the README because of https://github.com/fatih/vim-go-tutorial. Of course not yet, but eventually in the future.

PR #975, opened 2016-07-30
re-evaluate sameids after a rename

bhcleek:
Re-evaluate sameids after a rename. This is especially useful when `go:go_auto_sameids` is not set.

### DIFF
--- a/autoload/go/rename.vim
+++ b/autoload/go/rename.vim
@@ -72,6 +72,8 @@ function! go#rename#Rename(bang, ...)
   " we need a way to get the list of changes from gorename upon an success
   " change.
   silent execute ":e"
+
+  call go#guru#SameIdRenamed()
 endfunction
 
 " vim: sw=2 ts=2 et
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -1,5 +1,6 @@
 "  guru.vim -- Vim integration for the Go guru.
 
+let s:sameid = 0
 
 func! s:RunGuru(mode, format, selected, needs_scope) range abort
   "return with a warning if the binary doesn't exist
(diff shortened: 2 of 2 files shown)


fatih:
Hi @bhcleek. To be honest I didn't like this. I don't want that a feature is called because of other commands. If you really want this you should create custom vim function in your `vimrc` and use that. Otherwise this causes side effects. Every single command should just do what it's meant for. In long term code like this is not maintainable either. Feel free to reopen if you think we should not close this. Thanks.

bhcleek:
@fatih I think you may have misunderstood what this change does. It doesn't unconditionally highlight the identifier under the cursor after a rename; it only highlights the identifier under the cursor if it was already highlighted before the rename. Perhaps a demonstration will help:

master currently does this:
master

With this patch, the behavior is:
sameid-and-rename

The behavior on master seems wrong to me. The user has highlighted the identifier under the cursor and then renamed it. After renaming, I would expect the identifier under the cursor to be highlighted, but on master the length of the previous name of the identifier under the cursor has determined the highlight length.

fatih:
@bhcleek upon rethinking I think you are right. We can make vim-go a little bit smarter. Even though I believe it's the user's responsibility to clear before renaming, I assume there will be a lot of people doing this. It can't merged though, can you please rebase against master again ?

bhcleek:
PTAL

I'm not entirely sure the technique I've employed here is the right one. [`autoload/go/coverage.vim`](https://github.com/fatih/vim-go/blob/e7f5e5756e4722ec4998264a4344fa02c5820a75/autoload/go/coverage.vim#L239)  demonstrates another technique that may be preferable: hooking into the autocmd events for the current window. What would you think about doing something more like what is employed there? That way all the logic for dealing with the same id is contained in `autoload/go/guru.vim` instead of having something special over in `autoload/go/rename.vim`?

autoload/go/rename.vim
```diff
@@ -72,6 +72,8 @@ function! go#rename#Rename(bang, ...)
   " we need a way to get the list of changes from gorename upon an success
   " change.
   silent execute ":e"
+
+  call go#guru#SameIdRenamed()
```
fatih: I think we should do it different. There should be a function called `go#guru#HasSameId()` which returns a true if it's been highlighted. And then you call it like this:

```
if go#guru#HasSameId()
  " Call sameids again to redraw the highlighting  
  call go#guru#SameIds(-1)
endif
```

Does this makes sense? I didn't like the `SameIdRenamed()` function so I would be happy if you can remove it.

fatih:
> What would you think about doing something more like what is employed there? 

@bhcleek what do you have in mind? If you can show me a working example (maybe a different PR) I can take a look.

bhcleek:
After working up the other approach, I definitely like it better. It's a cleaner implementation and ensures the highlighting is applied correctly in a greater range of scenarios. I'm closing this in favor of #998.

PR #976, opened 2016-07-30
Add support for FZF for :GoDecls

wladh:
This pull request includes 2 fixes and one enhancement:
1) Added back guifg attributes for GoCoverage highlighting (this broke on colours for gvim and truecolors neovim)
2) Made GoInstall obey alternative $GOPATH, so the binaries get built correctly when godeps is used.
3) Added support for FZF in GoDecls and GoDeclsDir (FZF is an alternative to CtrlP).

### DIFF
--- a/ftplugin/go/commands.vim
+++ b/ftplugin/go/commands.vim
@@ -85,6 +85,12 @@ if globpath(&rtp, 'plugin/ctrlp.vim') != ""
   command! -nargs=? -complete=dir GoDeclsDir call ctrlp#init(ctrlp#decls#cmd(1, <q-args>))
 endif
 
+" -- fzf
+if globpath(&rtp, 'plugin/fzf.vim') != ""
+  command! -nargs=? -complete=file GoDecls call fzf#decls#cmd(0, <q-args>)
+  command! -nargs=? -complete=dir GoDeclsDir call fzf#decls#cmd(1, <q-args>)
+endif
+
 " -- impl
 command! -nargs=* -buffer -complete=customlist,go#impl#Complete GoImpl call go#impl#Impl(<f-args>)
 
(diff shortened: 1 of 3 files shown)


fatih:
These are too many features. I want to release `v1.8` today. I've cherry picked `1` and `2` from this PR as those are fixes. For FZF I need to test it and I don't want to add a big feature into the new version.

fatih:
@wladh can you rebase on top of master please.

wladh:
Done.

ftplugin/go/commands.vim
```diff
@@ -85,6 +85,12 @@ if globpath(&rtp, 'plugin/ctrlp.vim') != ""
   command! -nargs=? -complete=dir GoDeclsDir call ctrlp#init(ctrlp#decls#cmd(1, <q-args>))
 endif
 
+" -- fzf
+if globpath(&rtp, 'plugin/fzf.vim') != ""
+  command! -nargs=? -complete=file GoDecls call fzf#decls#cmd(0, <q-args>)
+  command! -nargs=? -complete=dir GoDeclsDir call fzf#decls#cmd(1, <q-args>)
+endif
```
fatih: We need to provide an options, such as `go_decls_mode` with the values of `ctrlp` and `fzf`. Otherwise if you have installed both `fzf` will override it.

fatih:
Hi @wladh 

Sorry I just had the ability to check it. Can you this new option `go_decls_mode` and make `ctrlp` the default. If `ctrlp` doesn't exist but `fzf` is there then `go_decls_mode` should be defaulted to `fzf`

fatih:
ping @wladh

mixedCase:
Shouldn't one want fzf to be the default if it's installed, given that it's leaps and bounds faster than CtrlP?

fatih:
I'm closing this. Didn't get any repose. If anyone wants to have `fzf` support feel free to open a new PR with my comments as well. Thanks

@mixedCase ctrlp is maintained and is still in use. Fzf on the other hand was introduced later.

ftplugin/go/commands.vim
mvalkon: @fatih can you point to somewhere in the codebase where a similar option was implemented. I'd love to see support for fzf in vim-go and am willing to contribute. I have little experience with vim-scripting so having an example would help.

PR #983, opened 2016-08-01
Fix :GoSameIds for light background, GUI, and after changing colour schemes

Carpetsmoker:
This does three things:

- For light backgrounds the "highlighted" text is typically the same colour as
  the normal text:  2016-08-01-060201_1920x1080_scrot
  So use the `background` setting to set a more appropriate colour.
  I know I can set my own colour, but it would IMHO be better if it just
  works by default.

- The second issue is that most (if not all) colour schemes do this at the
  start:

		" Remove all existing highlighting and set the defaults.
		hi clear

  Which resets all custom highlight groups (except for `:hi link`). So just
  using a "bare" `:hi ctermfg=...` won't work after changing colour schemes:

		:hi goSameId
		goSameId       xxx term=bold cterm=bold ctermfg=0 ctermbg=121

		:colorscheme default

		:hi goSameId
		goSameId       xxx cleared

  The solution is to hook into the `ColorScheme` autocommand.

- And it sets the `guifg`/`guibg` parameters, otherwise nothing gets highlighted when using Gvim...

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -366,13 +366,27 @@ if g:go_highlight_build_constraints != 0
   hi def link goPackageComment    Comment
 endif
 
-" :GoSameIds
-hi def goSameId term=bold cterm=bold ctermbg=white ctermfg=black
-
 " :GoCoverage commands
 hi def link goCoverageNormalText Comment
-hi def      goCoverageCovered    ctermfg=green guifg=#A6E22E
-hi def      goCoverageUncover    ctermfg=red guifg=#F92672
+
+function! s:hi()
+  " :GoSameIds
+  if &background == 'dark'
+    hi def goSameId term=bold cterm=bold ctermbg=white ctermfg=black
+  else
+    hi def goSameId term=bold cterm=bold ctermbg=lightgreen ctermfg=black
+  endif
+
+  " :GoCoverage commands
+  hi def      goCoverageCovered    ctermfg=green guifg=#A6E22E
+  hi def      goCoverageUncover    ctermfg=red guifg=#F92672
+endfunction
+
+augroup vim-go-hi
+  autocmd!
+  autocmd ColorScheme * call s:hi()
+augroup end
+call s:hi()
 
 " Search backwards for a global declaration to start processing the syntax.
 "syn sync match goSync grouphere NONE /^\(const\|var\|type\|func\)\>/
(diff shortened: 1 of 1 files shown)


syntax/go.vim
```diff
@@ -366,13 +366,27 @@ if g:go_highlight_build_constraints != 0
...
+  endif
+
+  " :GoCoverage commands
+  hi def      goCoverageCovered    ctermfg=green guifg=#A6E22E
+  hi def      goCoverageUncover    ctermfg=red guifg=#F92672
+endfunction
+
+augroup vim-go-hi
+  autocmd!
+  autocmd ColorScheme * call s:hi()
```
fatih: Not sure we apply ColorScheme to all files? And why do we have it?

Carpetsmoker: The pattern for the `ColorScheme` autocommand is matched against the colour scheme name, not the file name (see `:help ColorScheme`).

fatih: Alrigth, back to my question. Why do we need to call this automatically?

Carpetsmoker: It's needed because using `:colorscheme foo` will reset all custom highlight groups − see the message above. It's a bit ugly, but as far as I know there's no "clean" way to avoid this...

fatih: I'm trying to understand the case. So you mean we have to add this for users who override it? `colorscheme` is only called once in `vimrc`. Or is this for people who dynamically change the colorscheme ?

Carpetsmoker: There are a few different scenarios:

- Perhaps the most common case is when you reload the vimrc file with `:source ~/.vim/vimrc`. This will run the `colorscheme` command but won't re-run the syntax files (you can use `:e` to fix this though, but it's annoying).

- I've seen some people use different colour schemes for different filetypes − this autocommand will be run after the syntax file and there is no real way to fix this other than than the above `ColorScheme` autocommand.

- As you mentioned, it helps people who change colour schemes while Vim is running. I know some people who like to switch between a dark and light version depending on ambient light conditions. Again, `:e` will fix this, but it's annoying.

bhcleek:
This would be great. I change my colorscheme using `:colorscheme` often to accommodate people that are looking at my screen for one reason or another.

fatih:
Thanks @Carpetsmoker

PR #988, opened 2016-08-01
More robust resetting of the shell.

### LINKED ISSUE #986: go#util#System is broken on Windows
### Actual behavior

First symptom is that :GoFmt no longer works, I get a cryptic error:

    Error detected while processing function <SNR>12_fmt_autosave..go#fmt#Format..go#util#System:
    line    3:
    E484: Can't open file C:\Users\Sphax\AppData\Local\Temp\VIoCBDF.tmp

I thought I screwed up something in my vimrc regarding temp directories or something, but I checked the usual stuff and it wasn't that.

By debugging the `go#fmt#Format` function I managed to track down the problem to the `go#util#System` function. I then checked the git history of that file and saw #967 which changed it, so I tried reversing the change and everything works fine.

I'm guessing `/bin/sh` won't work on Windows (my shell is cmd.exe).

### Expected behavior

No errors.

### Steps to reproduce:

1. Open any .go file.
2. run :GoFmt (or save)

### Configuration

Minimal vimrc used:

    filetype off
    set rtp+=~/.vim/bundle/Vundle.vim
    call vundle#rc()

    Plugin 'VundleVim/Vundle.vim'
    Plugin 'fatih/vim-go'

    filetype plugin indent on
    syntax enable

Vim 7.4 for Windows, vim-go commit 3030ecc86602b460cb4c1fcbd4ed210f32498c1c

Carpetsmoker:
Don't reset it on Windows, and only reset it if `/bin/sh` is executable (which
should pretty much always be the case, but it's not in POSIX and IIRC POSIX
explicitly advises against relying on it).

This is basically the same method that vim-plug uses.

This improves #967 and should fix #986.

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -95,7 +95,7 @@ endif
 " systems if it is executable.
 function! go#util#System(str, ...)
   let l:shell = &shell
-  if !(has('win32') || has('win64')) && executable('/bin/sh')
+  if !go#util#IsWin() && executable('/bin/sh')
     let &shell = '/bin/sh'
   endif
 
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -91,12 +91,20 @@ else
   let s:vim_shell_error = ''
 endif
 
+" System runs a shell command. It will reset the shell to /bin/sh for Unix-like
+" systems if it is executable.
 function! go#util#System(str, ...)
   let l:shell = &shell
-  let &shell = '/bin/sh'
-  let l:output = call(s:vim_system, [a:str] + a:000)
-  let &shell = l:shell
-  return l:output
+  if !(has('win32') || has('win64')) && executable('/bin/sh')
+    let &shell = '/bin/sh'
+  endif
+
+  try
+    let l:output = call(s:vim_system, [a:str] + a:000)
+    return l:output
+  finally
+    let &shell = l:shell
+  endtry
 endfunction
 
 function! go#util#ShellError()


autoload/go/util.vim
```diff
@@ -91,12 +91,20 @@ else
...
 
+" System runs a shell command. It will reset the shell to /bin/sh for Unix-like
+" systems if it is executable.
 function! go#util#System(str, ...)
   let l:shell = &shell
-  let &shell = '/bin/sh'
-  let l:output = call(s:vim_system, [a:str] + a:000)
-  let &shell = l:shell
-  return l:output
+  if !(has('win32') || has('win64')) && executable('/bin/sh')
```
fatih: Use `go#util#IsWin()` here

fatih:
Maybe this is just easier (I was going to push it now, but you opened first :)):


```viml
function! go#util#System(str, ...)
  if go#util#IsWin()
    let l:output = call(s:vim_system, [a:str] + a:000)
    return l:output
  endif

  let l:shell = &shell
  let &shell = '/bin/sh'
  let l:output = call(s:vim_system, [a:str] + a:000)
  let &shell = l:shell
  return l:output
endfunction
```

autoload/go/util.vim
```diff
@@ -91,12 +91,20 @@ else
...
+  if !go#util#IsWin() && executable('/bin/sh')
+    let &shell = '/bin/sh'
+  endif
+
+  try
+    let l:output = call(s:vim_system, [a:str] + a:000)
+    return l:output
+  finally
+    let &shell = l:shell
+  endtry
```
fatih: I'm not sure why we have this `try` clause. Why not just : 

```viml
let l:output = call(s:vim_system, [a:str] + a:000)
let &shell = l:shell
return l:output
```

Carpetsmoker:
Well, in theory there could be systems that don't have `/bin/sh`... Those systems are probably very rare, but there's nothing wrong with being paranoid and doing an extra check can't hurt ;-)

The `try`/`finally` construct is better in case there are errors. Usually functions continue running commands unless the `abort` keyword is given, so not re-setting the shell back to `l:shell` *should* never happen ... But are you *sure* there are no situations where Vim doesn't stop running the function on errors? Using `finally` should guarantee this more...

fatih:
> But are you sure there are no situations where Vim doesn't stop running the function on errors? Using finally should guarantee this more...

So you mean `call()` might return an error , and in that case we're not able to set the shell back?

Carpetsmoker:
> >  But are you sure there are no situations where Vim doesn't stop running the function on errors? Using finally should guarantee this more...

> So you mean call() might return an error , and in that case we're not able to set the shell back?

Well, more like the `system()` function, but yeah.

Like I said, I'm not sure if this really makes a difference... Both versions are probably fine.

fatih:
Thanks @Carpetsmoker

PR #998, opened 2016-08-09
Re-evaluate sameids after buffer is re-entered

bhcleek:
When g:go_auto_sameids is not set, re-apply :GoSameIds after the buffer
is displayed in a window. This ensures that once the user has called
:GoSameIds, the highlighting will be applied in the buffer based on the
cursor position at the time of evaluation until the user calls
:GoSameIdsClear, even if the buffer is reloaded into the same window (
e.g. :GoRename, or even reloading the file from disk manually).

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -1,6 +1,5 @@
 "  guru.vim -- Vim integration for the Go guru.
 
-
 func! s:RunGuru(mode, format, selected, needs_scope) range abort
   "return with a warning if the binary doesn't exist
   let bin_path = go#path#CheckBinPath("guru") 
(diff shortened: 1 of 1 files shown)


autoload/go/guru.vim
```diff
@@ -364,6 +363,12 @@ function! go#guru#SameIds(selected)
     let pos = split(item, ':')
     call matchaddpos('goSameId', [[str2nr(pos[-2]), str2nr(pos[-1]), str2nr(poslen)]])
   endfor
+
+  if get(g:, "go_auto_sameids", 0)
+    " re-apply SameIds at the current cursor position at the time the buffer
+    " is redisplayed: e.g. :edit, :GoRename, etc.
+    autocmd BufWinEnter <buffer> nested call go#guru#SameIds(-1)
```
fatih: So this is called because we call `:e` inside `:GoRename`, which loads the buffer and thus it triggers this line right ? I wonder if there is a case that doesn't remove the trigger below.

bhcleek: That's exactly right. But the nice thing about this approach, is that it will do the right thing even the user does `:edit`.

I don't think there's a case that fails to remove the buffer-local autocommand; the only case I could think of is when the buffer is wiped out, but in that case the buffer-local autocommand will be removed by Vim.

fatih:
Alright, let us use this a little bit, if there is any errors we'll catch them up. Thanks @bhcleek 👍

bhcleek:
Sounds good to me! Thanks, @fatih.

PR #1004, opened 2016-08-12
Provide a descriptive quickfix/location list title.

datanoise:
Currently, the `w:quickfix_title` is being automatically set to the actual
Vim command that populates quickfix/location list. For example, running
any Guru command it will be displayed as `:    lgetexpr a:items` which
is not really informative.

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -235,7 +235,7 @@ function! go#guru#Implements(selected) abort
         \ 'needs_scope': 1,
         \ }
 
-  call s:run_guru(args, 'Implements')
+  call s:run_guru(args)
 endfunction
 
 " Report the possible constants, global variables, and concrete types that may
@@ -254,7 +254,7 @@ function! go#guru#Whicherrs(selected) abort
   "   call go#util#EchoSuccess("no error variables found. Try to change the scope with :GoGuruScope")
   "   return
   " endif
-  call s:run_guru(args, 'Whicherrs')
+  call s:run_guru(args)
 endfunction
 
 " Describe selected syntax: definition, methods, etc
@@ -266,7 +266,7 @@ function! go#guru#Describe(selected) abort
         \ 'needs_scope': 1,
         \ }
 
-  call s:run_guru(args, 'Describe')
+  call s:run_guru(args)
 endfunction
 
 function! go#guru#DescribeInfo() abort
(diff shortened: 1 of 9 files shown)


fatih:
Can you show an after/before screenshot? For example I've introduced a syntx error, but :GoFmt was not able to parse it and it errored. And a :GoBuild just displayed a normal quickfix list without any name.

datanoise:
It depends on if you are using `%q` in your `statusline` configuration. `%q` will display the type of the quickfix window (Quickfix or Location List) and the command that was used to generate this list.

`GoBuild` displays the title correctly:

```
[Quickfix List] :go build . errors
```

but GoFmt would display:

```
[Location List] :setloclist()
```

because the location list was set explicitly with `setloclist()` function without specifying any title.

fatih:
@datanoise This is something I want to merge definitely. I have some tasks on the vim-8.0 branch, I'll take another stab once I've merged it to master. Thanks for the work. I'll merge it if there is no problem or try to give a feedback if needed any.

fatih:
@datanoise can you please rebase against master if possible?

autoload/go/guru.vim
```diff
@@ -106,7 +106,7 @@ function! s:guru_cmd(args) range abort
 endfunction
 
 " sync_guru runs guru in sync mode with the given arguments
-function! s:sync_guru(args) abort
+function! s:sync_guru(args, title) abort
```
datanoise: ah, good point. I'll fix it.

autoload/go/guru.vim
```diff
@@ -235,7 +235,7 @@ function! go#guru#Implements(selected) abort
         \ 'needs_scope': 1,
         \ }
 
-  call s:run_guru(args)
+  call s:run_guru(args, 'Implements')
```
fatih: all `run_guru` functions should be fixed as well.

fatih:
Just tested it locally, but it doesn't work for me. This is what I got for `:GoBuild`

image

Can you please search for `Populate(...)` functions. For example there is one in `job.vim` which is the case for my error.

datanoise:
it should be fixed now.

fatih:
I've pulled the changes and tried `:GoMetaLinter` for the following code: https://play.golang.org/p/lZwwR-vXQl

However I don't see `:GoMetaLinter` in the title yet:

screen_shot_2016-11-24_at_10_41_04_pm

datanoise:
I'll check it later today. I suspect that there are still cases where I didn't propagate the title.

fatih:
Thanks @datanoise 👍

PR #1007, opened 2016-08-14
Highlight nested fields or chained fields

magicalbanana:
Currently, nested or chained fields are not getting highlighted.

screen shot 2016-08-14 at 12 53 42 pm

This will change that to this:

screen shot 2016-08-14 at 12 54 21 pm


fatih:
Thanks @magicalbanana

PR #1009, opened 2016-08-15
Fix appropriate bin_path may not work

chengwhy:
appropriate bin_path may not work, because $PATH will restore in some case.


fatih:
Why do we need it. We also prepend the gobin path before checking it. Can you give me an example what the real error is?

fatih:
Closing as I didn't receive any answer in one month

PR #1010, opened 2016-08-15
Add compatibility for GoAddTags and google/jsonapi

carlisia:
Not sure if it would be desirable to have library specific features like this?

This is the result when using `jsonapi` with the `:GoAddTags` command:
```
type User struct {
	ID  int  `jsonapi:"primary,id"`
	Name  string  `jsonapi:"attr,name"`
}
```

One caveat: the user will still need to manually change `id` to the name of the resource, in this case, like so:

```
	ID  int  `jsonapi:"primary,users"`
```

But it's still a lot of savings on typing if using this library.

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -276,6 +276,13 @@ function! go#util#AddTags(line1, line2, ...)
     endif
 
     let word = go#util#snippetcase(l:matched)
+    if l:keys == ["jsonapi"]
+      let prefix = "attr"
+      if word == "id"
+        let prefix = "primary"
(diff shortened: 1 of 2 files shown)


fatih:
Hi @carlisia,

Thanks for the contribution. The `:GoAddTags` has some feature requests that I'm going to work, and one of it's is to change the values for a selected field. This PR is too specific and not something general I can include. I hope once we have the `:GoAddTags` feature you'll be manipulate it easier, until then I suggest to change it in your source manually. Thanks!

PR #1011, opened 2016-08-16
Fix syntax goFormatSpecifier: check '%%'

ArkBriar:
It should not highlight format specifiers in strings like "something %%c".


fatih:
Thanks @ArkBriar

PR #1017, opened 2016-08-19
SameIds: Fix Sentence, add hint to 'updatetime'

fd0:
I turned on `g:go_auto_sameids` and nothing happened, in contrast to your tutorial. Then I investigated and found out that the cursor needs to rest at the same position at least for `updatetime` ms, which means 4s by default. That's way too long for me.

I've added a hint for `updatetime` in the manual, and thereby fixed a sentence that was incomplete.


fatih:
This looks good. Thanks @fd0

PR #1020, opened 2016-08-22
Rename GoMetalinterAutoSaveToggle to GoMetaLinterAutoSaveToggle

kaneshin:
Hi!

I've changed the `:GoMetalinterAutoSaveToggle` command name to become `Linter` because the command calls `go#lint#ToggleMetaLinterAutoSave()` which has already defined `MetaLinter`.
So, I just chose "`MetaLinter`" to the command name.

Thanks!

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -695,8 +695,8 @@ CTRL-t
 
     Toggles |'g:go_asmfmt_autosave'|.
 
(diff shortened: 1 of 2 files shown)


fatih:
This is a breaking change, but seems like we did it wrong in the first place. Nonetheless, I'll merge it because it's the correct thing. Thanks @kaneshin

PR #1024, opened 2016-08-25
Add go_highlight_extra_vars option

yyoshiki41:
## *Proposal*

In idiomatic Go code,
`ok` and `err` are commonly used.

So I add a highlight option for them.
(_By default it's disabled._)

## After

----

If you enable this setting

```vim
let g:go_highlight_extra_vars = 1
``` 

Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/fatih/vim-go/1024)

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1245,6 +1245,13 @@ By default it's enabled. >
 
   let g:go_highlight_format_strings = 1
 <
+                                                *'g:go_highlight_extra_vars'*
+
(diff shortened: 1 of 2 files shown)


fatih:
Hi @yyoshiki41. I don't want to include this as these are just ordinary variables. Thank for your contribution though!

yyoshiki41:
@fatih 
Thanks for your reply.
I'll follow your thoughts!

Issue #1026, opened 2016-08-25
ctrl+o broken after :GoDef


dmac:
### Actual behavior

After using :GoDef (or gd from normal mode) to jump to a definition, using ctrl+o to jump back to the previous location will often not work, sending you to the wrong location.

### Expected behavior

Using ctrl+o after gd should send you back to the previous location. It should be possible to gd then ctrl+o repeatedly to move back and forth between a usage and definition.

### Steps to reproduce:

foo.go:

```go
package main

func main() {
	foo()
}

func foo() {}
```

Repro:

1. Move to the `foo()` call in `main`.
2. Press `gd` and `ctrl+o` repeatedly to try to move back and forth between the usage and definition of `foo`.

I bisected and found 445e5bf10ed269665f766ce198671ed6ebbfbb0d to be the first commit where this repros for me.

### Configuration

* .vimrc:

```viml
set nocompatible
filetype off

set rtp+=~/.vim/bundle/Vundle.vim
call vundle#begin()

Plugin 'gmarik/Vundle.vim'
Plugin 'fatih/vim-go'

call vundle#end()
filetype plugin indent on
syntax on
```

* vim version:

```
$ vim --version
VIM - Vi IMproved 7.3 (2010 Aug 15, compiled Jun 14 2016 16:06:49)
Compiled by root@apple.com
Normal version without GUI.  Features included (+) or not (-):
-arabic +autocmd -balloon_eval -browse +builtin_terms +byte_offset +cindent
-clientserver -clipboard +cmdline_compl +cmdline_hist +cmdline_info +comments
-conceal +cryptv +cscope +cursorbind +cursorshape +dialog_con +diff +digraphs
-dnd -ebcdic -emacs_tags +eval +ex_extra +extra_search -farsi +file_in_path
+find_in_path +float +folding -footer +fork() -gettext -hangul_input +iconv
+insert_expand +jumplist -keymap -langmap +libcall +linebreak +lispindent
+listcmds +localmap -lua +menu +mksession +modify_fname +mouse -mouseshape
-mouse_dec -mouse_gpm -mouse_jsbterm -mouse_netterm -mouse_sysmouse
+mouse_xterm +multi_byte +multi_lang -mzscheme +netbeans_intg -osfiletype
+path_extra -perl +persistent_undo +postscript +printer -profile +python/dyn
-python3 +quickfix +reltime -rightleft +ruby/dyn +scrollbind +signs
[...]

PR #1029, opened 2016-08-26
update jumplist in :GoDef

### LINKED ISSUE #1028: Goto declaration doesn't update the location stack correctly anymore
### Actual behavior

In a normal go source file I use the `gd` command to go to the definition of a symbol.  This works fine, but when I attempt to go back using `Ctrl-O` it takes me to other places that make no sense.

### Expected behavior

`gd` + `Ctrl-O` should cancel out.  This used to work, but I can't tell you how far back it used to work.

### Steps to reproduce:

Minimal vimrc (in neovim):
```
Plug 'fatih/vim-go'
call plug#end()
```

Minimal go source file to see the problem:
```
package main

import "fmt"

const iso = "blah"

func main() {
	fmt.Printf("iso=%q", iso)
}
```

1. Manually navigate to the `Printf` line with your cursor on the `iso` variable.
2. Press `gd` to go to the definition of the constant.
3. Press `Ctrl-O` and the cursor returns to the first line of the file.

### Configuration

Add here your current configuration and additional information that might be
useful, such as:

* vim version: `NVIM 0.1.6-dev` (ubuntu 16.04 + neovim PPA)
* vim-go version `63bdcdf65b10a2080d2a534879177bc254e2761e`
* go version `1.7`

nhooyr:
Closes #1028 and #1026

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -113,8 +113,8 @@ function! s:jump_to_declaration(out, mode)
   " modes of switchbuf which we need based on the split mode
   let old_switchbuf = &switchbuf
 
-  let l:fname = fnamemodify(expand("%"), ':p:gs?\\?/?')
-  if filename != l:fname
+  normal! m'
+  if filename != fnamemodify(expand("%"), ':p:gs?\\?/?')
     " jump to existing buffer if, 1. we have enabled it, 2. the buffer is loaded
     " and 3. there is buffer window number we switch to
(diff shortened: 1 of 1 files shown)


nhooyr:
@dmac this look good?

dmac:
Yep, this works for me!

fatih:
Thanks @nhooyr for the fix 👍

PR #1030, opened 2016-08-27
Exclude iota and nil from goBoolean

yyoshiki41:
exclude iota and nil from `goBoolean`
It would be better to treat them separately.

## Before
before

## After
after

Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/fatih/vim-go/1030)

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -125,14 +125,14 @@ hi def link     goComplexes         Type
 
 
 " Predefined functions and values
-syn match       goBuiltins          /\<\v(append|cap|close|complex|copy|delete|imag|len)\ze\(/
-syn match       goBuiltins          /\<\v(make|new|panic|print|println|real|recover)\ze\(/
-syn keyword     goPredefinedVars    nil, iota
-syn keyword     goBoolean           true false
+syn match       goBuiltins                 /\<\v(append|cap|close|complex|copy|delete|imag|len)\ze\(/
+syn match       goBuiltins                 /\<\v(make|new|panic|print|println|real|recover)\ze\(/
+syn keyword     goPredefinedIdentifiers    nil, iota
+syn keyword     goBoolean                  true false
 
-hi def link     goBuiltins          Keyword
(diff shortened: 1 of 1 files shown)


syntax/go.vim
```diff
@@ -127,9 +127,11 @@ hi def link     goComplexes         Type
 " Predefined functions and values
 syn match       goBuiltins          /\<\v(append|cap|close|complex|copy|delete|imag|len)\ze\(/
 syn match       goBuiltins          /\<\v(make|new|panic|print|println|real|recover)\ze\(/
-syn keyword     goBoolean           iota true false nil
+syn keyword     goPredefinedVars    nil, iota
```
nhooyr: `goPredefinedVars` should be renamed to `goPredefinedIdentifiers`

yyoshiki41: thanks!

https://github.com/fatih/vim-go/pull/1030/commits/d62310922d24f944b1478a39a3cae65be55894fc

fatih:
Thanks @yyoshiki41, this looks great 👍

Carpetsmoker:
Very annoying change. Suddenly my `nil` has a different colour from everything else. There is no reason it has to.

For those looking to fix this Christmas tree nonsense:

    :highlight link goPredefinedIdentifiers goBoolean

nhooyr:
I think we should just link `goPredefinedIdentifiers` to `Boolean` by default.

fatih:
@Carpetsmoker  to be honest I didn't like it either. Maybe we should add it as another flag (just like we do with others). I'm open to merge a PR that put's it behind flag.

nhooyr:
@fatih why do we need a flag? Just do `highlight link goPredefinedIdentifiers goBoolean` by default. If anyone wants a different color they can.

fatih:
@nhooyr oh I got what you mean. Mind adding a PR now? If not I'll can do it later.

nhooyr:
Can't, in class :(

PR #1033, opened 2016-09-02
coverage: allow to run coverage on _test.go files in subdirs

pborzenkov:
Due to usage of ':t' modifier to extract filename when switching from
file_test.go to file.go, the original path to file is lost and Vim can't
open it if it's stored in some subdir. It only worked by accident if the
file was located in the same directory.

### DIFF
--- a/autoload/go/coverage.vim
+++ b/autoload/go/coverage.vim
@@ -204,7 +204,7 @@ function! go#coverage#overlay(file)
     let cnt += 1
   endwhile
 
-  let fname = expand('%:t')
+  let fname = expand('%')
 
   " when called for a _test.go file, run the coverage for the actuall file
   " file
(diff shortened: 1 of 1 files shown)


fatih:
Hi @pborzenkov 

Thanks for the fix. Much appreciated. Unfortunately I couldn't understand your exact workflow and what this fixes. Can you please provide a step to step example I can try myself to see what the problem is?

pborzenkov:
@fatih Whenever I open test file that is not located in current dir :GoCoverage doesn't work. For example,

vim path/to/some/file_test.go
After that :GoCoverate doesn't work because it's trying to open file.go instead of path/to/some/file.go

fatih:
@pborzenkov this looks good. Sorry for late review, I was abroad and was very busy.

PR #1035, opened 2016-09-05
Fix bug in syntax highlighting

yyoshiki41:
@fatih 

Sorry,
https://github.com/fatih/vim-go/pull/1030 has a bug.

Please, merge this pr 🙇 

thanks.


fatih:
Thanks for the quick fix @yyoshiki41. I should have try before merging, but things like this can happen, no worries ;)

PR #1047, opened 2016-09-13
Fix reference to go_highlight_{structs,interface}

abhinav:
This replaces a reference to `go_highlight_{structs,interface}` in the doc
with `go_highlight_types`.


fatih:
Thanks @abhinav

PR #1052, opened 2016-09-17
fix default sock type

mattn:
Windows doesn't work with `unix`. ref https://github.com/fatih/vim-go/pull/1000


mattn:
ping. Many go developer on windows may be confusing because vim-go doesn't work on windows now.

fatih:
@mattn I'm on vacation. I just saw this. I'll be online on monday. I'm merging now this, but can look again later next week. Thanks for the fix.

mattn:
Thanks!

PR #1055, opened 2016-09-20
Add `g:go_updatetime` to configure a custom `updatetime`

svanharmelen:
Currently this is hardcoded if you use `g:go_auto_type_info` and/or `go_auto_sameids`, but the docs clearly refer to the standard vim `updatetime` setting.

With this hardcoded setting of 800ms, the user configured value is overwritten which is unexpected and sometimes even not wanted.

### DIFF
--- a/ftplugin/go.vim
+++ b/ftplugin/go.vim
@@ -57,7 +57,7 @@ if get(g:, "go_textobj_enabled", 1)
 endif
 
 if get(g:, "go_auto_type_info", 0) || get(g:, "go_auto_sameids", 0)
-  setlocal updatetime=800
+  let &l:updatetime= get(g:, "go_updatetime", 800)
 endif
 
 " vim: sw=2 ts=2 et
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -989,7 +989,8 @@ with |:GoPlay|. By default it's enabled. >
 
 Use this option to show the type info (|:GoInfo|) for the word under the
 cursor automatically. Whenever the cursor changes the type info will be
-updated. By default it's disabled.
+updated. By default it's disabled. The delay can be configured with the
+'g:go_updatetime' setting.
 >
   let g:go_auto_type_info = 0
 <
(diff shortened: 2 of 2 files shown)


fatih:
This shouldn't update the setting in your `vimrc`, are you sure ? Because I've tested it and set it to `300` and it picked that up.

svanharmelen:
Well.. It overrides it for Go source files, right?

I currently have `set updatetime=500` in my `vimrc` and if I open a Go source file and do `:set updatetime?` it shows me 800 and not the expected 500.

fatih:
You're right. Seems like it's not picked up. However the default value is just bad. 4s seconds is too much to make both sameids or auto type info usable. I think a better solution to make this a setting, such as `g:go_updatetime = 800` with the appropriate doc. If the user set this to `0` then it'll be disabled. 

With vim-8.0 I'm going to completely move forward using timers and jobs, so it'll be not a problem. Actually I might start working on it soon.

svanharmelen:
Yeah, I also thought about something like `g:go_updatetime`, I'll update this PR accordingly...

Hope the vim-8.0 changes will also be compatible with neovim? Where you able to abstract things is such a way you can accommodate both? That would be really awesome 😁

svanharmelen:
@fatih I updated the PR according to our earlier conversation. Could you have a look if this inline with what you had in mind? Thx...

fatih:
@svanharmelen I've tried this now and it gives the following error:

```
E521: Number required after =: updatetime=g:go_updatetime
```

svanharmelen:
@fatih sorry about that, fixed it by using `let &l:` which allows using an expression and for the rest is effectively the same as `setlocal` (see `:help let-&` for more details).

ftplugin/go.vim
```diff
@@ -56,8 +56,12 @@ if get(g:, "go_textobj_enabled", 1)
...
 
-if get(g:, "go_auto_type_info", 0) || get(g:, "go_auto_sameids", 0)
-  setlocal updatetime=800
+if !exists("g:go_updatetime")
+  let g:go_updatetime = 800
+endif
+
+if g:go_updatetime && get(g:, "go_auto_type_info", 0) || get(g:, "go_auto_sameids", 0)
+  let &l:updatetime=g:go_updatetime
 endif
```
fatih: Can you change this to this:

```
if get(g:, "go_auto_type_info", 0) || get(g:, "go_auto_sameids", 0)
  let &l:updatetime= get(g:, "go_updatetime", 800)
endif
```

Then you don't need this part:

```
if !exists("g:go_updatetime")
  let g:go_updatetime = 800
endif
```

svanharmelen:
Done...

fatih:
Thanks @svanharmelen 👍

svanharmelen:
Thanks you too ;)

PR #1061, opened 2016-09-26
Use transparent background for image.

egonelbre:


fatih:
Omg, thanks a lot @egonelbre 👍 . This is way better now 😅

PR #1067, opened 2016-10-04
Clear most recent goSameId highlights groups

### LINKED ISSUE #1041: Highlights from :GoSameIds remain across files
### Actual behavior

Highlights created by `:GoSameIds` in a `.go` file remain at the same positions even after opening a different file.

### Expected behavior

The highlights must be hidden when switching to other files and must re-appear when switching back to the initially highlighted file.

### Steps to reproduce:

1. Open a `.go` file.
2. Highlight any identifier with `:GoSameIds`.
3. Open some other file in the same window.

### Configuration

* vim version: 7.4
* vim-go version: 8575d9e
* go version: 1.7.1

bhcleek:
Clear the most recent goSameId highlight groups on BufWinEnter so that
the highlight groups do not persist when a different buffer is loaded
into a window.

Fixes #1041

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -200,6 +200,10 @@ augroup vim-go
   autocmd BufWritePre *.s call s:asmfmt_autosave()
   autocmd BufWritePost *.go call s:metalinter_autosave()
   autocmd BufNewFile *.go call s:template_autocreate()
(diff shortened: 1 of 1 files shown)


fatih:
Thanks for the fix @bhcleek  👍

PR #1069, opened 2016-10-04
Moved the predifined functions since order matters

svanharmelen:
@fatih I needed to move the predefined functions as order matters here (they are both using syc-match so last match wins) to make sure they are still highlighted correctly.

Sorry I missed that in the initial PR!


svanharmelen:
@fatih I have a much better solution for this one :wink:

PR #1070, opened 2016-10-04
Make sure builtins are always hightlighted correctly

svanharmelen:
And (again for consistency) change `goMethod` to `goMethodCall` to be consistent with `goFunctionCall`.

@fatih this is of course a much cleaner and better way to support all use cases correctly.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -302,7 +302,7 @@ if g:go_highlight_functions != 0
   syn match goPointerOperator   /\*/ nextgroup=goReceiverType contained skipwhite skipnl
   syn match goReceiverType      /\w\+/ contained
   syn match goFunction          /\w\+/ contained
-  syn match goFunctionCall      /\w\+\ze(/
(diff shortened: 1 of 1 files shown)


syntax/go.vim
```diff
@@ -302,7 +302,7 @@ if g:go_highlight_functions != 0
   syn match goPointerOperator   /\*/ nextgroup=goReceiverType contained skipwhite skipnl
   syn match goReceiverType      /\w\+/ contained
   syn match goFunction          /\w\+/ contained
-  syn match goFunctionCall      /\w\+\ze(/
+  syn match goFunctionCall      /\w\+\ze(/ contains=GoBuiltins,goDeclaration
```
fatih: Why do we include GoBuiltins? What does this change?

fatih: Ok I see now :)

fatih:
Thanks @svanharmelen 👍

PR #1072, opened 2016-10-04
Update license to credit Renee French for Gopher

hitjim:
Update license to credit Renee French for original Go Gopher design, since the awesome new Vim-Gopher image has recently been added to the repo.

I placed in the license file since other prominent Go sites, even golang.org don't directly credit Renee on the main page.


fatih:
Thanks @hitjim. I've totally forget to add a license for this.

PR #1074, opened 2016-10-05
Add table-driven tests and subtests snippets

namsral:
These table-driven tests snippets where created from the examples in the "Using Subtests and Sub-benchmarks" post on [blog.golang.org](https://blog.golang.org/subtests)

### DIFF
--- a/gosnippets/UltiSnips/go.snippets
+++ b/gosnippets/UltiSnips/go.snippets
@@ -2,6 +2,40 @@
 
 priority -10
 
+snippet tt "table-driven tests"
+func Test${1:Func}(t *testing.T) {
+	testCases := []struct {
+		${2:input} ${3:Type}
+		want ${4:Type} 
+	}{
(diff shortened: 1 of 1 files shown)


fatih:
Hi @namsral 

There are many placholders, seems like a fairly complex snippet. Can you show a gif on how you use it? Maybe we should change some parts to static text.

namsral:
Will you settle for a .mov?

http://www.namsral.com/assets/table-driven-tests.mov

fatih:
@namsral the link above doesn't work

namsral:
@fatih fixed the link.

fatih:
I think this is a pretty complex snippet @namsral. Thanks for your contribution but I think you should use it privately. Both snippet engines allows you to define custom snippets on your own. However I don't want to include this to vim-go as I think it's too big and complex to navigate. Sorry for the late response.

PR #1084, opened 2016-10-16
Provide completions for unimported packages.

zmb3:
gocode recently added support for unimported packages.
https://github.com/nsf/gocode/commit/57b3fab1de11102eee896051605108949fbe75da

This comes in handy for example when you know approximately what
function you're looking for but don't remember the exact
name or signature.  It's must faster to get suggestions for
packages without having to import them first.

The feature is optional and must be enabled with `gocode set unimported-packages true`.
Enable the gocode feature by default, but allow it to be configured with
g:go_gocode_unimported_packages.


fatih:
Thanks @zmb3 👍

PR #1089, opened 2016-10-25
lint: add (go-lint) plug mapping

### LINKED ISSUE #1068: golint and gometalinter appear to be missing
I added a leader key for golint in .vimrc and it does not work. Key:

``` vim
au FileType go nmap <leader>l <Plug>(golint)
```
### Expected behavior

golint would run but does not.
gometalint does not work either. Are golint and gometalinter defined? :h vim-go says they should work.
### Steps to reproduce:

Add above leader key.
### Configuration

These work

```
:GoLint
:GoMetaLinter
```

tdilo:
Add plug mapping for golint

Fixes #1068

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -786,6 +786,10 @@ annotation.
 
 Calls `goimports` for the current package
 
+                                                                   *(go-lint)*
+
+Calls `golint` for the current package
+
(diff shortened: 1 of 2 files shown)


fatih:
Thanks @tdilo

PR #1099, opened 2016-11-15
fix: restore syntax state after gofmt

powerman:
When running gofmt on saving file syntax highlight is turned off.
I'm not sure why it happens at first place, so maybe I'm fixing consequences instead of main reason.
I didn't bisect this, but it doesn't happens at 4f67ae2, so this bug was introduced in last months.

I'm using vim-7.4.2102.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -161,6 +161,9 @@ function! go#fmt#Format(withGoimport)
     silent edit!
     let &fileformat = old_fileformat
     let &syntax = &syntax
+    if exists("g:syntax_on")
+      syntax enable
+    endif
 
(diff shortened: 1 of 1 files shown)


fatih:
Hi @powerman. 

Even though this fixes your problem, this does not fixes the underlying problem. `fmt.vim` doesn't touch and change any syntax highlighting. I would prefer to disable all plugins and test again.

powerman:
Yeah, you was right. I've traced down this issue to something strange related to augroup cleanup. Probably it's related to vim internal change in last versions. If you're curious the fix was: https://github.com/powerman/vim-plugin-autosess/commit/429e0794e3ef1affd2d789404c76f34662101cf2

PR #1100, opened 2016-11-16
Add abort to every autoload function

Alok:
Used `vint` to go over the codebase and applied the (easy to do)
suggestion of adding `abort` to every autoload function.


fatih:
Thanks @Alok, great addition 👍

PR #1101, opened 2016-11-19
add an option to show type info when completion is done

haya14busa:
Echoing `v:completion.info` after completion is done is annoying if users uses completion other than vim-go omni completion.

When I run completion with completefunc and this completion-item contains multi-line info, vim-go automatically echo this info and it messes up the window.
So, I added this option.


fatih:
This seems logical. Thanks @haya14busa for the fix 👍

haya14busa:
Thank  you for including it :bird: :+1:

PR #1113, opened 2016-11-22
go.vim func n method call style

liangzuobin:
make function deceleration and its call highlights in the same style.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -307,19 +307,19 @@ else
   syn keyword goDeclaration func
 endif
 hi def link     goFunction          Function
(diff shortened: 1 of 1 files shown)


fatih:
@liangzuobin Thanks but we shouldn't change default values as it's mess up people syntax highlighting they are used to. Can you please provide a screenshot that shows the new difference? Thanks

liangzuobin:
Hi, thx for your response. 
this is what I want
Before:
screen shot 2016-12-01 at 13 44 02

After:
screen shot 2016-12-01 at 13 42 18

PR #1121, opened 2016-11-24
Always use full path as statusline target directory

pborzenkov:
Some parts of the code use go package dir relative to the current GOPATH
as a target for statusline update message which leads to their messages
not being displayed in the statusline (because it expects full path to
the package). Fix this by always using full package path.


fatih:
Thanks @pborzenkov. I've fixed it for other commands, seems like I've forgot them. Thanks for fixing it.

Issue #1019, opened 2016-08-21
:GoTest errors appear in location list


myitcv:
### Actual behavior

`:GoTest` uses the location list for compile errors
### Expected behavior

`:GoTest` should use quickfix for compile errors

Per [the docs](https://github.com/fatih/vim-go/blob/d9c531fca612beb4d4edaa3b4e4efb8c4f5b2cbe/doc/vim-go.txt#L346-L348)
### Steps to reproduce:

```
cd `mktemp -d`
cat <<EOD > main.go
package main

func // this will not compile
EOD
go test
```

You should see the output:

```
./main.go:4: syntax error: unexpected EOF, expecting name or (
```

Open `main.go` in Neovim, the `:GoTest`

Observe that the location list is used
### Configuration
- `vimrc` you used to reproduce: basic `.config/nvim/init.vim` that loads `vim-go`
- vim version: NVIM v0.1.5-752-gfb98145
- vim-go version: a4ba0a5b1b619bfd0b41618620eed1fe6a22a0a2
- go version: go version go1.7 linux/amd64

fatih:
Hi @myitcv 

Seems like for normal vim we still use `quickfix` but it's [hardcoded](https://github.com/fatih/vim-go/blob/master/autoload/go/jobcontrol.vim#L152) for Neovim because we're using jobcontrol. 

To fix it I have to refactor the `Spawn` functions to accept an option argument so I can modify it. For now you can use the following to override it globally until it's fixed for the Neovim case:

```
let g:go_list_type = "quickfix"
```

myitcv:
Thanks for the suggestion of a workaround for now @fatih

Issue #1109, opened 2016-11-21
:GoInstall: doesn't close the quickfix window when successful in async mode 


vdemario:
### Actual behavior

`:GoInstall:` runs and reports errors in a quickfix window. I fix those errors and run it again. It finishes successfully but the quickfix window with the previous errors remains open.

This is the same as #526, but it's back now in async mode.

### Expected behavior

`:GoInstall:` runs and reports errors in a quickfix window. I fix those errors and run it again. The quickfix window closes.

### Steps to reproduce

1. Create a Go file with a compile error.
2. Call `:GoInstall:`.
3. Fix the error.
4. Call `:GoInstall:` again.

### Configuration

* vim version: version 8.0.94
* vim-go version: master@59c56f0903a1ed1136856acefee2d014574d28f0
* go version: go version go1.7.3 darwin/amd64

fatih:
Hi @vdemario 

I've just tested it and as you see below in the demo it works without any problem. Can you share a screencast at least? I can't reproduce it otherwise.

![vim-go-install](https://cloud.githubusercontent.com/assets/438920/20511002/39fd319a-b07d-11e6-9f25-4c88081eb31e.gif)

vdemario:
How did you record the screencast, @fatih ? I can do the same here, sure, but I've never done it.

You mentioned updating to at least 8.0.91 on #1097, but I'm on 8.0.94. It's installed via homebrew, I don't know if it matters.

vdemario:
Nevermind, got it, here's the recording.

![goinstall](https://cloud.githubusercontent.com/assets/6087424/20526468/5a68e45a-b0ab-11e6-8965-60a21b53d451.gif)

fatih:
Seems the problem is somewhere else. To be honest I don't have any clue why it fails for you guys. Can you past the `build` options of `Vim` here? Maybe there is a feature built in that prevents closing it.

vdemario:
Entire `vim --version`:

<details>
vitor@macvitor ~> vim --version
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Nov 21 2016 11:43:38)
MacOS X (unix) version
Included patches: 1-94
Compiled by Homebrew
Huge version without GUI.  Features included (+) or not (-):
+acl             +file_in_path    +mouse_sgr       +tag_old_static
+arabic          +find_in_path    -mouse_sysmouse  -tag_any_white
+autocmd         +float           +mouse_urxvt     -tcl
[...]

PR #1123, opened 2016-11-24
job: close quickfix window if a job has succeeded

pborzenkov:
Otherwise all previous build errors are shown even after successful
build.

### DIFF
--- a/autoload/go/job.vim
+++ b/autoload/go/job.vim
@@ -57,16 +57,17 @@ function go#job#Spawn(args)
       endif
     endif
 
+    let l:listtype = go#list#Type("quickfix")
     if exitval == 0
-      call go#list#Clean(0)
-      call go#list#Window(0)
+      call go#list#Clean(l:listtype)
+      call go#list#Window(l:listtype)
       return
     endif
 
-    call self.show_errors()
+    call self.show_errors(l:listtype)
   endfunction
 
-  function cbs.show_errors() dict
+  function cbs.show_errors(listtype) dict
     let cd = exists('*haslocaldir') && haslocaldir() ? 'lcd ' : 'cd '
     try
       execute cd self.jobdir
@@ -84,11 +85,10 @@ function go#job#Spawn(args)
     endif
 
     if self.winnr == winnr()
-      let l:listtype = "quickfix"
-      call go#list#Populate(l:listtype, errors)
-      call go#list#Window(l:listtype, len(errors))
+      call go#list#Populate(a:listtype, errors)
+      call go#list#Window(a:listtype, len(errors))
       if !empty(errors) && !self.bang
-        call go#list#JumpToFirst(l:listtype)
+        call go#list#JumpToFirst(a:listtype)
       endif
     endif
   endfunction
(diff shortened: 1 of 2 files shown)


autoload/go/job.vim
```diff
@@ -57,16 +57,17 @@ function go#job#Spawn(args)
       endif
     endif
 
+    let l:listtype = "quickfix"
```
fatih: I think we should use here:  `let l:listtype = go#list#Type("quickfix")`. the `Type()` function returns the user preference (can be set via `go_list_type`) , if there is none, then the argument itself is returned.

fatih:
Btw, the current code is all set to `quickfix`, not sure how this fixes it. What exactly is the problem? Is this fixing #1109, if yes what's the real problem here?

pborzenkov:
Yes, I am experiencing the very same problem as described in #1109. And it does indeed fix it for me.

Initially, I though that go#list#Window(0) would close 'localtionlist', but now I see that it should close 'quickfix' by default. So I'm not sure how this patch fixes it, anymore :) I'm going to dig a bit deeper here.

pborzenkov:
Ok, so I added some debug messages to list.vim:

```
--- a/autoload/go/list.vim
+++ b/autoload/go/list.vim
@@ -7,15 +7,20 @@ endif
 " closes the window
 function! go#list#Window(listtype, ...) abort
   let l:listtype = go#list#Type(a:listtype)
+  echom a:listtype
+  echom l:listtype
   " we don't use lwindow to close the location list as we need also the
   " ability to resize the window. So, we are going to use lopen and lclose
   " for a better user experience. If the number of errors in a current
   " location list increases/decreases, cwindow will not resize when a new
   " updated height is passed. lopen in the other hand resizes the screen.
   if !a:0 || a:1 == 0
+    echom "first if true"
     if l:listtype == "locationlist"
+      echom "closing location list"
       lclose
     else
+      echom "closing quickfix window"
       cclose
     endif
     return
```

And here is the output of successful build (e.g. the one that should close currently open quickfix window):

```
vim-go: building dispatched ...                                                                                                                               
vim-go: SUCCESS                                                                                                                                               
0                                                                                                                                                             
0                                                                                                                                                             
first if true                                                                                                                                                 
closing location list
```

For some reasons, even though l:listtype == 0, this condition ' if l:listtype == "locationlist"' is true.

pborzenkov:
Ok, looks like I understand what's going on here. vim casts "locationlist" to an int, since this is the type of the first comparison argument, and it becomes zero. That why the condition is true.

pborzenkov:
@fatih Ok, so I've updated the patch to use go#list#Type and also fixed the same problem (integer arguments to go#list#Window()) in jobcontrol.vim. I don't have neovim installed, so jobcontrol.vim change is untested.

autoload/go/jobcontrol.vim
```diff
@@ -99,9 +99,10 @@ function! s:on_exit(job_id, exit_status) abort
 
   call s:callback_handlers_on_exit(s:jobs[a:job_id], a:exit_status, std_combined)
 
+  let l:listtype = go#list#Type("locationlist")
```
fatih: Can you change this to `quickfix` as well? I know it was already like this before, but let us be consistent with the others as well. It also fixes https://github.com/fatih/vim-go/issues/1019 as well :)

fatih:
@pborzenkov this looks good. Just a minor comment, once that is in, it's ready to be merged.

pborzenkov:
@fatih Done.

fatih:
Thanks @pborzenkov 👍

Issue #1053, opened 2016-09-17
Option for autocreate to treat empty directories as a package named after that directory


jimmyfrasche:
I love opening a new file and having the package name filled in. 

I don't have much use for the default template, though. I set it to an empty file. 

It would be nice if there were an option so that opening a new go file in an empty directory caused the buffer to contain "package d", where d is the name of the directory.

I'd have to change it if I were writing a command, but more often than not it would be the correct action, at least in my case.

fatih:
Hi @jimmyfrasche 

If you feel adventerous and want to learn some vimscript (and also contribute to vim-go), the code is here: https://github.com/fatih/vim-go/blob/master/autoload/go/template.vim#L19

You can easily add a new setting, such as `go_template_no_file` which is added to line `14`. So we would check first if `go_template_no_file` is set to `1` or if `l:package_name` is `-1`. Because `go_template_no_file` would come first, the else branch will be executed, which in return does what you want, basically creating a file with the package name.

jimmyfrasche:
Thanks for taking the time to lay that out, @fatih.

I've managed to cobble something together after much googling. I'll do some tests locally and try to send a PR tomorrow

PR #1124, opened 2016-11-24
Add option to use cwd as package name instead of template

jimmyfrasche:
This adds an option to use the directory name that contains a package as the name of the package instead of using a template, as requested in issue #1053

### DIFF
--- a/autoload/go/template.vim
+++ b/autoload/go/template.vim
@@ -1,7 +1,7 @@
 let s:current_file = expand("<sfile>")
 
 function! go#template#create() abort
-  let l:go_template_no_file = get(g:, 'go_template_no_file', 0)
+  let l:go_template_use_pkg = get(g:, 'go_template_use_pkg', 0)
   let l:root_dir = fnamemodify(s:current_file, ':h:h:h')
 
   let cd = exists('*haslocaldir') && haslocaldir() ? 'lcd ' : 'cd '
@@ -13,12 +13,12 @@ function! go#template#create() abort
   " if we can't figure out any package name(no Go files or non Go package
   " files) from the directory create the template or use the cwd
   " as the name
-  if l:package_name == -1 && l:go_template_no_file != 1
+  if l:package_name == -1 && l:go_template_use_pkg != 1
     let l:template_file = get(g:, 'go_template_file', "hello_world.go")
     let l:template_path = go#util#Join(l:root_dir, "templates", l:template_file)
     exe '0r ' . fnameescape(l:template_path)
     $delete _
-  elseif l:package_name == -1 && l:go_template_no_file == 1
+  elseif l:package_name == -1 && l:go_template_use_pkg == 1
     " cwd is now the dir of the package
     let l:path = fnamemodify(getcwd(), ':t')
     let l:content = printf("package %s", l:path)
--- a/autoload/go/template.vim
+++ b/autoload/go/template.vim
@@ -1,6 +1,7 @@
 let s:current_file = expand("<sfile>")
 
 function! go#template#create() abort
+  let l:go_template_no_file = get(g:, 'go_template_no_file', 0)
   let l:root_dir = fnamemodify(s:current_file, ':h:h:h')
 
   let cd = exists('*haslocaldir') && haslocaldir() ? 'lcd ' : 'cd '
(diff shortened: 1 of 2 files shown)


autoload/go/template.vim
```diff
@@ -10,12 +11,19 @@ function! go#template#create() abort
...
+  " files) from the directory create the template or use the cwd
+  " as the name
+  if l:package_name == -1 && l:go_template_no_file != 1
     let l:template_file = get(g:, 'go_template_file', "hello_world.go")
     let l:template_path = go#util#Join(l:root_dir, "templates", l:template_file)
     exe '0r ' . fnameescape(l:template_path)
     $delete _
+  elseif l:package_name == -1 && l:go_template_no_file == 1
+    " cwd is now the dir of the package
+    let l:path = fnamemodify(getcwd(), ':t')
```
fatih: We already have the package name, via `l:package_name` can you use that instead of using the directory? Because directory and package names can be different in Go.

autoload/go/template.vim
```diff
@@ -10,12 +11,19 @@ function! go#template#create() abort
...
+    let l:content = printf("package %s", l:path)
+    call append(0, l:content)
+    $delete _
```
fatih: Per my comment above, I think you don't need three separate clauses. The following is sufficient:

```viml
  if l:go_template_no_file != 1 || l:package_name == -1
    let l:template_file = get(g:, 'go_template_file', "hello_world.go")
    let l:template_path = go#util#Join(l:root_dir, "templates", l:template_file)
    exe '0r ' . fnameescape(l:template_path)
    $delete _
  else
    let l:content = printf("package %s", l:package_name)
    call append(0, l:content)
    $delete _
  endif
```

doc/vim-go.txt
```diff
@@ -1454,6 +1458,13 @@ the `hello_world.go` file is used.
 >
   let g:go_template_file = "hello_world.go"
 <
+                                                     *'g:go_template_no_file'*
+
+Specifies that, rather than using a template, the directory name is used as the
+package name if a new Go file is created. Checkout |'g:go_template_autocreate'| for more info. By default it is disabled.
```
fatih: Should be `.. the package name is used if a new Go file is created`

Also can you wrap this just like the other paragraphs.

doc/vim-go.txt
```diff
@@ -1438,6 +1438,10 @@ with a Go code template. By default the template under
 `templates/hello_world.go` is used. This can be changed with the
 |'g:go_template_file'| setting.
 
+To use the package's directory as the name of the package instead of a template
```
fatih: Should  be `To use the package name instead of ...`

autoload/go/template.vim
jimmyfrasche: `g:go_no_template_use_cwd` or something like that?

autoload/go/template.vim
fatih: Please neglect this comment.

doc/vim-go.txt
fatih: This still applies.

doc/vim-go.txt
fatih: This still applies.

autoload/go/template.vim
jimmyfrasche: I see what you mean, but I think it would be confusing to a reader (of the docs or code) if the word directory isn't mentioned anywhere. 

This change, essentially, guesses that you most likely want to use the name of the directory that a new package is created in as the name of that package.

Using package instead of directory in the docs means that the docs would have to say something like, "use the package name for the name for new packages".

That's confusing because package means two different things in the same sentence.

But something like, "use the directory name of the package as the name for new packages," makes it clear what the two different things are and how they relate.

This would be the first time the directory is mentioned in the docs, but it would also be the first time that it was relevant.

fatih: > Using package instead of directory in the docs means that the docs would have to say something like, "use the package name for the name for new packages".

I disagree, it's up to us how to write it down. I went over again and here are the fixed doc sentences:


```
When a new Go file is created, vim-go automatically fills the buffer content
with a Go code template. By default the template under
`templates/hello_world.go` is used. This can be changed with the
|'g:go_template_file'| setting.

If the new file is created in an already prepopulated package (with other Go
files), in this case a Go code template with only the Go package declaration
(which is automatically determined according to the current package) is added.

To use always the package name instead of the template (`tempaltes/hello_world.go`)
when new Go file is created, enable the |'g:go_template_use_pkg'| setting.

By default it is enabled.
>
  let g:go_template_autocreate = 1
<
```

and 

```
Specifies that, rather than using a template, the package name is used 
if a new Go file is created. Checkout |'g:go_template_autocreate'| for more 
info. By default the `hello_world.go` file is used.

>
  let g:go_template_use_pkg = 0
<
```

As you see, it's not complicated and it makes sense to use it this way. It's understandable and there is nowhere the notion of directory. If you just update the docs with the ones above and change the variable name to `go_template_use_pkg` we're good to go.

fatih:
Thanks @jimmyfrasche  👍

PR #1128, opened 2016-11-25
coverage: no need to change turn on/off syntax

### LINKED ISSUE #1034: GoCoverageClear removes custom highlights
### Actual behavior

I have some custom highlights in my `.vimrc`:

``` vim
highlight ColorColumn ctermbg=235
highlight NonText ctermfg=239
highlight SpecialKey ctermfg=239
highlight goSameId term=bold cterm=bold ctermbg=250 ctermfg=239
```

When I run GoCoverageClear it resets them to the default values.
### Expected behavior

I would think it should not set the highlights to their default colors. Perhaps there is a way I can override this in my `.vimrc`?
### Steps to reproduce
1. Set a custom highlight in your `.vimrc` like above.
2. Run `:GoCoverageClear` or `:GoCoverageToggle`
### Configuration
- vim version: 7.4
- vim-go version: current master (c63c798f71790c7952a9adafa866be1523f38589)
- go version: 1.7

fatih:
Just removing the matches is enough for us.

Fixes #1034

### DIFF
--- a/autoload/go/coverage.vim
+++ b/autoload/go/coverage.vim
@@ -85,10 +85,7 @@ endfunction
 
 " Clear clears and resets the buffer annotation matches
 function! go#coverage#Clear() abort
-  " only reset the syntax if the user has syntax enabled
-  if !empty(&syntax)
-    if exists("g:syntax_on") | syntax enable | endif
-  endif
+  call clearmatches()
 
   if exists("s:toggle") | let s:toggle = 0 | endif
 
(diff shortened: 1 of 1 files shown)


jsageryd:
Thank you for this! Finally the background doesn't get all grey when I turn of the coverage report. 🎉

jsageryd:
Thank you for this! Finally the background doesn't get all grey when I turn of the coverage report. 🎉

PR #1129, opened 2016-11-25
Add path conversion for MSYS2 environments

tdilo:
In a MSYS2-based environment on Windows (this includes the shell
provided by Git for Windows) Vim operates as if running on a Unix
system. The external tools used by vim-go on the other hand expect path
names in Windows path notation as command line arguments. When called
from a MSYS2 environment, they were supplied with a path in Unix notation
instead (e.g. /c/gopath/src/foo.go). To resolve this discrepancy, all
command line arguments passed to external tools must be translated into
Windows path notation in such an environment. Similarly, output created
by external tools must be converted back to Unix notation when it
contains path names.

This commit introduces functions for translating paths between internal
and external (native to the tool) notation. Path translation is
performed with the help of the cygpath utility provided by the Cygwin
environment on which MSYS2 is based. As Windows path names contain
backslashes, further escaping is often required when using the
conversion functions introduced here.

The path conversion functions are then used in the vim-go commands where
command line arguments for external tools are constructed or tool output
and environment variables are parsed. The translation is a no-op on
other platforms and when using a native Vim build on Windows.

Thanks to @fawick for testing and rebasing to the Vim 8.0 changes!

### DIFF
--- a/autoload/go/coverage.vim
+++ b/autoload/go/coverage.vim
@@ -17,7 +17,7 @@ function! go#coverage#BufferToggle(bang, ...) abort
 endfunction
 
 " Buffer creates a new cover profile with 'go test -coverprofile' and changes
-" teh current buffers highlighting to show covered and uncovered sections of
+" the current buffers highlighting to show covered and uncovered sections of
 " the code. Calling it again reruns the tests and shows the last updated
 " coverage.
 function! go#coverage#Buffer(bang, ...) abort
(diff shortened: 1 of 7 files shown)


fatih:
Hi @tdilo 

First of all thanks for your contribution. This is a fairly complex change and usually I don't want to rely on third party tools for specific internal changes (tools for manipulating Go is welcome though, such as goimports ,etc..). If we merge this, I'll be the sole person to maintain and I don't have any clue about cygpath and co. 

I also think that `ToExternal` and `ToInternal` functions tremendously complicates the code base. Going further, if someone reads the source code, they have to understand it before they can contribute anything. I wonder if there is a more simpler way. How are you working right now ? I'm open to fix anything that Vim supports, but if Vim doesn't support it I think  vim-go should not try to support other platforms. I have limited time space I can spend in the future. I hope you understand me and let me know if there is anything I can do.

PR #1132, opened 2016-11-27
handle :GoDef from modified buffer

### LINKED ISSUE #1125: :GoDef in modified buffer fails
This is a regression of changes introduced in #944 to avoid errors when trying to `:GoDef` in a modified buffer. The regression was introduced in #1083 (apparently deliberately?).

### Actual behavior

vim-go displays an error,
>  Error detected while processing function go#def#Jump[86]

### Expected behavior

A buffer should be opened and the cursor placed at the position of the definition of the identifier that was under the cursor when `:GoDef` was run.

### Steps to reproduce:

1. Open a file and modify it without saving
2. Place the cursor on an identifier that is declared in another file
3. run `:GoDef`

### Configuration
* vim-go version: 1.10

bhcleek:
Handle jumping to type definitions from a modified buffer. When the
definition is in the same file, skip trying to open the file that
contains the definition - just set the cursor position. When the
definition is in a different file and the current buffer has unsaved
modifications, use :hide edit instead of :edit.

Fixes #1125

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -146,19 +146,27 @@ function! s:jump_to_declaration(out, mode, bin_name) abort
     if get(g:, 'go_def_reuse_buffer', 0) && bufloaded(filename) != 0 && bufwinnr(filename) != -1
       " jumpt to existing buffer if it exists
       execute bufwinnr(filename) . 'wincmd w'
-    elseif a:mode == "tab"
-      let &switchbuf = "usetab"
-      if bufloaded(filename) == 0
-        tab split
+    else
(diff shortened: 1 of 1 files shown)


bhcleek:
I inadvertently made these changes on an older version of master than I should have. I'm working to fix them now, and will push a rebased commit once I have conflicts resolved.

bhcleek:
I've rebased on top of current master.

fatih:
Thanks @bhcleek. Just had time to look after this. Confirmed that it was due `hidden` being set. This fixes it 👍

PR #1133, opened 2016-11-27
allow popping from jump list when buffer is modified

bhcleek:
Properly pop to the requested location in the jump list, even when the
current buffer is modified. When current buffer is the target, don't try
to load the target file using edit or hide edit; just set the cursor
position. When the current buffer is modified and is not the target, use
hide edit to hide the current buffer.


fatih:
Thanks @bhcleek

Issue #1131, opened 2016-11-26
Too many arguments for function: setloclist


magisterquis:
When saving a go source file, Vim 7.4.1467 gives the following errors:

```Error detected while processing function <SNR>14_fmt_autosave[3]..go#fmt#Format[134]..go#list#Populate:```

```E118: Too many arguments for function: setloclist```

**Comment by fatih** [comment]
Hi @magisterquis 

Can you paste the documentation of `:help setloclist()` please? Seems like we added a feature that your Vim version doesn't support yet.

**Comment by wyattanderson** [comment]
Also seeing this error.

Output of `:help setloclist()`:

```
setloclist({nr}, {list} [, {action}])                   setloclist()
                Create or replace or add to the location list for window {nr}.
                When {nr} is zero the current window is used. For a location
                list window, the displayed location list is modified.  For an
                invalid window number {nr}, -1 is returned.
                Otherwise, same as setqflist().
                Also see location-list.
```

Version:
```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Jun 16 2016 10:50:38)
Included patches: 1-1689
Modified by pkg-vim-maintainers@lists.alioth.debian.org
Compiled by pkg-vim-maintainers@lists.alioth.debian.org
Huge version without GUI.  Features included (+) or not (-):
+acl             +dialog_con      +job             -mouse_sysmouse  +scrollbind      +visual
+arabic          +diff            +jumplist        +mouse_urxvt     +signs           +visualextra
+autocmd         +digraphs        +keymap          +mouse_xterm     +smartindent     +viminfo
-balloon_eval    -dnd             +langmap         +multi_byte      +startuptime     +vreplace
-browse          -ebcdic          +libcall         +multi_lang      +statusline      +wildignore
++builtin_terms  +emacs_tags      +linebreak       -mzscheme        -sun_workshop    +wildmenu
+byte_offset     +eval            +lispindent      +netbeans_intg   +syntax          +windows
+channel         +ex_extra        +listcmds        +packages        +tag_binary      +writebackup
+cindent         +extra_search    +localmap        +path_extra      +tag_old_static  -X11
-clientserver    +farsi           -lua             -perl            -tag_any_white   -xfontset
-clipboard       +file_in_path    +menu            +persistent_undo -tcl             -xim
+cmdline_compl   +find_in_path    +mksession       +postscript      +terminfo        -xsmp
[...]

PR #1135, opened 2016-11-27
list: set title only for recent Vim versions

fatih:
The last argument was introduced with 7.4.2200. Make sure it's only used
with that version.

https://github.com/vim/vim/commit/d823fa910cca43fec3c31c030ee908a14c272640

Fixes: #1131


peteretelej:
Tested, resolves issue #1131 on older vim

PR #1136, opened 2016-12-03
fix bug in guru vim script

jdhenke:
`:GoDescribe` and other `guru` functions failed with 

```
Error detected while processing function go#guru#Freevars..<SNR>33_run_guru..<SNR>33_sync_guru..<SNR>33_guru_cmd:
line   42:
E107: Missing parentheses: result.tags = tags
```

This change fixed this for me and now `:GoDescribe` works.


fatih:
Thanks @jdhenke  👍

PR #1137, opened 2016-12-04
Fold Function

mencattini:
The function mark the bracket of a go  ```func```. Then you can unfold with ```zo``` fold with ```zc``` or toogle folding with ```za```.


fatih:
Hi @mencattini 

Can you add this behind a setting please? Checkout the file to see more examples. And also please add the documentation to `docs/vim-go.txt` as well. Thanks.

PR #1141, opened 2016-12-07
Change default value of `g:go_asmfmt_autosave` to 0.

### LINKED ISSUE #1140: Change go_asmfmt_autosave default to 0
The default configuration in vim-go is to automatically format assembly files (suffix `.asm`) with `asmfmt`. It is not at all obvious that a plugin focused on Go would format or otherwise modify an assembly file.

I can see that this is a useful feature for those working on the language runtime etc., but the majority of people who write Go will most likely never do this.

My suggestion is therefore to change the `go_asmfmt_autosave` default to 0. It should be enabled by those working on assembly files where `asmfmt` is the chosen tool for formatting.

### Steps to reproduce:

1. Run vim with the vim-go plugin enabled.
2. Open a new assembly file (e.g. `test.asm`)
3. Type the following (contrived) example:
```asm
label:
mov $0, %eax;
```
4. Save the file
5. Notice that the buffer contents change to:
```asm
label:
    mov $0, %eax;
```
(4 spaces inserted by me, actual result is a tab character preceding the second line.)

### Configuration

* vim-go version: 08c881

oystedal:
Fixes #1140.

### DIFF
--- a/autoload/go/asmfmt.vim
+++ b/autoload/go/asmfmt.vim
@@ -11,7 +11,7 @@
 "
 " Options:
 "
-"   g:go_asmfmt_autosave [default=1]
+"   g:go_asmfmt_autosave [default=0]
 "
 "       Flag to automatically call :Fmt when file is saved.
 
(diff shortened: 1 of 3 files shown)


fatih:
Thanks @oystedal

PR #1143, opened 2016-12-09
Fix indentation broken when using a action with a minus sign like `{{-'

kechako:
Golang's template has supported a action with a minus sign like `{{-` since go1.6.

Using this syntax breaks the indentation.


fatih:
Thanks @kechako 👍

Issue #1142, opened 2016-12-09
go_metalinter_deadline setting not respected


horgh:
### Behavior

I have changed my `go_metalinter_deadline` to 15s. However, I found that metalinter still times out after 5 seconds.

I believe this is because gometalinter defaults to a timeout of 5s, and in lint.vim there is a comment that says "we add deadline only for sync mode". I wonder if in the past gometalinter had no default deadline? It seems like we need it regardless.

I verified that if I move the line that adds `--deadline` to the command above `if go#util#has_job() && has('lambda')`, that gometalinter uses my 15s deadline and gometalinter can complete.

It seems like quite a small change to adjust this, but I thought I should make an issue rather than a PR in case there is a reason I do not understand that it is this way. If you like, I can make a PR.

Thank you for your time. This project is hugely useful!

### Steps to reproduce:

1. Have a slow computer :)
2. Set `g:go_metalinter_deadline = "15s"` in vimrc
3. Open a Go file
4. Save it, see "vim-go: linting started..."
5. Wait 5 seconds
6. See WARNING: deadline exceeded [etc] (after 5 seconds)

I normally develop on a faster machine, but I've been doing a little on an older server as I am not currently at home.

### Configuration

Add here your current configuration and additional information that might be
useful, such as:

* `vimrc` you used to reproduce:

```
set nocompatible

execute pathogen#infect()

let g:go_fmt_command = "goimports"
let g:go_metalinter_autosave = 1
let g:go_metalinter_autosave_enabled = ['vet', 'golint', 'errcheck']
let g:go_metalinter_deadline = "15s"

set statusline+=%#goStatuslineColor#
set statusline+=%{go#statusline#Show()}
set statusline+=%*
```

* vim version: 8.0.95
* vim-go version: 1.10
* go version: 1.7.3

fatih:
Hi @horgh 

As you said it's a very small change indeed. The reason it's just for `sync` mode is that because async linting doesn't block anymore, so I thought it should not have any kind of deadline at all.

I'm ok if you open a PR. But please make it so the following logic applies

* If `async`, the deadline should be passed only if the has defined `g:go_metalinter_deadline` itself
* If `sync`, the deadline should be passed as the default `5s` value, but if the user passed `g:go_metalinter_deadline` it should be overridden.

[...]

PR #1146, opened 2016-12-18
#1142: go_metalinter_deadline applies in async mode now

horgh:
This change is for issue #1142.

Previous to this change, the option g:go_metalinter_deadline would only
be respected when calling gometalinter if we called it synchronously.

This change removes the default value of g:go_metalinter_deadline of 5
seconds.

For async mode: We pass in a --deadline option now only if we've defined
g:go_metalinter_deadline. For sync mode: We always pass in a --deadline
value. If g:go_metalinter_deadline is not set, we pass in 5 seconds.

Please note that gometalinter itself has a default deadline of 5 seconds
if no deadline is passed to it, so in either case the default of 5
seconds will apply.

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -10,10 +10,6 @@ if !exists("g:go_metalinter_enabled")
   let g:go_metalinter_enabled = ['vet', 'golint', 'errcheck']
 endif
 
-if !exists("g:go_metalinter_deadline")
-  let g:go_metalinter_deadline = "5s"
-endif
-
 if !exists("g:go_golint_bin")
   let g:go_golint_bin = "golint"
 endif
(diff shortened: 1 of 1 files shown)


horgh:
I should mention: I did look at the doc/vim-go.txt and it looks like the information about this option is still valid.

I tried to follow the approach you requested in the issue I opened. However as you can see, I found that gometalinter has a deadline of 5s even if no `--deadline` is specified. It may make sense to simply apply the same logic for async and sync mode (always pass in a deadline). Since in either case the default will still be 5 seconds, just in the async case we do not specify it at all.

I tested this:

In async mode:

  * If no deadline configured, none was passed in.
  * If a deadline is configured, it is passed in.

In sync mode:

  * If no deadline configured, 5s was passed in.
  * If a deadline is configured, it was passed in.

Please let me know if you see anything I should change.

Thanks again!

fatih:
Thanks @horgh for the contribution 👍

PR #1152, opened 2016-12-24
If GOPATH is empty check for the value of go env GOPATH

ismail:
Go 1.8 has a default GOPATH (~/go) , if there is no GOPATH set we should also check for the value of go env GOPATH before bailing out.


fatih:
Thanks @ismail 👍

PR #1153, opened 2016-12-25
check package existence with govendor if installed

genzj:
I use glide for package management however vim-go cannot import packages under `PROJ/vendor/` so I came out with the idea that using `govendor list` to detect package existence if possible.

It works for my case, but not sure if there are better ideas for other package management practice. Any suggestion will be greatly appreciated.

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -150,9 +150,34 @@ function! go#tool#ExecuteInDir(cmd) abort
   return out
 endfunction
 
+function! go#tool#HasGoVendor() abort
+    let command = "command -v govendor"
+    let out = go#tool#ExecuteInDir(command)
+    return go#util#ShellError()
+endfunction
+
+" Exists checks whether the given importpath exists in vendor path or not. 
+" GoVendor github.com/kardianos/govendor is called under the hood. It returns
+" 0 if the importpath exists under GOPATH, -1 if not exists, -2 if govendor
+" not exists in the PATH.
+function! go#tool#ExistsInVendor(importpath) abort
+    if go#tool#HasGoVendor() == 0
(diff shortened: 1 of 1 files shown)


fatih:
Hi @genzj 

If you have properly GOPATH, the newest `goimports` supports importing packages from `vendor/` folder.  Govendor, glide, etc.. do not change the package layout, they just help to vendor your packages. This PR introduces a wrong solution, please update your `goimports` and you're good to go. Thanks.

genzj:
Hi @fatih 

Thanks for your prompt response. I understand your idea and upgrading tools/checking GOPATH are also the first thing I was trying to do. However I found out `goimports` actually works well and the `:GoImports` can import missing packages from vendor folder too.

The problem is that `:GoImport` or `go#tool#Exists`, more specifically, isn't vendor-aware in my project folder structure which looks like:

```
~/go  <-- GOPATH
    - src
        - github.com/genzj   <-- my project
            - vendor
                - github.com/spf13/viper
                - github.com/Sirupsen/logrus
```

Changing `GOPATH` didn't make `go list` or `:GoImport github.com/spf13/viper` work, unless I **add a new `src` folder in `~/go/github.com/genzj/vendor` than move all original content into it**, then append `~/go/github.com/genzj/vendor` to the `GOPATH` (`vim-go` will automatically append it if I move `vendor/*` to `vendor/src/*`). But `vendor/src/.../` seems not the correct vendor folder layout according to golang doc (https://golang.org/cmd/go/#hdr-Vendor_Directories), neither the default layout for glide.

`govendor list` is a quick answer to vendor-aware package existence check so I added it. You mentioned I shouldn't change the package layout, did I make mistake when setting up work dir or gopath?

Thanks a lot.

PR #1154, opened 2016-12-28
Fix aysnchronous :GoReferrers when g:go_auto_sameids = 1

bhcleek:
Fix :GoReferrers so that it populates the location list window correctly
when g:go_auto_sameids = 1. Because the close_cb function was being
referenced using function() instead of funcref(), the callback was
referring to it by name instead of reference; if multiple guru jobs were
executing concurrently, the callback that was executed for all was the
function that was used a closure when the last job was started. Because
the callback function closes over a:args, the value of the last
a:args.custom_parse was used to output the results of all jobs.

Use guru's -modified flag when running asynchronously instead of writing
the current buffer out to a temp file.

Check guru_cmd's return value for a stdin_content key instead of
checking &modified of the current buffer when deciding whether to
pass result.stdin_content in both sync_guru and async_guru.

Remove an unused variable, stdin_content (NOT result.stdin_content!) in
guru_cmd.

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -34,7 +34,6 @@ function! s:guru_cmd(args) range abort
   let cmd = [bin_path]
 
   let filename = fnamemodify(expand("%"), ':p:gs?\\?/?')
-  let stdin_content = ""
   if &modified
     let sep = go#util#LineEnding()
     let content  = join(getline(1, '$'), sep )
@@ -127,7 +126,7 @@ function! s:sync_guru(args) abort
 
   " run, forrest run!!!
   let command = join(result.cmd, " ")
-  if &modified
+  if has_key(result, 'stdin_content')
     let out = go#util#System(command, result.stdin_content)
   else
     let out = go#util#System(command)
@@ -152,7 +151,7 @@ function! s:async_guru(args) abort
     return
   endif
 
-  let status_dir =  expand('%:p:h')
+  let status_dir = expand('%:p:h')
   let statusline_type = printf("%s", a:args.mode)
 
   if !has_key(a:args, 'disable_progress')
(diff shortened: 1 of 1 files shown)


autoload/go/guru.vim
```diff
@@ -198,23 +197,26 @@ function! s:async_guru(args) abort
...
+        \ 'err_mode': 'nl',
         \ }
 
-  if &modified
-    let l:tmpname = tempname()
-    call writefile(split(result.stdin_content, "\n"), l:tmpname, "b")
-    let l:start_options.in_io = "file"
-    let l:start_options.in_name = l:tmpname
-  endif
-
```
fatih: Thanks for the work. But I've didn't understand why you're removing this piece. Because that's the only way to pass the stdin content. Seems like you changed it below, which starts first the job and then sends the content of the stdin. Why do we need this? I've read your PR comment but couldn't understand it unfortunately.

bhcleek: For a few reasons: parity with how `guru_sync` works, symmetry with the command that `guru_cmd` constructs, and performance (it's faster to write to the current buffer's contents to guru's `stdin` than to write a temp file).

`guru_sync` correctly uses `guru -modified` when the current buffer is modified, but `guru_async` did not. I'm not sure the behavior of guru is even defined when `-modified` is provided, but nothing is written to its stdin.

fatih: I think we're using it it for `async_guru` as well. The command is the same for both `sync_guru` and `async_guru`. They both get it from `guru_cmd`. So if it's modified, it uses the `-modified` flag. If you also check `def.vim`, you'll see that we're also using it there as well for `guru`. I just want the `stdin` a part of the `start_options`. However right now you changed, we start the job first, and then we send some other content to it. I think that would cause a race condition. I'm ok with your other changes, but I'm not sure this one is needed.

bhcleek: I realize now that this code wasn't working quite the way I thought it was; it was, in fact, honoring the -modified flag as guru_cmd constructed it. I still think there's value in writing directly to the channel instead of writing to a temp file first and reading stdin from the file, though.

bhcleek: There won't be a race condition with guru using this method. The `-modified` flag tells guru to read archives from stdin, and it reads from stdin until EOF before doing anything else.

With that in mind, I can revert that portion of this PR if you still want me to, but I'd rather see it left in due to the performance advantage and elimination of the need to manage temp files (note that the temporary file wasn't being deleted either)...

bhcleek:
I'm closing this. I'm going to break out the fix that I was actually targeting here from the changes to how the guru job is managed and submit separate PRs for them. I will only submit the change to the async job management if I can empirically demonstrate a performance improvement.

PR #1155, opened 2016-12-28
use funcref() instead of function() for closures used as callbacks

bhcleek:
Use funcref() instead of function() to get a Funcref that will retrieve
the function by reference instead of by name. function() returns a
Funcref that resolves the function by name, which caused multiple
concurrent jobs of the same type to both execute the callback that
closed over the arguments of the last job to be started. funcref()
returns a Funcref that returns the function by reference (not by name),
so each job is assured that its callbacks execute the function that
closed over the arguments of that job.

Also refactored a couple of the guru functions to check the result of
guru_cmd for stdin_content explicitly instead of assuming that
guru_cmd's result has a stdin_content key when the buffer is modified.

The easiest way to see the problems that this PR addresses is to set `g:go_auto_sameids=1`, and then run `GoReferrers` on some reference. One would expect the location list to be opened, but it never did.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -419,7 +419,7 @@ function s:cmd_job(args) abort
     call go#statusline#Update(status_dir, status)
   endfunction
 
-  let a:args.error_info_cb = function('s:error_info_cb')
+  let a:args.error_info_cb = funcref('s:error_info_cb')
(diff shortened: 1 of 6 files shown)


fatih:
Thanks @bhcleek, nice fix 👍

PR #1156, opened 2016-12-30
clear buffer-local BufWinEnter properly

bhcleek:
Fix the check for a buffer-local BufWinEnter autocmd in
go#guru#ClearSameIds() so that multiple BufWinEnter autocmds are not
registered to rehighlight the sameids when re-entering the local buffer.


fatih:
Thanks @bhcleek

PR #1160, opened 2017-01-02
Use go_metalinter_command as List after split().

thsnr:
After splitting `go_metalinter_command` it is already a List. Wrapping the result inside another list causes a "using List as a String" error and linting to fail.


fatih:
Thanks for the fix @thsnr 👍

PR #1166, opened 2017-01-07
Split already creates a list

junkblocker:


junkblocker:
I see #1160 is already open on this.

PR #1145, opened 2016-12-17
fixes #1144: "E118: Too many arguments for function: <SNR>173_on_exit"

entombedvirus:
Changes in neovim job control API broke vim-go integration. This commit
updates alters the integration to follow recommended practices from the
neovim wiki: https://github.com/neovim/neovim/wiki/Following-HEAD#20161212

PR #1176, opened 2017-01-12
fix :GoBuild issues in neovim

joshuarubin:
1. honor `g:go_echo_command_info` when dispatching builds in neovim
1. fix `:GoBuild` error in neovim due to invalid jobcontrol handler function signatures (`s:on_stdout`, `s:on_stderr`)
1. update statusline before and after `go#jobcontrol#Spawn` command is executed

also, as an aside, it looks like vim's job capability is utilized much more than neovim's job-control. it would be nice to abstract the differences and bring neovim up to parity again.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -32,8 +32,11 @@ function! go#cmd#Build(bang, ...) abort
           \})
     return
   elseif has('nvim')
+    if get(g:, 'go_echo_command_info', 1)
+      call go#util#EchoProgress("building dispatched ...")
+    endif
+
     " if we have nvim, call it asynchronously and return early ;)
-    call go#util#EchoProgress("building dispatched ...")
     call go#jobcontrol#Spawn(a:bang, "build", args)
     return
   endif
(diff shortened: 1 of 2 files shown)


joshuarubin:
btw, (2) was caused by #1145
it removed the 3rd arg from the callback. here is the place that states that 3 args are _required_. my patch keeps the `dict` option though.
https://github.com/neovim/neovim/issues/5763#issuecomment-266722407

joshuarubin:
Any chance this can get merged?

andviro:
Haha, missed this one. Did a similar job in #1183

joshuarubin:
Any update on this?

fatih:
@joshuarubin thanks for the PR. Sorry for my late response. I'm busy lately and was sick as well. I also went abroad to give a talk and thus didn't had any time to look at my open source projects. I'll check it out the following days, if not next week. For now please use it locally. Thanks.

fatih:
Thanks @joshuarubin this looks good.

> also, as an aside, it looks like vim's job capability is utilized much more than neovim's job-control. it would be nice to abstract the differences and bring neovim up to parity again.

Agree, the reason is that I Vim is supported more than Neovim right now. I'm open to improvements though for both platforms, unless they require complex changes. Eventually going forward, even if it takes some time, they both will have some feature parity.

PR #1177, opened 2017-01-13
Correctly report the value of the 'g:go_guru_tags' variable

francescomari:
When reporting the value of the Guru tags, the `a:1` variable is used. The `g:go_guru_tags` variable should be used instead.


fatih:
Thanks @francescomari 👍

Issue #1172, opened 2017-01-07
neovim GoBuild errors


nhooyr:
### Behavior

I just ran `:GoBuild` in a package with errors and I get

```
vim-go: building dispatched ...
E118: Too many arguments for function: <SNR>121_on_stderr
vim-go: FAILED
Error detected while processing function <SNR>121_on_exit:
line   31:
E684: list index out of range: 0
Error detected while processing function <SNR>121_on_exit:
line   31:
E116: Invalid arguments for function go#util#EchoError
```

fatih:
Can you send a patch if you now what the problem is? I don't have time to look soon. Thanks

jasdel:
Ran into this issue my self also. I think something in a recent update of nvim is causing this issue. I've reverted vim-go back a few versions and reproduced similar errors while keeping nvim constant.

fatih:
Closing as I have limited information. @nhooyr please send a PR next time :)

nhooyr:
I haven't been programming in go for a while but it looks like someone created a fix?

https://github.com/andviro/vim-go/commit/537a0491fb5b2e005768bfc4dbfc76d2120f792e

PR #1183, opened 2017-01-24
fix for #1172

andviro:
Hi!
Newer versions of Neovim expect the `on_stdout`/`on_stderr` handler to have the same function signature as the `on_exit` handler. This caused empty `std_combined` list and the "invalid argument" error. I've added an extra guard condition on it, just in case.


fatih:
This is now merged with #1176. Thanks @andviro anyways 👍

andviro:
All thanks to you, for the excellent plugin. Keep up the great work!

PR #1184, opened 2017-01-25
Support 'encoding != utf-8'

mattn:
All of files should be written in utf-8 if the file will be passed to external command.


fatih:
Thanks @mattn 👍

PR #1185, opened 2017-01-25
log.Fatal(err) is useful

mattn:
In the most cases I write `if err != nil {}` is this.


fatih:
Agree I also use this a lot nowadays. Thanks @mattn

PR #1193, opened 2017-02-02
Set GOPATH for both possible go guru execution paths

jblebrun:


autoload/go/guru.vim
```diff
@@ -217,11 +213,19 @@ endfunc
 
 " run_guru runs the given guru argument
 function! s:run_guru(args) abort
+
```
fatih: Can you remove this newline please

autoload/go/guru.vim
```diff
@@ -217,11 +213,19 @@ endfunc
...
+  let old_gopath = $GOPATH
+  let $GOPATH = go#path#Detect()
   if go#util#has_job()
-    return s:async_guru(a:args)
+    let abort = s:async_guru(a:args)
```
fatih: This has several meanings. For async it's the job id however for sync_guru it's the output. So the caller defines what to look. Can you rename it to `res` as for `result`? I think that would make more sense.

autoload/go/guru.vim
```diff
@@ -217,11 +213,19 @@ endfunc
...
   endif
 
-  return s:sync_guru(a:args)
+  let abort = s:sync_guru(a:args)
+
+  let $GOPATH = old_gopath
+
+  return abort
+
```
fatih: Can you remove this newline as well please?

autoload/go/guru.vim
jblebrun: I have no idea why I called it `abort`. I think I was distracted by the `abort` in the function signature. 

On closer examination... does the result need to be returned at all? No callers seem to use it. Can we just call an return nothing?

fatih: I was returning with the previous one, as we might need it in the future and it's handy. You can also just return however I'm not sure how to unset it easily (as Vim doesn't have defer like we have in Go)

jblebrun: Ok, I've left it and changed the name to `res` as you suggested.

autoload/go/guru.vim
```diff
@@ -217,11 +213,17 @@ endfunc
...
 function! s:run_guru(args) abort
+  let old_gopath = $GOPATH
+  let $GOPATH = go#path#Detect()
   if go#util#has_job()
-    return s:async_guru(a:args)
+    let res = s:async_guru(a:args)
   endif
 
-  return s:sync_guru(a:args)
+  let res = s:sync_guru(a:args)
```
bhcleek: This line should now be in an else block so that guru isn't called twice when `async_guru` was called.

fatih:
Thanks @jblebrun

PR #1194, opened 2017-02-02
Ensure no trailing : in gopath

jblebrun:
This can happen if, for example, you don't currently have a $GOPATH set.


fatih:
Thanks @jblebrun

PR #1195, opened 2017-02-03
Modify the resize behavior for godoc

haleyrc:
The default behavior for resizing the split when using godoc was to limit the
vertical size, even when using the vertical split. This is now constrained to
only limit vertical size for horizontal splits and vice-versa.


fatih:
Thanks @haleyrc

PR #1197, opened 2017-02-04
Document 'noshowmode' requirement for echo_go_info

jhillyerd:
After playing around with vim 8 and a minimal vimrc, I found that noshowmode was the only way to see post-completion usage information.


fatih:
Thanks @jhillyerd

PR #1198, opened 2017-02-05
Introduce GoDebug

jodosha:
Hi, `vim-go` it's an essential plugin for me, thank you!

---

This is a proposal to introduce `GoDebug` command. It runs a debugging session in (Neo)Vim.
The debugger engine is [delve](https://github.com/derekparker/delve), which needs to be installed on the developer's machine.

I haven't added it to the list of the `GoInstallBinaries` because on OSX requires Homebrew for automatic installation. In case of manual installation things get more [complicated](https://github.com/derekparker/delve/blob/master/Documentation/installation/osx/install.md#manual-install).

---

Another addition of this PR is the `GoToggleBreakpoint` command, to toggle a breakpoint from a buffer. It adds a visual sign in the gutter column.

---

**Please note that this PR works only for NeoVim**. I didn't manage to get it to work for Vim too. Please advise. 😄 

I'm not sure if you want to add this debugging feature. If so, let me know, so I can fix the missing parts.

---

See it in action:

vim-go-debug

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -385,6 +385,61 @@ function! go#cmd#Generate(bang, ...) abort
   let $GOPATH = old_gopath
 endfunction
 
+function! go#cmd#ToggleBreakpoint(file, line, ...) abort
+  " Compose the breakpoint for delve:
+  " Example: break /home/user/path/to/go/file.go:23
+  let breakpoint = "break " . a:file. ':' . a:line
+
+  " Define the sign for the gutter
+  exe "sign define gobreakpoint text=◉ texthl=Search"
+
+  " If the line isn't already in the list, add it.
+  " Otherwise remove it from the list.
+  let i = index(g:go_breakpoints, breakpoint)
+  if i == -1
+    call add(g:go_breakpoints, breakpoint)
+    exe "sign place ". a:line ." line=" . a:line . " name=gobreakpoint file=" . a:file
+  else
+    call remove(g:go_breakpoints, i)
+    exe "sign unplace ". a:line ." file=" . a:file
+  endif
+endfunction
+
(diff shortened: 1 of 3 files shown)


fatih:
Hi @jodosha 

Thanks for the PR. Much appreciated. Unfortunately this is not a direct integration into vim and relies on Neovim's term feature. I think we should add a general solution, but both Vim and Neovim work differently in terms of jobs API. I'm not sure how to make it abstract enough. And usually Vim is first class citizen in vim-go. 

Additionally I'm explicitly asked to open a Github issue (https://github.com/fatih/vim-go/blob/master/.github/CONTRIBUTING.md) to discuss features like this before opening a PR. Because I don't want everyone to spend on something so that later they learn that I'm not going to merge it (due the concerns outlined above). 

 We should not use delve's client instead Vim should be the client itself. I'm looking forward for a more integrated, Vim only, where Vim is a client to the delve server. Thanks again for your contribution but this is not something that I have in my mind. I'm closing this, I hope you understand my point.

jodosha:
Hey @fatih thanks for getting back so quickly! No worries.

---

A one note about Vim vs NeoVim. This PR works only w/ NeoVim because I didn't succeed with Vim and I was looking for advice. 😃  I'm aware of your "Vim first" choice. 😉 

If you look at the code, I copied and pasted the logic from `GoTest`, trying to make it to work with Vim too: https://github.com/fatih/vim-go/pull/1198/files#diff-a3998a862a20bc1665a15c84f401349fR422

---

I'll try to extract a standalone plugin for this. Thank you very much for `vim-go`.

Carpetsmoker:
Note: this PR was continued in this plugin by the PR author: https://github.com/jodosha/vim-godebug

(For those who ended up here through a search).

Issue #1182, opened 2017-01-23
loclist jump doesn't work for :GoFmt


alaska:
### Behavior

Sometime during the last couple of releases (working to bisect), the errors listed in the location list from gofmt no longer jump to the line where the error is listed.  <CR> does nothing

### Steps to reproduce:

1. Open any .go file
2. Add a random number sign `#` - `U+0023` to an empty line
3. Save

### Expected results
- Loclist opens with output like the following:
`|184 col 2| : illegal character U+0023 '#'`
- hitting enter `<CR>` jumps to the line in the buffer you came from

### Actual results
- Loclist appears as above
- hitting enter `<CR>` has no visible effect

### Configuration

All other plugins disabled

### vimrc
```
set nocompatible
filetype off
set rtp+=~/.vim/bundle/Vundle.vim
call vundle#begin()
Plugin 'VundleVim/Vundle.vim'
Plugin 'fatih/vim-go'
call vundle#end()
filetype plugin indent on
let g:go_template_use_pkg = 1

"disable arrows
noremap <up> <nop>
inoremap <up> <nop>
noremap <down> <nop>
inoremap <down> <nop>
noremap <left> <nop>
inoremap <left> <nop>
noremap <right> <nop>
inoremap <right> <nop>

syntax on

" Colors and theme
set t_Co=256
set background=light
let g:nofrils_strbackgrounds=1
colorscheme nofrils-light

" Set up <TAB> length and behavior
set tabstop=4
set shiftwidth=4
set expandtab
" This makes backspace treat four spaces like a tab:
set softtabstop=4
set shiftround
```

### version
```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Jul 28 2016 14:11:47)
MacOS X (unix) version
Included patches: 1-2109
Compiled by Homebrew
Huge version without GUI.  Features included (+) or not (-):
+acl             +channel         +comments        +diff            +extra_search    +fork()          +keymap          -lua             -mouse_gpm       +multi_byte      +perl            +quickfix        +startuptime     -tcl             -toolbar         +vreplace        -xim
+arabic          +cindent         +conceal         +digraphs        +farsi           -gettext         +langmap         +menu            -mouse_jsbterm   +multi_lang      +persistent_undo +reltime         +statusline      +termguicolors   +user_commands   +wildignore      -xsmp
[...]

PR #1199, opened 2017-02-07
Set the filename in the location-list

tarrant:
At some point in the near past the location list seems to have broken as shown by https://github.com/fatih/vim-go/issues/1182. When comparing an older version of vim-go and master the difference seems to be due to not including the filename in the location list.

According to the help page for `setqflist()` the filename should be added to the dictionary as the key `filename`. This lead me to pass the filename into the function that generates the individual lines of the location list and then set the file on each error.

This fixes the #1182 on my computer, but my knowledge of vim script is a bit question and would appreciate further verification.


fatih:
Thanks @tarrant, this looks a good fix 👍

Issue #1091, opened 2016-10-31
GoAddTags bug when using  *interface{} fields


marcosnils:
### Steps to reproduce:

Given the following struct

```go
type MyStruct struct {
	Lala interface{}
}
```

Running GoAddTags doesn't do anything on this field or any following declared fields

This also happens for any type with interface{} sub-type

fatih:
Fixed with #1204 please, feel free to test and reopen if it's not the case.

Issue #1064, opened 2016-09-29
GoAddTags should not add twice


gnhuy91:
Sorry for the bad issue title, couldn't think of a better one.

```
    Number       string `json:"number"`
    TextAllowed  bool
```
### Actual behavior

```
    Number       string `json:"number"` `json:"number"`
    TextAllowed  bool `json:"text_allowed"`
```
### Expected behavior

```
    Number       string `json:"number"`
    TextAllowed  bool `json:"text_allowed"`

```

fatih:
Hi @gnhuy91 

Thanks for the bug report. This is one of the cases where an AST based tool is needed or things get complex easily. Let's tackle this as well with the other `:GoAddTags` related issues.

fatih:
Fixed with #1204 please, feel free to test and reopen if it's not the case.

Issue #990, opened 2016-08-03
:GoAddTags with nested structs


dvcrn:
Thanks a lot for `:GoAddTags`. It already saved me so much time!

I noticed that tag generation for nested structs seem to not work correctly:

```
type Foo struct {
    One   string
    Two   string
    Three string
    Bar   struct {
        Four string
        Five string
    }
}
```

after :GoAddTags

```
type Foo struct {
    One   string `json:"one"`
    Two   string `json:"two"`
    Three string `json:"three"`
    Bar   struct { `json:"bar"`   <------
        Four string `json:"four"`
        Five string `json:"five"`
    }
}
```

How it should be:

```

type Foo struct {
    One   string `json:"one"`
    Two   string `json:"two"`
    Three string `json:"three"`
    Bar   struct {
        Four string `json:"four"`
        Five string `json:"five"`
    } `json:"bar"`         <------
}
```

fatih:
Thanks @dvcrn 

I'm planning to move `:GoAddTags` to use the tool motion (which supports AST based output). It'll cover all these cases for us. Thanks!

dvcrn:
Hi fatih, is there anything new on this? 
If you could give me some implementation details I could also try to help you with this

fatih:
@dvcrn I'm currently working on this, will update and close this issue once it's ready

fatih:
Fixed with #1024 please, feel free to test and reopen if it's not the case.

Issue #985, opened 2016-08-01
GoAddTags: tag options


tobstarr:
At least for my workflow it would be nice to be able to also provide e.g. the `omitempty` option when adding a `json` tag to a struct field. 

I would propose to change the behaviour of `GoAddTags json,omitempty` to produce `json:"name,omitempty"` instead of `json,omitempty:"name"` but I do not know if there are cases where one actually would want to add a tag including a comma character.

fatih:
Hi @tobstarr 

This and the second PR (#984) you opened, I wanted to do in the next release ;)  I didn't wanted to put a lot of features. So adding them one by one is the way to go 👍 

Tag options is easy, however for the second one we need to parse the string (might be a little bit complicated, but doable). I'll keep an eye on this. If someone else want to tackle it feel free to pick it up, but let me know before if you do it.

fatih:
Fixed with #1024 please, feel free to test and reopen if it's not the case.

tobstarr:
@fatih is it intended that `GoAddTags json,omitempty` only works if the tag is already present? In that case things work fine, otherwise nothing happens (but something __is__ happening as the vim buffer is then marked as changed - `+`)

fatih:
Yeah it only works if it exists. Previously it was the opposite. The buffer changes because gomodifytags changes the file again, thus the creation time changes as well. Open an issue to https://github.com/fatih/gomodifytags to track if you want :)

tobstarr:
Any particular reason for that? I would find it really handy for `GoAddTags json,omitemty` to also create the tag of it does not exist. 

I just checked gomodifytag and the call `gomodifytags -file main.go -add-tags json -struct User -add-options json=omitempty` would do the trick (add the tag with the option if it does not exist).

Behaviour before was not the opposite, but that it created this: `json,omitempty:"name"`.

fatih:
When I mean `oppposite`, the current implementation was like you said. I've removed it because I wanted to make it explicitly. It's very easy to change it because I've like that previously. I'm not sure why I reverted it (don't remember now). Let me check it again.

PR #1204, opened 2017-02-13
New GoTags improvements

fatih:
* Use new [gomodifytags](https://github.com/fatih/gomodifytags) tool to manipulate and modify tags

Checkout demo:

vim-go-gomodifytags

* `:GoAddTags` is able to add tags and options
* `:GoRemoveTags` is able to remove tags and options

Usage:

Following command add multiple tags. Does not modify if the key already
exist. If `,optionaname` is given, it adds the given option to that key.
If called without arguments, it adds by default tags with `json` key.
```
:GoAddTags
:GoAddTags json
:GoAddTags json,omitempty
:GoAddTags json hcl bson
:GoAddTags json,omitempty hcl bson
```

Following command removes multiple tags and options. If `,optionname` is
given, it removes the option belonging to that key instead of removing
the key. If called without arguments, it removes all tags belonging to a
struct.

```
:GoRemoveTags
:GoRemoveTags json
:GoRemoveTags json,omitempty
:GoRemoveTags json hcl bson
:GoRemoveTags json,omitempty hcl bson
```

Fixes following issues:

 #984
 #985
 #990
 #1064
 #1091

### DIFF
--- a/autoload/go/tags.vim
+++ b/autoload/go/tags.vim
@@ -1,0 +1,206 @@
+function! go#tags#Add(start, end, count, ...) abort
+  let fname = fnamemodify(expand("%"), ':p:gs?\\?/?')
+  if &modified
+    " Write current unsaved buffer to a temp file and use the modified content
+    let l:tmpname = tempname()
+    call writefile(getline(1, '$'), l:tmpname)
+    let fname = l:tmpname
+  endif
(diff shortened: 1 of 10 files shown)


fsouza:
Looks Awesome To Me.

PR #1205, opened 2017-02-14
add :GoModifyTags

mattn:
gomodifytags


mattn:
Ah, you already done. https://github.com/fatih/vim-go/pull/1204

closing.

PR #1211, opened 2017-02-19
Respect go_fmt_options when running goimports

cespare:
I want to use goimports's `-local` flag, but vim-go ignores `go_fmt_options` when using goimports for some reason. This just removes that limitation.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -168,10 +168,9 @@ function! s:fmt_cmd(bin_name, source, target)
   " start constructing the command
   let cmd = [bin_path]
   call add(cmd, "-w")
+  call extend(cmd, split(g:go_fmt_options, " "))
 
-  if a:bin_name != "goimports"
-    call extend(cmd, split(g:go_fmt_options, " "))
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @cespare

hori-ryota:
@fatih I want to disuse `go_fmt_options` with `GoImports`. I have modified it before for using `-s` option with `gofmt`.

[Disuse go\_fmt\_options with GoImports by hori\-ryota · Pull Request \#590 · fatih/vim\-go](https://github.com/fatih/vim-go/pull/590)

fatih:
Ok. Thanks for the reminder. Seems like we have to use a different option  for `goimports` ,so it  doesn't break `goimports` when you set a flag that is a valid flag for `gofmt`, but not `goimports`

fatih:
I've opened this : https://github.com/fatih/vim-go/issues/1212 Dont' have time to tackle it now, but will soon. My plan is to introduce a `go_goimports_options` (or `go_imports_options` flag). If anyone wants to tackle it let me know.

PR #1214, opened 2017-02-20
Add go_imports_options resolve #1212

### LINKED ISSUE #1212: Breaking change https://github.com/fatih/vim-go/pull/1211
Should be reverted or fixed

hori-ryota:
Add `go_imports_options` for `goimports`.

Witch do you like `go_imports_options` or `go_goimports_options` ?

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -21,6 +21,10 @@ if !exists('g:go_fmt_options')
   let g:go_fmt_options = ''
 endif
 
+if !exists('g:go_imports_options')
+  let g:go_imports_options = ''
+endif
+
 if !exists("g:go_fmt_experimental")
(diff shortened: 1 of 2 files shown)


cespare:
Are we going to have go_X_options for every possible tool someone ever uses instead of gofmt?

fatih:
I agree with cespare that `:GoFmt` and `:GoImports` commands might be confusing when you change the command. I think we need to start using a dictionary  with a mapping of `command <-> options`. However I agree with @hori-ryota and I still believe that having both `:GoFmt` and `:GoImports` is much needed for large monorepos. I've pulled the ideas from both of you and created this PR: https://github.com/fatih/vim-go/pull/1308 I think this is a good compromise for now. Thanks for the feedback. Closing this in favor of #1308

Issue #389, opened 2015-04-14
:GoRename ignores files with +build tags


oryband:
I have a file with `+build integration` tag, which is ignored when renaming types with `:GoRename`.

fatih:
This is normal, because `integration` is not a satisfied tag and the go toolset will ignore it. Check the docs here:

http://golang.org/pkg/go/build/

Basically you do the same as `// +build ignore`.

oryband:
@fatih is there a way to send a `-tags` parameter to `:GoRename`? Right now I can't do anything. :\

fatih:
Fortunately not, this is a constraint of the go tools. Gorename doesn't have a clue about it. You can ask this to the golang-nuts mailing list, the source code is here: https://github.com/golang/tools/ and the author is Alan Donovan.

fatih:
Btw here is the list that satisfies the rule: 

```
- the target operating system, as spelled by runtime.GOOS
- the target architecture, as spelled by runtime.GOARCH
- the compiler being used, either "gc" or "gccgo"
- "cgo", if ctxt.CgoEnabled is true
- "go1.1", from Go version 1.1 onward
- "go1.2", from Go version 1.2 onward
- "go1.3", from Go version 1.3 onward
- "go1.4", from Go version 1.4 onward
- any additional words listed in ctxt.BuildTags
```

fatih:
@oryband ok I have found a fix for it, I need to open a PR to the official Go repo. It'll take some time though.

oryband:
@fatih thanks! can you elaborate?

fatih:
Sure: https://groups.google.com/forum/#!topic/golang-nuts/M7P0xae6iWM

oryband:
Thanks!

fatih:
Related feature set, Alan added it:  https://go-review.googlesource.com/#/c/9172/1 Once it's in usable state I'm going to introduce a new "go_build_tags" setting and once set it will be passed to all commands that support this new `-tags` flag.

oryband:
:+1:

PR #1225, opened 2017-03-01
Add build tags support for gorename

dim13:
Reopening issue #389, it seems like `gorename` supports `-tags` flag for a while now, still undocumented however.

See https://github.com/golang/tools/blob/master/cmd/gorename/main.go#L29 for reference.

This patch add support for this flag. I'm sure, I'm not the only one who misses this feature badly.

### DIFF
--- a/autoload/go/rename.vim
+++ b/autoload/go/rename.vim
@@ -41,6 +41,11 @@ function! go#rename#Rename(bang, ...) abort
 
   let cmd = [bin_path, "-offset", offset, "-to", to_identifier]
 
+  if exists('g:go_rename_tags')
+    let tags = get(g:, 'go_rename_tags')
+    call extend(cmd, ["-tags", tags])
+  endif
+
   if go#util#has_job()
     call go#util#EchoProgress(printf("renaming to '%s' ...", to_identifier))
     call s:rename_job({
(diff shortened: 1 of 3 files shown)


fatih:
Hi @dim13 

Thanks for the feature. Can you also add the `:GoRenameTags`  command. You can checkout `:GoGuruTags` and see checkout how it works.

dim13:
Updated.

On the second thought, combining of `GoGuruTags` with `GoRenameTags` to something like `GoBuildTags` would be IMHO more preferable solution.

If you don't mind, I may try to achieve it on separated pull-request, however it will break existing settings for go guru.

fatih:
@dim13 yeap, that's the goal actually. However I would prefer we first introduce :GoRenameTags and then we can later remove both and add `:GoCmdTags` or something similar (such as `:GoBuildTags`. I don't want to add two separate things into one pr.

dim13:
Well then, I'm with you. Please go ahead and test it (I'm not really savvy with vim-script, but it seems to work).

For me and my co-workers it already saves a lot of pain (we stuck with `+build integration` too). I would be happy to see it upstream.

And thank you for this great plugin! Could not live without. ;)

fatih:
Thanks @dim13. A nice addition to the family 👍

fatih:
Now that we have `:GoAddTags` and `:GoRemoveTags`, I'm thinking that we should change the command name and combine them (to avoid any confusion) . I'm going to work on that :)

fatih:
@dim13 see: https://github.com/fatih/vim-go/pull/1232

PR #1233, opened 2017-03-08
Updated Broken Link

rigelrozanski:
Old link for `doc/vim-go.txt` would lead to the URL https://github.com/fatih/vim-go/blob/master/.github/doc/vim-go.txt


fatih:
Thanks @rigelrozanski

PR #1235, opened 2017-03-10
Allow trimming around comments in gotexttmpl syntax.

thsnr:
In `text/template` starting from Go 1.6, whitespace around actions can be trimmed with "- " and " -" to result in cleaner output (https://golang.org/pkg/text/template/#hdr-Text_and_spaces). Update `syntax/gotexttmpl.vim` to allow using this construction around comments.


fatih:
Thanks @thsnr 👍

Issue #1237, opened 2017-03-17
goimports removes hyphenated imports if project is symlinked into GOPATH


jmrodri:
### Behavior
With goimports enabled in .vimrc, a file with "github.com/op/go-logging" in its imports, saving the file will remove the import. Removing the import causes the project not to build. After much digging this only happens if the go project was SYMLINKED into the GOPATH.

Expected Behavior:
When saving the file, I expect `goimports` to find the proper import and name it correctly instead of deleting it:  `logging "github.com/op/go-logging`

### Steps to reproduce:

Setup project with a symlink.

0. enable goimports in `.vimrc`: `let g:go_fmt_command = "goimports"`
1. mkdir -p /tmp/go/{src,pkg,bin}
2. export GOPATH=/tmp/
3. mkdir -p /tmp/mycode
4. cd /tmp/mycode
5. git clone https://github.com/fusor/ansible-service-broker.git
6. mkdir -p $GOPATH/src/github.com/fusor/
7. ln -s /tmp/mycode/ansible-service-broker/ $GOPATH/src/github.com/fusor/ansible-service-broker
8. cd $GOPATH/src/github.com/fusor/ansible-service-broker
9. vi ./pkg/broker/broker.go
10. NOTICE that there is a `"github.com/op/go-logging"` import.
```
import (                                                   
    "fmt"                                                  
                                                           
    "github.com/fusor/ansible-service-broker/pkg/ansibleapp"
    "github.com/fusor/ansible-service-broker/pkg/dao"      
    "github.com/op/go-logging"                             
    "github.com/pborman/uuid"                              
)                                                          
```
11. :w (save the file)
12. NOTICE that the `"github.com/op/go-logging"` import was removed.
```
import (                                                   
    "fmt"                                                  
                                                           
    "github.com/fusor/ansible-service-broker/pkg/ansibleapp
    "github.com/fusor/ansible-service-broker/pkg/dao"      
    "github.com/pborman/uuid"                              
)                                                          
```
### Configuration

Add here your current configuration and additional information that might be
useful, such as:

* `vimrc` you used to reproduce

```
" vim-go settings
let g:go_highlight_functions = 1                  " enable function highlighting
let g:go_highlight_methods = 1                    " enable method highlighting
let g:go_highlight_fields = 1                     " enable fields highlighting
let g:go_highlight_types = 1                      " enable types highlighting
let g:go_highlight_operators = 0                  " disable operator highlighting
[...]

PR #1238, opened 2017-03-17
goimports fails when proj symlinked in GOPATH

jmrodri:
Fixes Issue #1237 . Instead of using the fully qualified filename path, we just need the relative path. This allows `goimports` to always work either if the project is symlinked into the GOPATH or if it is directly checked into the GOPATH.

Tested on:
* Fedora 25
* VIM version 8.0.425
* vim-go: HEAD as of 17 March 2017


jmrodri:
I will investigate why the test fails on travis but works locally.

fatih:
You need to add the following:

```patch
diff --git a/scripts/test.sh b/scripts/test.sh
index b22fd7d..7c913ab 100755
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -6,6 +6,7 @@ cd $(dirname $0)
 
 # install dependencies
 go get github.com/fatih/gomodifytags
+go get golang.org/x/tools/cmd/goimports
 
 # cleanup test.log
 if [ -f "test.log" ]; then

```

fatih:
Fixed the test and merged it. Thanks for your contribution @jmrodri

jmrodri:
@fatih thanks for creating the project. It's awesome.

PR #1240, opened 2017-03-18
:GoGet added

thomasheller:
Adding support for `:GoGet`.

For now, just passing args (if any) to `go get` and not checking for errors. Not sure how to handle the output of `go get`, because unlike `go build` errors don't correspond to line numbers in the source files.

Any thoughts how this could be improved?


fatih:
Hi @thomasheller 

`:GoImport!`  already does it, see our doc `:help :GoImport`:

```
                                                                   *:GoImport*
:GoImport[!] [path]

    Import ensures that the provided package {path} is imported in the current
    Go buffer, using proper style and ordering. If {path} is already being
    imported, an error will be displayed and the buffer will be untouched.

    If [!] is given it will download the package with `go get`
```

Thanks

PR #1241, opened 2017-03-18
Add support for quicktemplate syntax to gohtmltmpl filetype

codelitt:
This is a fairly popular template package. It would be great to
support this. This causes no conflicts that I can see and is a very
simple addition. For more details see: [quicktemplate](https://github.com/valyala/quicktemplate)


fatih:
Thanks @codelitt for the PR. I don't think I want to merge this. This can be perfectly provdided by an external plugin. It's an addition for an additional 3rd party engine, and I'm not using it as well. Merging this means I have to maintain it for many months and even years. I hope you understand my decision. Thanks

PR #1242, opened 2017-03-23
Update GoDef help: remove explicit arguments

danielheller:
The GoDef command does not support specifiying identifiers on the
command line due to limitations in the Guru and Godef tools, which
operate on file positions. This change updates the help for GoDef to
remove an incorrect mention of explicit arguments.


fatih:
Hi @danielheller 

This looks good. `godef` itself actually also works on direct types instead of offset, however we don't support it for vim-go. Thanks for the PR.

PR #1247, opened 2017-03-28
job: fix race between channel close and job exit

pborzenkov:
When channel's close_cb is called, the job this channel is assigned to
may still be alive and the result status of the job will be set to
FAILED (as no actual return code is available at that moment). This is
easily reproduced for me on a single CPU Linux VPS.

To fix this, check for job's status in exit_cb instead.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -446,7 +446,7 @@ function s:cmd_job(args) abort
 
   let start_options = {
         \ 'callback': callbacks.callback,
-        \ 'close_cb': callbacks.close_cb,
+        \ 'exit_cb': callbacks.exit_cb,
         \ }
 
   " modify GOPATH if needed
--- a/autoload/go/coverage.vim
+++ b/autoload/go/coverage.vim
@@ -288,7 +288,7 @@ function s:coverage_job(args)
 
   let start_options = {
         \ 'callback': callbacks.callback,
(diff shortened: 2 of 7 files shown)


fatih:
I remember that exit_cb was not working well. I'm not sure if it's fixed recently. Thanks @pborzenkov , need to look into it and be sure it works fine. Also we have other places that still uses close_cb which is not fixed with this PR (such as `guru.vim`)

pborzenkov:
@fatih I believe there might be a potential problem with exit_cb, because there still can be data in channel's buffer at the time the callback is called:
```
                        Note that data can be buffered, callbacks may still be
                        called after the process ends.
```

I think this can be fixed by the following patch:
```
   function cbs.exit_cb(job, exitval) dict
+    let chan = job_getchannel(a:job)
+    while ch_canread(chan)
+      call add(self.messages, ch_read(chan))
+    endwhile
+
```

Though, I haven't encountered such problems yet (I have tested the patch on macOS and Linux).


Sure, I'll fix the rest of the places.

fatih:
@pborzenkov 

yeah that was the problem I had. Let me test this branch locally for couple of days. It's a very critical one that might break async build and co on anything.

fatih:
Alright extensively tested it. I think we can merge it to master (because master is a development branch). This would make it easier to spot errors, because people fetch from master frequently. Thanks @pborzenkov for the fix again.

Issue #1208, opened 2017-02-17
:GoRun does not respect default GOPATH (go 1.8+)


kron4eg:
### Behavior

When I'm trying to `:GoRun` main that contains import of library that is in default `$GOPATH`, it doesn't work.

```
main.go:7:2: cannot find package "github.com/kezhuw/toml" in any of:
	/usr/local/go/src/github.com/kezhuw/toml (from $GOROOT)
	/Users/kron/src/github.com/kezhuw/toml (from $GOPATH)
```

Note that `github.com/kezhuw/toml` is in `~/go/src/github.com/kezhuw/toml/`.
### Steps to reproduce:

Using Go 1.8 run

    unset $GOPATH
    go get github.com/kezhuw/toml

using this Go file

```
package main

import (
	"fmt"
	"time"

	"github.com/kezhuw/toml"
)

func main() {
	data := []byte(`key = 2016-01-07T15:30:30.123456789Z`)
	var out struct{ Key time.Time }

	err := toml.Unmarshal(data, &out)
	if err != nil {
		panic(err)
	}

	fmt.Println(out.Key.Format(time.RFC3339Nano))
}
```

Run `:GoRun` and observe errors (vim-go seems to setup default GOPATH to $HOME).

Run
```
$ go run main.go
2016-01-07T15:30:30.123456789Z
```

### Configuration

* vim version: 8.0.0329
* vim-go version: latest master
* go version: go version go1.8 darwin/amd64

fatih:
Hi. I've opened a PR with the fix: https://github.com/fatih/vim-go/pull/1248 can you please test it? Thanks

fatih:
Fixed with #1248

PR #1248, opened 2017-03-29
Go 1.8 compatible GOPATH improvements

### LINKED ISSUE #1243: Default GOPATH not respected in vim-go
This is related to #1208 but I believe the problem is more widespread.

When the environment variable $GOPATH isn't set, Go 1.8 uses $HOME/go as the default. 
In Vim the function :GoPath returns this path as expected however many operations depending on the $GOPATH being set don't work properly.

Vim-go can't identify the "Go binaries" are installed and a message is provided saying they aren't installed.
GoInstallBinaries can't identify existing binaries and re-installs them every time it's run.

GoCode can't be found and many features don't work.

GoFmt can't be found and an error is returned on file save.

I'm sure there are many similar manifestations of this similar issue.

It's easy to reproduce. Simply stop exporting $GOPATH and try to use vim-go.

fatih:
This PR fixes the following cases:

* If no GOPATH is set, vim-go uses the default GOPATH via `go env
  GOPATH` (starting Go 1.8)
* `:GoPath` now shows the correct `GOPATH` if no GOPATH is set
* `:GoInstallBinaries` now installs to the correct GOPATH if no GOPATH
  is set

fixes #1243 #1208

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -218,7 +218,7 @@ function! go#cmd#Install(bang, ...) abort
   if !empty(errors) && !a:bang
     call go#list#JumpToFirst(l:listtype)
   else
-    call go#util#EchoSuccess("installed to ". $GOPATH)
+    call go#util#EchoSuccess("installed to ". go#path#Detect())
   endif
 
   let $GOPATH = old_gopath
@@ -403,7 +403,6 @@ function! go#cmd#Generate(bang, ...) abort
   let &makeprg = default_makeprg
   let $GOPATH = old_gopath
 endfunction
-
 
 " ---------------------
 " | Vim job callbacks |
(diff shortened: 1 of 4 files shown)


fatih:
@spf13 can you try this branch? I've removed my GOPATH via `unset GOPATH` and then just started vim-go. It should now without issues.

kron4eg:
@fatih I'm getting

```Error detected while processing function <SNR>53_fmt_autosave[3]..go#fmt#Format[40]..go#fmt#run:
line    2:
E691: Can only compare List with List```

When I'm trying to save. On `master` branch it's all good (expect default GOPATH obviously :) )

fatih:
@kron4eg there was a bug, I've fixed and pushed to this branch (amended). I'm also testing it currently during my work. Please don't hesitate to give any feedback and thanks! :)

kron4eg:
Yep, now it works, all good. As well as default GOPATH. Thanks @fatih!

PR #1258, opened 2017-04-11
Include keyify support

thalesmello:
Includes support for https://github.com/dominikh/go-tools/tree/master/cmd/keyify
through the `GoKeyify` command.

It turns unkeyed struct literals into keyed struct literals.

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -21,6 +21,7 @@ let s:packages = [
       \ "github.com/fatih/gomodifytags",
       \ "github.com/zmb3/gogetdoc",
       \ "github.com/josharian/impl",
+      \ "github.com/dominikh/go-tools/tree/master/cmd/keyify",
       \ ]
 
 " These commands are available on any filetypes
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -21,7 +21,7 @@ let s:packages = [
       \ "github.com/fatih/gomodifytags",
       \ "github.com/zmb3/gogetdoc",
       \ "github.com/josharian/impl",
-      \ "github.com/dominikh/go-tools/tree/master/cmd/keyify",
+      \ "github.com/dominikh/go-tools/cmd/keyify",
       \ ]
 
 " These commands are available on any filetypes
--- a/autoload/go/keyify.vim
+++ b/autoload/go/keyify.vim
@@ -16,7 +16,7 @@ function! go#keyify#Keyify()
 
   " We want to output the error message in case the result isn't a JSON
   if type(result) != type({})
-    echoerr s:chomp(output)
+    call go#util#EchoError(s:chomp(output))
     let $GOPATH = old_gopath
     return
   endif
--- a/autoload/go/keyify.vim
+++ b/autoload/go/keyify.vim
@@ -34,7 +34,16 @@ function! go#keyify#Keyify()
   " Replace contents between start and end with `replacement`
   call setpos("'<", start)
   call setpos("'>", end)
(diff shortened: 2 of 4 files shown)


plugin/go.vim
```diff
@@ -21,6 +21,7 @@ let s:packages = [
       \ "github.com/fatih/gomodifytags",
       \ "github.com/zmb3/gogetdoc",
       \ "github.com/josharian/impl",
+      \ "github.com/dominikh/go-tools/tree/master/cmd/keyify",
```
fatih: This is a wrong import path, should be: `github.com/dominikh/go-tools/cmd/keyify`

fatih:
Hi @thalesmello 

I just tried it. The example I've used is:

```go
package main

import "fmt"

type Foo struct {
	foo string
	bar string
	qux string
}

func main() {
	f := Foo{"foo", "bar", "qux"}
	fmt.Printf("f = %+v\n", f)
}
```

1. When I put my cursor on top of `f` or before `Foo` nothing happens. I think we should support that case, not sure if we can do it without adding support to `keyify`. If not we should return a message saying that the cursor is not on top of the struct. `keyify` already does it, we should display it.

2. If I put my cursor on `Foo` it converts it to:

```go
package main

import "fmt"

type Foo struct {
	foo string
	bar string
	qux string
}

func main() {
	Foo{
		foo: "foo",
		bar: "bar",
		qux: "qux",
	}
	fmt.Printf("f = %+v\n", f)
}
```

As you see it removes the variable all together. I assume the replacement is not done correctly.

plugin/go.vim
thalesmello: Thanks. Changed it.

thalesmello:
@fatih I agree with you on 1.

I think we could implement some kind of mechanism to try to figure out the position of the struct, however, I can't think of a good way of implementing it in a good, clear way.

But I totally agree that it at least should output some nice error message.

In my first implementation, I simply did nothing when I `keyify` didn't return a refactor.

What I did, then, was simply return `keyify`'s error message, which is actually pretty useful.

Now it returns 

```
Error detected while processing function go#keyify#Keyify:
line   18:
no composite literal found near point
```
When it doesn't find any struct. What do you think about this?

2. Hum, strange. It work correctly for me.

A common issue with `keyify` is that it has to run inside the $GOPATH/src folder, otherwise it isn't able to work.

Could you please check my changes?

fatih:
> When it doesn't find any struct. What do you think about this?

Do not use `echoerr`. Instead use our own echo messages as `echoerr` is not user friendly. Here is an example: 

```
    call go#util#EchoError(s:chomp(output))
```

> A common issue with keyify is that it has to run inside the $GOPATH/src folder, otherwise it isn't able to work.

How do we know that it doesn't work. Does it return an error? I'm always inside a GOPATH. vim-go also has with Go 1.8 a default GOPATH as well. 

> Hum, strange. It work correctly for me.

The problem occurs if you previously have selected anything. For example do a line seletion via "shift-v" and then call `:GoKeyify" when your cursor is on `Foo`, you'll see it'll replace with the wrong selection.

thalesmello:
@fatih 

I changed the echo function used.

About the bug, here is what happened:

1. You used a "V" visual mode
2. GoKeyify relies on using a visual range in order to replace some text
3. Because the previous visual mode was "V", it's used when you type "gv"
4. GoKeyify replaces the entire lines content because of this

In order to prevent this bug, now the function checks if the correct visual mode is being used.

Could you check this last version, please?

fatih:
Thanks @thalesmello this looks now good 👍

PR #1264, opened 2017-04-18
support <C-LeftMouse> as well

lilydjwg:


fatih:
Hi @lilydjwg 

Why did you introduced this change? And for what is this used?

lilydjwg:
It's the same as '<C-]>` in vim's default behavior. They are listed together in the doc.

It's convenient when one is browsing the code.

fatih:
I see thanks @lilydjwg

PR #1268, opened 2017-04-23
Include if with inline error handling

Foxboron:
This pattern is widely enough used to be a snippet i believe.
Examples from one golang.org example.
 ```
// ...
	if err := cmd.Start(); err != nil {
		log.Fatal(err)
	}
// ...
	if err := json.NewDecoder(stdout).Decode(&person); err != nil {
		log.Fatal(err)
	}
	if err := cmd.Wait(); err != nil {
		log.Fatal(err)
	}
 ```

### DIFF
--- a/gosnippets/UltiSnips/go.snippets
+++ b/gosnippets/UltiSnips/go.snippets
@@ -143,6 +143,13 @@ else {
 }
 endsnippet
 
+# if inline error
+snippet ife "If with inline erro"
+if err := ${1:condition}; err != nil {
+	${0:${VISUAL}}
+}
+endsnippet
+
 # error snippet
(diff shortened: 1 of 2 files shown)


fatih:
I think the current snippets are well defined @Foxboron You're free to define your own ones. Plus no ones fatal.

We can improve two things here:

1. remove the `log.Fatal` part and jump there with the next step
2. also add it to neopsnippet file

Once these are done, I think we can merge this PR.

Foxboron:
Done. Also modified the commit message. I don't use neosnippet so i hope that definition is correct.

fatih:
Thanks @Foxboron 👍

Issue #1224, opened 2017-02-28
Vim-Go broken on freebsd


soardex:
### Behavior
Getting this error while running `:GoInstallBinaries` and saving go file.

```
Error detected while processing function <SNR>25_fmt_autosave[3]..go#fmt#Format[40]..go#fmt#run[11]..go#util#System:
line    7:
E484: Can't open file /tmp/vvztQvl/1
Error detected while processing function <SNR>25_fmt_autosave:
line    3:
E171: Missing :endif
```
Just using minimal .vimrc using vim-plug

* vim version: 8.0 (2016 Sep 12, compiled Feb 23 2017 03:04:12)
* vim-go version: master
* go version: go1.8 freebsd/amd64

fatih:
Hi @soardex 

I'm not using freebsd and to be honest I don't have zero clue what to do. If you know how to fix it (if you thinks there is a problem), please open a pull request. Other than this vim-go works fine on linux, macOS and windows. Thanks

sopparus:
I've noticed it works with sh and bash if you set it with vipw, so it seems to break only with default shell tcsh

creaky:
This is a compatibility bug between shells. I have raised a pull request to work around it.

vim-go system calls fails when using tcsh.

Having the [t]csh shells set as the login shell results in temporary file access errors when calling ksh or dash system shells. /bin/sh is often aliased to these shells on modern Linux and BSD systems including VoidLinux, Debian and FreeBSD.

As more Linux systems move away from bash to dash for the system shell, this issue will occur more often.

PR #1276, opened 2017-05-05
Fixed issue Vim-Go broken on freebsd #1224

creaky:
vim-go system calls fails when using tcsh.

Having the [t]csh shells set as the login shell results in
temporary file access errors when calling ksh or dash system
shells. /bin/sh is often aliased to these shells on modern
Linux and BSD systems including VoidLinux, Debian and FreeBSD.

As more Linux systems move away from bash to dash for the system
shell, this issue will occur more often.

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -99,7 +99,13 @@ endfunction
 function! go#util#System(str, ...) abort
   let l:shell = &shell
   if !go#util#IsWin() && executable('/bin/sh')
-    let &shell = '/bin/sh'
+    " Having the [t]csh shells set as the login shell results in temporary
+    " file access errors when calling ksh or dash system shells. /bin/sh is
+    " often aliased to these shells on modern Linux and BSD systems including
+    " VoidLinux, Debian and FreeBSD.
+    if match(l:shell, 'csh$') == -1
+      let &shell = '/bin/sh'
+    endif
   endif
 
   try
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -94,25 +94,29 @@ function! go#util#osarch() abort
   return go#util#goos() . '_' . go#util#goarch()
 endfunction
 
-" System runs a shell command. It will reset the shell to /bin/sh for Unix-like
-" systems if it is executable.
+" System runs a shell command. If possible, it will temporary set 
+" the shell to /bin/sh for Unix-like systems providing a Bourne
+" POSIX like environment.
 function! go#util#System(str, ...) abort
+  " Preserve original shell and shellredir values
   let l:shell = &shell
+  let l:shellredir = &shellredir
+
+  " Use a Bourne POSIX like shell. Some parts of vim-go expect
+  " commands to be executed using bourne semantics #988 and #1276.
+  " Alter shellredir to match bourne. Especially important if login shell
+  " is set to any of the csh or zsh family #1276.
   if !go#util#IsWin() && executable('/bin/sh')
-    " Having the [t]csh shells set as the login shell results in temporary
-    " file access errors when calling ksh or dash system shells. /bin/sh is
-    " often aliased to these shells on modern Linux and BSD systems including
-    " VoidLinux, Debian and FreeBSD.
-    if match(l:shell, 'csh$') == -1
-      let &shell = '/bin/sh'
-    endif
+      set shell=/bin/sh shellredir=>%s\ 2>&1
   endif
 
   try
     let l:output = call('system', [a:str] + a:000)
     return l:output
   finally
+    " Restore original values
     let &shell = l:shell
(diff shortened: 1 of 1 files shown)


autoload/go/util.vim
```diff
@@ -99,7 +99,13 @@ endfunction
...
   let l:shell = &shell
   if !go#util#IsWin() && executable('/bin/sh')
-    let &shell = '/bin/sh'
+    " Having the [t]csh shells set as the login shell results in temporary
+    " file access errors when calling ksh or dash system shells. /bin/sh is
+    " often aliased to these shells on modern Linux and BSD systems including
+    " VoidLinux, Debian and FreeBSD.
+    if match(l:shell, 'csh$') == -1
+      let &shell = '/bin/sh'
+    endif
```
Carpetsmoker: I'm confused. This will *not* reset the shell when using csh, right?

This will break commands that assume Bourne shell syntax. Resetting the shell to `/bin/sh` is a feature (I added this in #988).

Carpetsmoker: By the way, another way to do this is to introduce a new `g:go_shell` setting or some such. That's that syntastic does (see https://github.com/vim-syntastic/syntastic/issues/1355).

Carpetsmoker: Also related: https://github.com/junegunn/vim-plug/issues/159

creaky: Correct, this change will NOT reset the shell when running under *csh family. So far this has not caused any compatibility problems. 

The feature added in #988 breaks functionality and limits the portability of vim-go. The existing Windows bypass is evidence of this, and in the same vein, a bypass for *csh users (on systems using dash, zsh or ksh as /bin/sh) is needed until something more sophisticated is coded.

An option is to allow the user to specify an appropriate bourne like shell to drop down to ala g:go_shell specifying bash however do note FreeBSD and Solaris does not have bash installed by default. Further, this requires the user to do more configuring and really impacts the user experience of "it just works".

Alternatively, if someone has time, a very deep dive into why temporary file creation fails under dash/zsh/ksh as subshell to *csh and fix it would address a root cause of needing this workaround.

Best option is to keep command system calls simple, compatible with common shells and not be dependent on any shell specific behaviour, thus removing any need to override the user's shell.

So far, commands observed using this system call falls into the best option category and thus using the *csh skip is better than the alternative of doing nothing and breaking vim-go on a number of systems which is what is happening today.

Carpetsmoker: This patch will almost certainly break things. I didn't add it just for fun, I added it because stuff broke ;-) I don't remember what exactly though, and unfortunately I neglected to mention what exactly in the commit/PR description.

What problems are you having? Did you try setting the shellredir option as mention? Because I think this will fix it, and should be what we're changing?

> The feature added in #988 breaks functionality and limits the portability of vim-go.

?

It does exactly the opposite, by setting a known shell we ensure that all commands *always* work.

creaky: > What problems are you having?

The regression introduced by #988 causes the following problems in a *csh, zsh login environment on a system with /bin/sh pointing to dash, ksh or BSD sh.

Problem 1. Running 'vim test.go' results in errors and no template.
Problem 2. Unable to save a modified *.go file. Errors in save process prevents successful write.
Problem 3. Most vim-go commands resulting in calling the sub-shell fail with error.

Example error message produced from 'vim test.go':
`Error detected while processing function <SNR>27_template_autocreate[3]..go#temp
late#create[8]..go#tool#PackageName[2]..go#tool#ExecuteInDir[4]..go#util#env[7].
.go#util#goroot[1]..go#util#System:
line   17:
E484: Can't open file /tmp/vQAaQRG/1
Error detected while processing function <SNR>27_template_autocreate[3]..go#temp
late#create[8]..go#tool#PackageName[2]..go#tool#ExecuteInDir[4]..go#util#env:
line    7:
E171: Missing :endif
Error detected while processing function <SNR>27_template_autocreate:
line    3:
E171: Missing :endif
`

creaky: The shellredir is the right way to go. Expanding on the deep dive option mentioned above:

The 'shellredir' option controls what is captured by system(). From Vim help page: "The default is ">".  For Unix, if the 'shell' option is "csh", "tcsh" or "zsh" during initializations, the default becomes ">&".  If the 'shell' option is "sh", "ksh" or "bash" the default becomes ">%s 2>&1".  This means that stderr is also included.

The initialization of this option is done after reading the ".vimrc" and the other initializations, so that when the 'shell' option is set there, the 'shellredir' option changes automatically unless it was explicitly set before."

So from #988  the shell was changed without updating shellredir to match which causes the error for the csh/tcsh/zsh users.

I have tested this successfully and will update the pull request.

fatih:
Thanks @creaky @Carpetsmoker

PR #1278, opened 2017-05-05
Fixed indent for 'ife' snippet

dorons:
Fixed neosnippet errors when loading a go file due to missing indentation.


fatih:
Thanks @dorons

PR #1280, opened 2017-05-07
path: return the unmodified GOPATH if autodetect is disabled

### LINKED ISSUE #1257: Change in default detection of GOPATH breaks when $GOPATH is a list
In 6db12a966ec2ce9be9351621fa5a4ad790b7f623 the default handling of `$GOPATH` was changed:

https://github.com/fatih/vim-go/commit/6db12a966ec2ce9be9351621fa5a4ad790b7f623#diff-827b83031bfe1e9b104191a47bf8c2a7R79

_The use case here is `:GoInfo` incidentally; `go#path#Detect()` is called from [`complete.vim`](https://github.com/fatih/vim-go/blob/6db12a966ec2ce9be9351621fa5a4ad790b7f623/autoload/go/complete.vim#L26)_

Prior to this commit, unless `g:go_autodetect_gopath`was set to `1` (which it is not in my case), then `go#path#Detect()` would return `$GOPATH`. Following this commit, `go#path#Detect()` returns `go#path#Default()` _which is a transformation of `$GOPATH`_.

For setups where `$GOPATH` is a single value, I don't this this change has any effect.

But it breaks setups where `$GOPATH` is set to a list of paths. 

I couldn't see a specific mention of this in the [release notes](https://github.com/fatih/vim-go/blob/master/CHANGELOG.md#112---march-29-2017). 

### Steps to reproduce:

1. Set `$GOPATH` to a list of paths
2. Run `:GoInfo` on a type that you would expect to resolve as valid, and it will show as `invalid type` (because the `GOPATH` passed to `gocode` is now incorrect)

If you insert an `echom` in an appropriate place in `complete.vim` you can see what's going on:

```vim
" complete.vim

...
  " we might hit cache problems, as gocode doesn't handle well different
  " GOPATHS: https://github.com/nsf/gocode/issues/239
  let old_gopath = $GOPATH
  let old_goroot = $GOROOT
  let $GOROOT = go#util#env("goroot")

  echom $GOPATH
...
```

Before 6db12a966ec2ce9be9351621fa5a4ad790b7f623 the output for me is:

```
/home/myitcv/mgo/src/abc.com/g/_vendor:/home/myitcv/mgo
```

(which is correct). Following 6db12a966ec2ce9be9351621fa5a4ad790b7f623 the output for me is:

```
/home/myitcv/mgo/:/home/myitcv/mgo/src/abc.com/g/_vendor
```

which is incorrect and causes `gocode` to return `invalid type`.

### Configuration

No special config required to reproduce this

* `vimrc` plain and simple
* vim version: `NVIM v0.2.0-1171-g4566f7c7`
* vim-go version: 6db12a966ec2ce9be9351621fa5a4ad790b7f623
* go version: `go version go1.8 linux/amd64`
* I'm using https://github.com/mdempsky/gocode as opposed to https://github.com/nsf/gocode

fatih:
fixes #1257

### DIFF
--- a/autoload/go/path.vim
+++ b/autoload/go/path.vim
@@ -33,22 +33,16 @@ function! go#path#GoPath(...) abort
   let $GOPATH = a:1
 endfunction
 
-" Default returns the default GOPATH. If there is a single GOPATH it returns
-" it. For multiple GOPATHS separated with a the OS specific separator, only
-" the first one is returned. If GOPATH is not set, it uses the default GOPATH
-" set starting with GO 1.8. This GOPATH can be retrieved via 'go env GOPATH'
+" Default returns the default GOPATH. If GOPATH is not set, it uses the
+" default GOPATH set starting with Go 1.8. This GOPATH can be retrieved via
+" 'go env GOPATH'
 function! go#path#Default() abort
   if $GOPATH == ""
     " use default GOPATH via go env
     return go#util#gopath()
   endif
 
-  let go_paths = split($GOPATH, go#util#PathListSep())
-  if len(go_paths) == 1
-    return $GOPATH
-  endif
-
-  return go_paths[0]
+  return $GOPATH
 endfunction
 
 " HasPath checks whether the given path exists in GOPATH environment variable
@@ -134,8 +128,8 @@ function! go#path#BinPath() abort
   elseif $GOBIN != ""
     let bin_path = $GOBIN
   else
-    " GOPATH (user set or default GO)
-    let bin_path = expand(go#path#Default() . "/bin/")
+    let go_paths = split(go#path#Default(), go#util#PathListSep())
+    let bin_path = expand(go_paths[0] . "/bin/")
   endif
 
   return bin_path


autoload/go/path.vim
```diff
@@ -134,8 +128,8 @@ function! go#path#BinPath() abort
   elseif $GOBIN != ""
     let bin_path = $GOBIN
   else
-    " GOPATH (user set or default GO)
-    let bin_path = expand(go#path#Default() . "/bin/")
+    let go_paths = split(go#path#Default(), go#util#PathListSep())
```
myitcv: @fatih very much a drive-by comment... i.e. I haven't spent a huge amount of time looking into the context.

But can you not use `go list` to find the location of a binary?

```bash
go list -f '{{.Dir}}' github.com/nsf/gocode
```

That way you can avoid any assumptions about which part of `GOPATH` to use...

fatih: @myitcv this returns the path where the binaries might be installed, not the path of the binary itself. However I assume we can use your approach inside `CheckBinPath()` where it looks for the path of the binary itself

myitcv: 👍

Issue #268, opened 2014-12-17
Run GoTest from a remote directory. 


sheki:
First of all, I am a huge fan of Vim-go. Thanks for doing this. 

I launch vim from a central directory and run GoTest on an open file, this runs fine but the QuickFix Window does not work as it has file entries with the 

Path of the File and the PWD appended, is there a way to run GoTest if I am not in the directory the test_file is in?

fatih:
Ping @sheki  :)

sheki:
I do not have a good repro at this point of time. I will reopen the task if I run into it again, with a screen cast. Cheers

PR #1288, opened 2017-05-12
Fix GoDef for paths with spaces

Weyker:


fatih:
We already remove the whitespaces before and after the binary. Can you give a more concrete example how the binary looks like?

Weyker:
The problem its not the binary name ( godef or guru ), its the path to the file and the binary.

I have this path **/Volumes/Macbook HDD/Users/b...**. (the space between the Macbook and the HDD its the problem).

When i execute the ***:GoDef*** command, i have this error: 
`vim-go: /bin/sh: /Volumes/Macbook: No such file or directory`

So i print the `fname` and the `bin_path` and the final command and i got this:
`/Volumes/Macbook HDD/Users/beckmann/Development/go/bin/guru definition /Volumes/Macbook HDD/Users/beckmann/Development/go/src/github.com/jorgeaguirre/oju/cmd/filserver/main.go:#268`
(can't find guru... and the same problem with the file)

I Use NeoVim 0.1.7
go-vim updated with master.

fatih:
Can you use the `go#util#Shellescape()` function please?  It does it the quoting for you already.

PR #1292, opened 2017-05-18
Highlight built-in init() function differently

rhysd:
In Go, `func init()` has built-in special meaning. I thought it should be distinguished from other normal functions. So I modified highlight rules to highlight `init()` in `goBuiltins` color.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -294,17 +294,19 @@ hi def link     goOperator          Operator
 
 " Functions;
 if g:go_highlight_functions != 0
-  syn match goDeclaration       /\<func\>/ nextgroup=goReceiver,goFunction skipwhite skipnl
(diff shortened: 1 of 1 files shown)


fatih:
Hi @rhysd 

I don't think this is usable. This is how it looks like:

screen shot 2017-05-22 at 9 42 21 pm

if `init` is special, so is `main` as well. It's a function name and we shouldn't add too much coloring. I'm ok if you add a syntax new group but do not put it under the highlight setting. That way you can change it in your own `vimrc` without affecting other people's current setting.

rhysd:
hmm... I agree your idea. I did not consider other special function such as `main`. And I also think adding too much coloring is not good. Thank you for your comment. I think this should be closed.

Issue #1263, opened 2017-04-16
GoLint only operates on the current file.


alercah:
### Behavior

When `GoLint` is run with no arguments, it only lints the current file rather than the entire package, as `GoBuild` and `GoTest` do.

### Steps to reproduce:

1. Create a package with two files `a.go` and `b.go`. Both should build and a.go should have a lint error.
2. Open `b.go`.
3. Run `GoLint`.
4. Observe no errors reported.
5. Run `GoLint path/to/package`.
6. Observe an error reported in `a.go`

### Configuration

Vim 8.0
vim-go 1.11
Go 1.7.4

fatih:
Hi @alercah 

The feature was added recently (2 months ago) to golint. One of the reasons it's per file is that golint only could do it on a single file. Here is the recent change to it: https://github.com/golang/lint/pull/271 
I'm labeling this as `help-wanted`. If I have time I'll do it, if not someone else can tackle it. /cc @bhcleek

fatih:
`:GoLint` now works on the current directory instead of the file. Please pull the latest master. Thanks and feel free to open another issue if there are problems.

PR #1295, opened 2017-05-19
Change GoLint command to lint the current directory

etdub:
With no arguments passed to GoLint, the golint command will be executed on the current directory.

#1263


fatih:
Thanks @etdub 👍

PR #1297, opened 2017-05-23
Specifies the type of list to use in GoMetaLinter, by go_list_type

812lcl:
For async mode, GoMetaLinter will use quickfix as default type of list and it can't be configured.


fatih:
can you please rebase on top of master @812lcl ?

fatih:
I've did it by cherry picking your changes and rebased it on top of master. Thanks for your contribution.

Issue #1283, opened 2017-05-11
go_highlight_types for interface


0x04C2:
### Behavior

Hi, 

I set `let g:go_highlight_types = 1` and have highlight issue here. Please check the following picture. I think `interfaceEncoder` should not be highlighted. Does it make senses from original design? 

The reference code is [here](https://github.com/fatih/vim-go/blob/master/syntax/go.vim#L327). 

![screen shot 2017-05-11 at 10 42 55](https://cloud.githubusercontent.com/assets/17510957/25930186/7456075a-3637-11e7-824e-a778edeaec04.png)

### Steps to reproduce:

Re produce it with a minimal `vimrc` with all plugins disabled and
only `vim-go` enabled:

```
let g:go_highlight_functions = 1
let g:go_highlight_methods = 1
let g:go_highlight_fields = 1
let g:go_highlight_types = 1
let g:go_highlight_operators = 1
let g:go_highlight_build_constraints = 1
```

### Configuration

Add here your current configuration and additional information that might be
useful, such as:

* vim version: `VIM - Vi IMproved 8.0`
* vim-go version: `v1.12`
* go version: `go version go1.8 darwin/amd64`

zkry:
The error is caused by the syntax check of the keywords interface and struct.
syn match goDeclType             /\<interface\|struct\>/ skipwhite skipnl

Interestingly enough, this also causes the "struct" keyword to be highlighted when it is at the end of a word, like in encoder**Struct**. I made a pull request to fix this.

0x04C2:
@zkry Thank you! 
@fatih does it make senses?

fatih:
This is fixed now with https://github.com/fatih/vim-go/pull/1301 thanks to @zkry. Please pull latest master and thanks for the feedback.

PR #1301, opened 2017-05-25
Fix Issue #1283 Resolve highlighting error by surrounding expression in \(\)

zkry:
/\<interface\|struct\>/  would highlight **interface** or **struct** when **interface** was at the beginning of the word or **struct** at the end. By putting this in parenthesis, the word boundary checks correctly.


fatih:
Looks good, thanks @zkry 👍

PR #1306, opened 2017-05-28
Add on-demand loading option of vim-plug to README

178inaba:


fatih:
vim-go already uses the `autoload` folder, so it's only available if you use a `go`file. Besides that we have some commands that needs to be available even if it's not a Go file (Such as `:GoInstallBinaries`). Thanks

PR #1318, opened 2017-06-04
Use a different template for test files

### LINKED ISSUE #1216: Template for test files (enhancement)
When we open a new `.go` file, vim-go inserts the `hello_world.go` template.

When opening a new `_test.go` file, I would expect vim-go to insert a slightly more appropriate template, like:

```go
package main

import "testing"

func TestHelloWorld(t *testing.T) {
}
```

bhcleek:
Use a template more appropriate for test files than templates/hello_world.go
when creating a test file in a directory with no Go files present. A new
option, g:go_template_test_file, allows users to specify their own template to
use.

Fixes #1216

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1485,9 +1485,9 @@ Specifies whether `gocode` should use a different socket type. By default
                                                   *'g:go_template_autocreate'*
 
 When a new Go file is created, vim-go automatically fills the buffer content
-with a Go code template. By default the template under
-`templates/hello_world.go` is used. This can be changed with the
-|'g:go_template_file'| setting.
+with a Go code template. By default, the templates under the `templates`
+folder are used.  This can be changed with the |'g:go_template_file'| and
+|'g:go_template_test_file'| settings.
 
 If the new file is created in an already prepopulated package (with other Go
 files), in this case a Go code template with only the Go package declaration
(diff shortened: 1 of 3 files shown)


templates/hello_world_test.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import "testing"
+
+func TestHelloWorld(t *testing.T) {
+	t.Fatal("not implemented")
```
fatih: I think this should be commented out by default. It's better to give people a clean pkg instead of a test that's fails. Could you comment this out? The function body should be like:

```
// t.Fatal("not implemented")
```

doc/vim-go.txt
```diff
@@ -1508,6 +1508,14 @@ the `hello_world.go` file is used.
...
   let g:go_template_file = "hello_world.go"
 <
+                                                        *'g:go_template_test_file'*
+
+Specifies the file under the `templates` folder that is used if a new Go test
+file is created. Checkout |'g:go_template_autocreate'| for more info. By
+default the `hello_world_test.go` file is used.
+>
+  let g:go_template_test_file = "hello_world_test.go"
+
```
fatih: the `<` character is missing here

doc/vim-go.txt
```diff
@@ -1485,9 +1485,9 @@ Specifies whether `gocode` should use a different socket type. By default
                                                   *'g:go_template_autocreate'*
 
 When a new Go file is created, vim-go automatically fills the buffer content
-with a Go code template. By default the template under
-`templates/hello_world.go` is used. This can be changed with the
-|'g:go_template_file'| setting.
+with a Go code template. By default, the templates under `templates` are used.
```
fatih: I think `under templates folder are used ...` is better. People might not know that we have a `templates` folder.

fatih:
Thanks @bhcleek. A couple of improvements, once you fix them this is good to go

bhcleek:
PTAL

fatih:
Thanks @bhcleek 👍

PR #1322, opened 2017-06-07
GoImpl: support relative (vendor) imports

boz:
This fixes `:GoImpl` for cases when the interface package is in `vendor` or similar.  It depends on josharian/impl#18, which adds the `-dir` argument (`-dir` defaults to `$PWD` so hopefully the fallback won't be used much).

### DIFF
--- a/autoload/go/impl.vim
+++ b/autoload/go/impl.vim
@@ -40,7 +40,8 @@ function! go#impl#Impl(...) abort
   endif
 
   try
-    let result = go#util#System(join(go#util#Shelllist([binpath, recv, iface], ' ')))
+    let dirname = fnameescape(expand('%:p:h'))
+    let result = go#util#System(join(go#util#Shelllist([binpath, '-dir', dirname, recv, iface], ' ')))
     let result = substitute(result, "\n*$", "", "")
     if go#util#ShellError() != 0
       call go#util#EchoError(result)


autoload/go/impl.vim
```diff
@@ -35,8 +35,14 @@ function! go#impl#Impl(...) abort
...
-    return
+    call go#util#EchoWarning("impl failed.  trying with package path")
+
+    let s:dirname = fnameescape(expand('%:p:h'))
+    let result = go#util#System(printf("%s -dir '%s' '%s' '%s'", binpath, s:dirname, recv, iface))
+    if go#util#ShellError() != 0
+      call go#util#EchoError(result)
+      return
+    end
   endif
```
fatih: I'm not sure why we run another if this fails. If you could give me an example why it fails I can see why you did this. Why if we always pass the -dir? What exactly does this fix?

boz: * The current version of the `impl` command ignores `vendor` when finding the interface to generate.  It will fail if the interface is installed in `vendor` but not `$GOPATH`.
 * josharian/impl#18 adds `-dir <dir>` to root the normal go package resolution rules at `<dir>`, which fixes the `vendor` issue.  `<dir>` defaults to `$PWD`.
 * The current version of `impl` does not take a `-dir <dir>` argument.

Maybe an example will help.  `impl.new` is from josharian/impl#18 and `impl.orig` is the current release:

```
$ mkdir -p $GOPATH/src/test.com/user/vim-go-impl
$ cd $GOPATH/src/test.com/user/vim-go-impl
$ mkdir -p vendor/src/test.com/foo
$ cat > vendor/src/test.com/foo/foo.go <<EOF
> package foo
> type I interface {
>   Bar()
> }
> EOF
$ impl.orig 'i *I' foo.I
interface foo.I not found: couldn't find package src/test.com/foo: cannot find package "src/test.com/foo" in any of:
  /usr/local/Cellar/go/1.8.3/libexec/src/src/test.com/foo (from $GOROOT)
  /Users/abozanich/gopath/src/src/test.com/foo (from $GOPATH)
$ impl.new 'i *I' foo.I
func (i *I) Bar() {
  panic("not implemented")
}

$ cd ..
$ impl.new 'i *I' foo.I
unrecognized interface: foo.I
$ impl.new -dir $PWD/vim-go-impl 'i *I' foo.I
func (i *I) Bar() {
  panic("not implemented")
}

$ impl.orig -dir $PWD/vim-go-impl 'i *I' foo.I
impl <recv> <iface>

impl generates method stubs for recv to implement iface.

Examples:

impl 'f *File' io.Reader
impl Murmur hash.Hash

Don't forget the single quotes around the receiver type
to prevent shell globbing.
```
It runs another if it fails because I didn't want to break installs with an older `impl` build.  Omitting the `-dir <dir>` is usually ok for my workflow because I'm generally cd'd into the directory that has `vendor`.

Carpetsmoker: I think we can remove this workaround now that the `impl` PR is merged?

If you could rebase on master then I can test and merge it.

Carpetsmoker:
Could you tell me how to reproduce this issue? It's not really clear from this PR or the `impl` PR.

On master (without this patch) I tried using:

    :GoImpl l Log github.com/sirupsen/logrus.StdLogger

Which seems to work correct?

This package is vendored for the project I have open. I made sure it's *not* in my regular GOPATH.

What am I doing wrong?

boz:
@Carpetsmoker: impl defaults to `-dir $PWD`.  I wrote an example [here](https://github.com/fatih/vim-go/pull/1322#discussion_r120688380), take a look at what happens after the `cd ..`.

Carpetsmoker:
Ah right, didn't see that because it was collapsed :+1:

PR #1333, opened 2017-06-25
add echo message for GoCoverageToggle

812lcl:
I dont know if it has been running when I execute GoCoverageToggle, so add some echo message like GoTest.


fatih:
Thanks @812lcl 👍

Issue #963, opened 2016-07-26
Improvement suggestion: Syntax based folding for imports


rbjorklin:
Would it be a good idea to have syntax based folding for imports much like Intellij does it?

I think it could be achieved by adding the `fold` [word at line 193](https://github.com/fatih/vim-go/blob/master/syntax/go.vim#L192-L193).

fatih:
Thanks @rbjorklin.

Unfortunately I don't have much time for implementing this feature, nor do I know if this is worth implementing. I don't close this and mark this as feature if someone wants to tackle it. If no one comes in 1-2 months I'm going to close it. Thanks for the suggestion! :)

Carpetsmoker:
> I think it could be achieved by adding the fold word at line 193.

Well, that would also fold `var ( .. )` and `const ( .. )`. Some might consider that a feature, others might not :-)

At any rate, I did some work on more fine-grained folding settings today − I'll finish it up later.

fervic:
Hi @Carpetsmoker did you make any progress on this?  I think `var( ... )` and `const( ... )` blocks should be foldable as well.

Carpetsmoker:
Man, I can't believe it's been ten months... >_<

Either way, I created a PR @fervic @rbjorklin: #1339. Let me know if you have problems!

rbjorklin:
I haven't written any Go in quite some time now and had honestly completely forgot about this. Feel free to close this issue. Thanks for putting in some effort though @Carpetsmoker!

fervic:
Better late than never, I will try your PR out @Carpetsmoker. Thanks!

fatih:
Thanks to @Carpetsmoker we have now an initial support for syntax based folding. It's on master, please pull, use it and any feedback is weldome. Please add @Carpetsmoker to the issue when reporting (issues or feature requests) :)

PR #1339, opened 2017-06-30
Fold import, var, and const blocks

Carpetsmoker:
This adds `foldmethod=syntax` folding for `import`, `var`, and `const` blocks.

I added a new `g:go_fold_enable` setting to selectively enable which regions get
folded. I find this useful personally as I only want to fold `import` blocks.

Future extension to this might be:

- `struct` and `interface`, to selectively fold `type foo struct { .. }` and
  `type foo interface { .. }` definitions;

- `package`, to fold the package-level comment.

But that's another PR for another day ;-)

Fixes: #963

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1606,6 +1606,10 @@ Enable folding of only imports:
 >
   let g:go_fold_enable = ['import']
 <
+Disable everything (same as not setting 'foldmethod' to `syntax`):
+>
+  let g:go_fold_enable = []
+<
 
 ==============================================================================
 DEVELOPMENT                                               *go-development*
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -690,8 +690,8 @@ CTRL-t
       :GoAddTags xml db
 <
     If [option] is passed it'll either add a new tag with an option or will
-    modify existing tags. An example of adding `omitempty` to all `json` fields
-    would be:
+    modify existing tags. An example of adding `omitempty` to all `json`
+    fields would be:
 >
       :GoAddTags json,omitempty
 <
@@ -1565,7 +1565,6 @@ default it's 60 seconds. Must be in milliseconds.
 >
       let g:go_statusline_duration = 60000
 <
-
                                                    *'g:go_addtags_transform'*
 
 Sets the `transform` option for `gomodifytags` when using |:GoAddTags| or if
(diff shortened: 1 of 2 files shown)


doc/vim-go.txt
```diff
@@ -1589,6 +1588,23 @@ By default "snakecase" is used. Current values are: ["snakecase",
...
+- "varconst"    `var` and `const` blocks.
+
+By default it's 1.
+>
+  let g:go_fold = 1
+<
+Enable folding of only imports:
+>
+  let g:go_fold = ['import']
+<
```
fatih: I don't like changing variables to two different types. I would rather have two settings than one. One of enabling/disabling and one for changing the settings.

doc/vim-go.txt
```diff
@@ -1589,6 +1588,23 @@ By default "snakecase" is used. Current values are: ["snakecase",
 >
       let g:go_addtags_transform = 'snakecase'
 <
+                                                        *'g:go_fold_enable'*
```
fatih: I think we should add a line explaining that they get this when they also add `set foldmethod=syntax` to their vimrc

syntax/go.vim
```diff
@@ -89,12 +89,30 @@ if !exists("g:go_highlight_generate_tags")
...
+  endif
+endif
+
 syn case match
 
-syn keyword     goDirective         package import
-syn keyword     goDeclaration       var const
+syn keyword     goPackage           package
+syn keyword     goImport            import    contained
+syn keyword     goVarConst          var const contained
```
fatih: This changes the semantics of the syntax settings. In go const, var and type all are declarations, so `goDeclaration` is the correct  setting here.  Actually things like `import` or `func` are also declarations. I'm not sure why they were added as `goDirective` back then.  If you need to access them, I think the better way would be using `goVar` and `goConst`, instead of combining them. At least that would give it a correct meaning (even if it's more fine grained now)

syntax/go.vim
```diff
@@ -189,8 +207,31 @@ syn region      goCharacter         start=+'+ skip=+\\\\\|\\'+ end=+'+ contains=
...
+
+" var, const
+if s:fold_varconst
+  syn region    goVarConst          start='\(var\|const\) (' end=')' transparent fold contains=ALLBUT,goParen,goBlock
+else
+  syn region    goVarConst          start='\(var\|const\) (' end=')' transparent contains=ALLBUT,goParen,goBlock
+endif
+
+" Single-line var, const, and import.
+syn match       goSingleImport      /\(import\|var\|const\) [^(]\@=/ contains=goImport,goVarConst
```
fatih: goSingleImport -> goSingleDecl

syntax/go.vim
```diff
@@ -189,8 +207,31 @@ syn region      goCharacter         start=+'+ skip=+\\\\\|\\'+ end=+'+ contains=
...
+
+"syn region goImport start='var ' end=''contains=goVarConst
```
fatih: Should be removed probably?

doc/vim-go.txt
Carpetsmoker: Could also just make the default be `let g:go_fold = ['import', 'block', 'varconst']`? The `1` was just a shortcut for this.

I'm not sure if we need two variables, and it will confuse things if they're both set.

fatih: Ok that works. But can you add a piece of text on how to disable it (i.e: setting to empty list) so people know it.

PR #1335, opened 2017-06-27
Add g:go_doc_max_height setting

Carpetsmoker:
To control the maximum height of the window created by `:GoDoc` and `K`.

Also slightly tweak the documentation for `g:go_doc_keywordprg_enabled`, since
it doesn't actually use the `keywordprg` setting.

Fix some other small errors that [vimhelplint](https://github.com/machakann/vim-vimhelplint)
reported (wrong tag names, lines too long).

PR #987, opened 2016-08-01
Doc cleanup

Carpetsmoker:
- This merges @svanburen 's PR #935 − resolving the conflicts.

- Sync "troubleshooting" section with the one on the Wiki. IMHO we're probably best off deleting the wiki page after this is merged?

- Mention Vim's packages feature for installing.

- Run [vimhelplint](https://github.com/machakann/vim-vimhelplint) and fix all the reported errors. Many are boring line width changes, but there were also quite a few "missing link" errors (e.g. `|g:go_foo|` instead of `|'g:go_foo'|` was used).

- Add a few settings that were undocumented.

- Fix a few minor typos.

PR #1340, opened 2017-06-30
Add vimhelplint to Travis tests

Carpetsmoker:
There have been a few instances where this helped me catch some errors, for
example in #987 and #1335 where there were a few wrong tag names (e.g.
`|g:go_foo|` instead of `|'g:go_foo'|`).


fatih:
Thanks @Carpetsmoker 👍

Issue #1181, opened 2017-01-21
af text object for a literal function should also include the assignment


fatih:
For the given function below

<img width="275" alt="screen shot 2017-01-21 at 12 30 46 pm" src="https://cloud.githubusercontent.com/assets/438920/22174079/761237e8-dfd5-11e6-9962-484a6ca4b443.png">

If we do a `vaf`, the following is selected:

<img width="289" alt="screen shot 2017-01-21 at 12 30 15 pm" src="https://cloud.githubusercontent.com/assets/438920/22174082/7dd97aae-dfd5-11e6-9b23-2a7b8f769c5c.png">

It should select the following:

<img width="283" alt="screen shot 2017-01-21 at 12 30 30 pm" src="https://cloud.githubusercontent.com/assets/438920/22174083/83e4a144-dfd5-11e6-84b7-8b5f0c067f7e.png">

Using now this feature for couple of months since I've implemented, I think it should also select the assignment as well.

PR #1341, opened 2017-06-30
Also select variable when using af on anonymous functions

Carpetsmoker:
When using `af` on an anonymous function that is assigned, such as:

	peanuts := func(a string) error {
		return errors.New("I'm alergic!")
	}

we will now also select the `peanuts` variable.

This could be improved by not assuming a single space between `=` and the
`func`. Dunno if that's worth the effort ... use gofmt? ;-)

Fixes: #1181

### DIFF
--- a/autoload/go/textobj.vim
+++ b/autoload/go/textobj.vim
@@ -57,11 +57,22 @@ function! go#textobj#Function(mode) abort
 
   if a:mode == 'a'
     " anonymous functions doesn't have associated doc. Also check if the user
-    " want's to include doc comments for function declarations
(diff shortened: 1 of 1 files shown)


fatih:
Hi @Carpetsmoker 

This seems to be complicated. I've implemented a simpler version: https://github.com/fatih/vim-go/pull/1345 what do you think? This is configurable, and second it works by using the existing information (only for one liners do we need to calculate the column number of the variable).

Carpetsmoker:
Your solution is probably better @fatih. I'll close this. Thanks :+1:

PR #1344, opened 2017-07-05
Fix warning when goimports doesn't support srcdir

robot-dreams:


fatih:
Thanks @robot-dreams 👍

PR #1345, opened 2017-07-06
textobj: include function variable for anonymous functions

### LINKED ISSUE #1181: af text object for a literal function should also include the assignment

fatih:
fixes #1181

### DIFF
--- a/autoload/go/textobj.vim
+++ b/autoload/go/textobj.vim
@@ -6,6 +6,10 @@ if !exists("g:go_textobj_include_function_doc")
   let g:go_textobj_include_function_doc = 1
 endif
 
+if !exists("g:go_textobj_include_variable")
+  let g:go_textobj_include_variable = 1
+endif
+
 " ( ) motions
 " { } motions
 " s for sentence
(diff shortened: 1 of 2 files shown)


autoload/go/textobj.vim
```diff
@@ -60,6 +64,16 @@ function! go#textobj#Function(mode) abort
     " want's to include doc comments for function declarations
     if has_key(info, 'doc') && g:go_textobj_include_function_doc
       call cursor(info.doc.line, info.doc.col)
+    elseif info['sig']['name'] == '' && g:go_textobj_include_variable
```
Carpetsmoker: I think you meant `g:go_textobj_include_function_assignment` here, and not `g:go_textobj_include_variable`?

fatih: I made the changes but seems like I've forgot to push. Thanks for the reminder!

PR #1349, opened 2017-07-10
feat(docs): add vundle install to docs

ellerbrock:
# Vundle

- add vundle installation to docs

cheers maik

hurray


fatih:
I want to keep the readme small. Thanks for this though :)

PR #1352, opened 2017-07-11
Update Readme - Add vim-plug option to atomatically run GoInstallBinaries

fvosberg:
Hi, 

when you use Vim-Plug, you can configure it to install the go binaries after it installs or updates vim-go. This is a really handsome feature and I think, that others will appreciate this, too. 

What do you think? Are there any unintended side effects?


adelowo:
There shouldn't be any side effects I guess

fatih:
Hi @fvosberg  @adelowo 

There are actual. Some repos don't have `.git` folders for their packages under GOPATH as its get stripped when they vendor it. So in that case :GoInstallBinaries fail to work. That's the reason it's not recommended to do it actually.  Advanced users probably already know the `do` syntax so there is no need to add this to our readme. Thanks!

PR #1356, opened 2017-07-13
Clarify :helpt prerequisite for using :help vim-go

antoineco:
It took me a while to figure out the help tags had to be generated in order to use `:help vim-go`.
Making this more explicit might save some wandering souls a Google search :)

For most people `:helptags ALL` will do the right thing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,11 @@ Official documentation can be found under [doc/vim-go.txt](doc/vim-go.txt). You 
 ```
 :help vim-go
 ```
+
+Depending on your installation, you may have to generate the plugin's [help
+tags](https://github.com/vim/vim/blob/v8.0.0711/runtime/doc/helphelp.txt#L206-L227)
+manually (eg. `:helptags ALL`).
+
 We also have an [official vim-go
 tutorial](https://github.com/fatih/vim-go-tutorial).
 


Carpetsmoker:
I believe that plugin managers such as Vundle and Vim-plug already take care of this; you only need to do this when installing through Vim 8's package feature or pathogen.

Maybe it's better to mention it in the install section there?

antoineco:
I though the *Usage* section was a good place because this is where the user actually learns about the included Vim help. Maybe a note stating something like

> depending on your installation, you may have to generate the plugin's help tags *bla bla bla*

right after the example command would be better? Whatever feels clearer to you.

fatih:
Hi @antoineco 

> Maybe a note stating something like " depending on your installation, you may have to generate the plugin's help tags bla bla bla" right after the :help vim-go command block would be better? Whatever feels clearer to you.

I like this better. Can you update it please?

antoineco:
@fatih sure thing, I just amended my previous commit (and took the liberty of moving all links to the end of the markdown file).

fatih:
Hey @antoineco I appreciate the clarification but I don't like the shortcut syntax of markdown :/ I have to jump to the bottom know to understand for what the link stand for. Can you please just remove all add your single line of clarification without using any shortcut links?

antoineco:
Alright sorry, since the link I added is pretty I thought I could make the whole markdown a bit more readable. Reverting.

fatih:
Thanks @antoineco 👍

PR #1364, opened 2017-07-22
Don't add offset to goTypeConstructor.

### LINKED ISSUE #1357: Last character of type being colored incorrectly.
### Behavior

Last character of type being colored incorrectly.

### Steps to reproduce:

Try the following code in vim, with only `vim-go` and any color scheme. Look at the second last line.

```go
package main

type Video struct {
	ID  string `json:"id,omitempty"`
	URL string `json:"url,omitempty"`
}

func main() {
	v := &Video{}
}
```

### Configuration

```
call plug#begin('~/.config/nvim/plugged')
Plug 'fatih/vim-go'
call plug#end()

let g:go_auto_sameids = 1
let g:go_auto_type_info = 1
let g:go_fmt_command='goimports'
let g:go_highlight_build_constraints = 1
let g:go_highlight_extra_types = 1
let g:go_highlight_fields = 1
let g:go_highlight_methods = 1
let g:go_highlight_operators = 1
let g:go_highlight_structs = 1
let g:go_highlight_types = 1
```

### Screenshots

screen shot 2017-07-13 at 6 01 21 pm
screen shot 2017-07-13 at 6 05 36 pm

Carpetsmoker:
This is a regression from cfa9713. Previously the offset was needed since the
`{` was matched as well. Now we don't match the `{` the offset means the last
character of the type won't be matched.

Fixes #1357

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -321,7 +321,7 @@ hi def link    goField              Identifier
 
 " Structs & Interfaces;
 if g:go_highlight_types != 0
-  syn match goTypeConstructor      /\<\w\+{\@=/he=e-1 
+  syn match goTypeConstructor      /\<\w\+{\@=/
   syn match goTypeDecl             /\<type\>/ nextgroup=goTypeName skipwhite skipnl
   syn match goTypeName             /\w\+/ contained nextgroup=goDeclType skipwhite skipnl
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @Carpetsmoker 👍

PR #1367, opened 2017-07-22
fmt.vim: fix cursor staying on the same line when goimports is used

### LINKED ISSUE #1342: using goimports for go_fmt_command moves the cursor on save
### Behavior

Anytime I add or delete a line that requires adding or deleting an import, my cursor gets moved when I save the file (when adding a line, my cursor gets moved up one line from where I was; when I delete a line, the cursor gets moved down a line, because goimports has added or removed lines). It would be nice if my cursor stayed locked to the same line of code I was on already, before saving.

### Steps to reproduce:

1. make sure to configure vim with `let g:go_fmt_command = "goimports"`
2. open a new Go file (using the default template)
3. add a line after `fmt.Println("vim-go")` with the contents:  
`    user := os.Getenv("user")`
4. position cursor on just-inserted line, at the `u` in `user`
5. save the file with `:w`

### Configuration

* `vimrc` contains only `let g:go_fmt_command = "goimports"`
* vim version: NVIM v0.2.1-15-g08b23d08
* vim-go version: bcf3f23dd76f39f3c3b840e3f95772ab46312219
* go version: go1.8.2 darwin/amd64

fatih:
fixes #1342

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -70,6 +70,8 @@ function! go#fmt#Format(withGoimport) abort
   endif
 
   let out = go#fmt#run(bin_name, l:tmpname, expand('%'))
+  let diff_offset = len(readfile(l:tmpname)) - line('$') 
+
   if go#util#ShellError() == 0
     call go#fmt#update_file(l:tmpname, expand('%'))
   elseif g:go_fmt_fail_silently == 0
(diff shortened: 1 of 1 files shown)


autoload/go/fmt.vim
```diff
@@ -95,6 +97,9 @@ function! go#fmt#Format(withGoimport) abort
     " Restore our cursor/windows positions.
     call winrestview(l:curw)
   endif
+
+  " be smart and jump to the line the new statement was added/removed
+  call cursor(line('.') + diff_offset, 0)
```
clee: Ah! That would do it. I'm guessing the second argument to `cursor()` is the column number?

PR #1369, opened 2017-07-23
Preserve cursor column through formatting

clee:
Save it before calling the formatter, and use the column argument
when setting the position after.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -69,6 +69,7 @@ function! go#fmt#Format(withGoimport) abort
     let bin_name = "goimports"
   endif
 
+  let cursor_pos = getpos(".")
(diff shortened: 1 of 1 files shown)


autoload/go/fmt.vim
```diff
@@ -69,6 +69,7 @@ function! go#fmt#Format(withGoimport) abort
     let bin_name = "goimports"
   endif
 
+  let cursor_pos = getpos(".")
```
fatih: You can use `col(".")` which returns the column directly. Also I think it's better to change the variable to `let current_col`

fatih:
Btw according to the doc it says

```
		If {col} is zero, the cursor will stay in the current column.
```

that was the reason I didn't pass the column directly. But it seems it doesn't work well

fatih:
Thanks @clee 👍

PR #1372, opened 2017-07-24
Allow specifying which filestypes to list for :GoFiles

### LINKED ISSUE #1346: Whishlist: TestGoFiles, XTestGoFiles command wrappers
There is `:GoFiles` command which lists the go files. I wish there to be similar commands for listing test files only, something like:
`:GoTestFiles` for `go list -f '{{join .TestGoFiles "\n"}}'`
`:GoXTestFiles` for `go list -f '{{join .XTestGoFiles "\n"}}'`

Default output of such listing commands is not very useful and not very readable. Therefore I propose to integrate such listing commands with [fzf](https://github.com/junegunn/fzf.vim) plugin seamlessly. For example, if fzf is installed, override `:GoFiles` to something like following and if there is no fzf, fallback to default.
```VimL
command! FZFGoFiles call fzf#run(fzf#wrap({
    \ 'source': go#tool#Files(),
    \ 'options': '--preview "(highlight -O ansi {} || cat {}) 2> /dev/null | head -'.&lines.'"'
    \}))
```
(--preview option can be safely omitted)

Carpetsmoker:
Instead of always using `GoFiles`, you can now use `:GoFiles TestGoFiles` to
list all test files. Or `:GoFiles GoFiles TestGoFiles` to list both all regular
Go files and test files.

Fixes #1346

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -1,11 +1,42 @@
-function! go#tool#Files() abort
-  if go#util#IsWin()
-    let format = '{{range $f := .GoFiles}}{{$.Dir}}\{{$f}}{{printf \"\n\"}}{{end}}{{range $f := .CgoFiles}}{{$.Dir}}\{{$f}}{{printf \"\n\"}}{{end}}'
-  else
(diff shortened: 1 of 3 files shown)


fatih:
Nice @Carpetsmoker 👏

PR #1373, opened 2017-07-24
Document how to use a different Go binary

### LINKED ISSUE #1222: Feature Request: Configurable "go" binary name
Currently, vim-go does not allow you to configure the binary name for "go". This is frustrating when working with the Google App Engine SDK, since it replaces "go" with a binary "goapp" which has some additional commands and sets a different environment. Without using a workaround, it is extremely difficult to use vim-go effectively in this environment because vanilla "go" cannot find the appengine packages.

Carpetsmoker:
Tested my wrapper script from #1222 and confirmed this works fine.

IMHO it's not worth making an option for this, as:

- very few people need it;
- is quite a bit of effort to implement;
- will make the code less readable;
- an easy workaround exists.

Fixes #1222

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1716,16 +1716,8 @@ wrapper script; for example if you would like to run `goapp` instead of `go`:
 >
      PATH="$HOME/gobin/:$PATH" vim
 <
-   Or from within your vimrc:
->
-     augroup go_path
-       autocmd!
-       autocmd FileType go 
-         \  if stridx($PATH, expand("$HOME/gobin")) == -1
-         \|   let $PATH = expand("$HOME/gobin:$PATH")
-         \| endif
-     augroup end
-<
+   Alternatively you you could set `$PATH` in your vimrc with an |autocmd|.
+
 
 How do I use vim-go with syntastic?~
 
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1700,6 +1700,33 @@ example sometimes code.google.com times out. To test, just execute a simple
 You'll see a more detailed error. If this works, vim-go will work too.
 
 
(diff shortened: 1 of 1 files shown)


doc/vim-go.txt
```diff
@@ -1700,6 +1700,33 @@ example sometimes code.google.com times out. To test, just execute a simple
...
+>
+     augroup go_path
+       autocmd!
+       autocmd FileType go 
+         \  if stridx($PATH, expand("$HOME/gobin")) == -1
+         \|   let $PATH = expand("$HOME/gobin:$PATH")
+         \| endif
+     augroup end
+<
+
```
fatih: I think we don't need to explain how to use PATH's and co. I'm all for providing the wrapper and then tell the person to put it under PATH before the PATH entry where it contains `go`.  It's getting complex and someone who don't want to use `go` but something else probably already knows how to modify and fit their PATH.

Carpetsmoker: I'm afraid that I'm not sure that I understand what you're trying to say. Do you mean that we don't need this FAQ entry? Or that we don't need this `autocmd` "tip", or that we should just provide a brief description on how to fix this and don't need any of the code examples?

fatih: I'm ok for the FAQ entry, but we don't nee the `autocmd` tip :) The wrapper should be enough.

Carpetsmoker: Alright; I've removed the code snipper, but kept a single-line "hint" is that's okay?

fatih: Yeap totally ok, thanks!

fatih:
@Carpetsmoker seems like `vimhelplint` is not happy with this, but it doesn't show why it fails.

fatih:
Thanks @Carpetsmoker

PR #1374, opened 2017-07-24
Various fixes for paths and GOPATH with spaces

Carpetsmoker:
Fixes #1288, as well as a number of other commands.

There may be other places as well; this is just the result from a quick test
with my personal default settings. Since there are a lot of if/else branches
relating to jobs etc. it's kind of hard to test everything without a full audit.

It's perhaps a good idea to refactor some of this in a slightly more structured
way at some point: instead of passing strings we should pass a list which is
always escaped (like `os.Exec()`, Python's subprocess, etc.)

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -25,7 +25,8 @@ function! go#def#Jump(mode) abort
       let $GOPATH = old_gopath
       return
     endif
-    let command = printf("%s -f=%s -o=%s -t", bin_path, go#util#Shellescape(fname), go#util#OffsetCursor())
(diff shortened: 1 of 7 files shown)


fatih:
Thanks @Carpetsmoker much appreciated 👍 

> It's perhaps a good idea to refactor some of this in a slightly more structured
way at some point: instead of passing strings we should pass a list which is
always escaped (like os.Exec(), Python's subprocess, etc.)

We do it already for jobs, but not for old style system call. For those for some of the newly refactored code (Such as lint.vim (gometalinter) I'm converting the list to a string before passing it to `system()`. But it's a constant improvement so we can do these kind of things one by one  :)

PR #1377, opened 2017-07-24
Fold package-level comments

Carpetsmoker:
This enables folding of the `// Package foo` and `/* Package foo` comments at
the top of files.

Ideally I'd also like to fold the `package ...` line in here, but this is kind
of hard to do... Maybe I'll take a look at that in a future date. For now, this
will do :-)


fatih:
Thanks @Carpetsmoker 👍

PR #1379, opened 2017-07-24
Don't show window if "gogetdoc" command isn't found

Carpetsmoker:
When using `:GoGetDoc` without a `gogetdoc` binary it would still open the split
with the content `-1`. This fixes that.

An error message is not needed as `go#path#CheckBinPath` already shows that.

Related: #1346


fatih:
Thanks @Carpetsmoker 👍

PR #1381, opened 2017-07-25
Fix filepaths in the quickfix list for :GoVet

### LINKED ISSUE #1376: Using "vim subpkg/file.go" and :GoVet will clear the buffer when there are errors
1. `vim subpkg/file.go`
2. Make sure there is a `go vet` error in the file.
3. Run `:GoVet`
4. The error(s) will be correctly reported, but the buffer is now completely blank(!)
5. This doesn't happen with `cd subpkg; vim file.go`.
6. The buffer's filename has also changed from `subpkg/file.go` to `file.go`.

Carpetsmoker:
The `go vet` command would be run with `go#tool#ExecuteInDir()`, which
translates to `cd dir && go vet`.

The problem with this is that if I have `subpkg/file.go` open, the filename that
`go vet` reports will be `file.go`, rather than `subpkg/file.go`. As a result,
the filepath set in the quickfix or location list is also wrong and jumping to
this error will jump to a new (usually non-existing) buffer.

Fix it by explicitly passing the package path, rather than changing the cwd.

Fixes #1376

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -147,9 +147,9 @@ function! go#lint#Vet(bang, ...) abort
   call go#cmd#autowrite()
   echon "vim-go: " | echohl Identifier | echon "calling vet..." | echohl None
   if a:0 == 0
-    let out = go#tool#ExecuteInDir('go vet')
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @Carpetsmoker

PR #1385, opened 2017-07-25
Show message when using :GoDef and opening a new buffer

### LINKED ISSUE #1207: Unnecessary "vim-go: SUCCESS" status message after :GoDef
### Behavior

As of dc8073386754bf5ca09a8dbe2bd9cc8886c61446, calling `:GoDef` prints `vim-go: SUCCESS` in the status line after the definition is found. This isn't useful, and replaces the old useful behavior of printing the name of the file that was opened (when the definition is in another file).

### Steps to reproduce:

Call `:GoDef` or `gd` on any symbol, especially one defined in another file.

### Configuration

* vim-go version: dc8073386754bf5ca09a8dbe2bd9cc8886c61446..master
* vim version:

        VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Feb 16 2017 04:51:36)
        MacOS X (unix) version

Carpetsmoker:
Instead of:

	vim-go: SUCCESS

We now show:

	vim-go: /usr/lib/go/src/strings/strings.go:340:6: defined here as func strings.Join

Showing the output of the command is optional; if we remove it we get the old
behaviour with the output of an `:edit` command:

	"/usr/lib/go/src/strings/strings.go" [readonly] 777L, 20971C

Showing the command output seemed more useful to me personally though,

I moved the `custom_cb()` invocation after the message; this has no real effect
as far as I can see (it's used in only two places) and it allows us output
messages after the SUCCESS.

Fixes #1207

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -97,6 +97,7 @@ function! s:jump_to_declaration_cb(mode, bin_name, job, exit_status, data) abort
   endif
 
   call go#def#jump_to_declaration(a:data[0], a:mode, a:bin_name)
+  call go#util#EchoSuccess(a:data[0])
 endfunction
 
 function! go#def#jump_to_declaration(out, mode, bin_name) abort
--- a/autoload/go/job.vim
+++ b/autoload/go/job.vim
@@ -31,10 +31,6 @@ function go#job#Spawn(args)
   endfunction
 
   function cbs.exit_cb(job, exitval) dict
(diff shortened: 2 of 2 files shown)


fatih:
Seams a fair improvment. Merging it but gonna change the print to:

```
  call go#util#EchoSuccess(fnamemodify(a:data[0], ":t"))
```

as the file name might be very long .

PR #1386, opened 2017-07-25
Some small :GoImpl improvements

### LINKED ISSUE #1077: GoImpl creates method stubs inside struct definiton
When using `GoImpl` directly on a struct field the implementation is actually put inside the struct so I am going streight from

```
type MyStringer struct {
}
```

to 

```
type MyStringer struct {

func (m *MyStringer) String() string {
    panic("not implemented")
}

}
```

Carpetsmoker:
- When using `:GoImpl` on `type foo struct{}` it would work, but with:

        type foo struct{
        }

  or with a struct with fields, it would create the generated methods *inside*
  the struct. Fixes #1077.

- Allow using `:GoImpl` on the `type` and `struct` parts too. Makes it a wee
  bit easier to use ;-)

- Output would include extra newline, and error would include trailing newline
  from shell command:

        vim-go: invalid receiver: "} *}"<00>

- Small documentation fix.

### DIFF
--- a/autoload/go/impl.vim
+++ b/autoload/go/impl.vim
@@ -6,11 +6,13 @@ function! go#impl#Impl(...) abort
 
   let recv = ""
   let iface = ""
+  let interactive = 0
+
+  let pos = getpos('.')
 
   if a:0 == 0
-    " user didn't passed anything,  just called ':GoImpl'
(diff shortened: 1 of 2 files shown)


fatih:
Thanks @Carpetsmoker as always 👍

PR #1387, opened 2017-07-26
Allow comments in import blocks

Carpetsmoker:
Otherwise they won't get highlighted and folding won't work correctly if an
`import` block contains a comment.

e.g.:

2017-07-26-113553_508x91_scrot

----

2017-07-26-113721_623x239_scrot

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -218,9 +218,9 @@ endif
 
 " import
 if s:fold_import
-  syn region    goImport            start='import (' end=')' transparent fold contains=goImport,goString
+  syn region    goImport            start='import (' end=')' transparent fold contains=goImport,goString,goComment
 else
-  syn region    goImport            start='import (' end=')' transparent contains=goImport,goString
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
I think this should have been in 1.14 btw @fatih ... It fixes a reasonably visible bug in the new folding code.

fatih:
Seems like that slip away as I thought this was an addition (I've already merged your other fix). People usually use master anyway, but if you insist we can release a 1.14.1 with the fixes you think we should include :)

Carpetsmoker:
What's the point of making releases if people don't use them? ;-)

Either way, I'm fine with leaving it as-is; just wanted to remind you that it would be a good idea to merge this sooner rather than later ;-)

fatih:
Right, the thing is they just update their plugin via `PlugUpdate` or cd into it and do a `git pull origin master`. Releases are great to keep things, for debugging and I think some linux distros also package vim-go. So anyway, we have it if needed. Thanks for reminding 👍

PR #1388, opened 2017-07-26
[WIP] enable vim8 terminal

mattn:
DO NOT MERGE.

I'm working this currently. 

godebug

### DIFF
--- a/autoload/go/test.vim
+++ b/autoload/go/test.vim
@@ -38,7 +38,16 @@ function! go#test#Test(bang, compile, ...) abort
     call go#term#new(a:bang, ["go"] + args)
   endif
 
-  if go#util#has_job() || has('nvim')
+  if go#util#has_term()
+    " use nvims's job functionality
+    if get(g:, 'go_term_enabled', 0)
+      let id = go#term#new(a:bang, ["go"] + args)
+    else
+      let id = go#jobcontrol#Spawn(a:bang, "test", args)
+    endif
+
+    return id
+  elseif go#util#has_job() || has('nvim')
     " use vim's job functionality to call it asynchronously
     let job_options  = {
           \ 'bang': a:bang,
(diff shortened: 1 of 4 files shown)


mattn:
Now latest vim has `job` and `terminal` both. `:GoTest` should be run on terminal?

fatih:
@mattn definitely yes (as I believe it improves the UX/Usability immense and also fixes some of the complaints, such as to see the test output and not just the errors). But not many are using HEAD vim. Even most use still 7.4. If we do it, we should guard it with `has()` clauses  (maybe checking if the feature is there?) so it works for all. 

So users who use latest VIM will get Terminal in `:GoTest`, otherwise they will not get it. Both old and new Vim users still will see quickfix list with errors (if there is any).

Carpetsmoker:
`:help terminal.txt` says "WARNING: THIS IS ONLY PARTLY IMPLEMENTED, ANYTHING CAN STILL CHANGE". Looking through the `git log` of `terminal.c` not a while lot *has* actually changed in the last few months; not sure how experimental it *actually* is.

Carpetsmoker:
Are you going to continue working on this @mattn? Or what is the status? I don't mind potentially working on it, just wanting to know what your plans/ideas are. Thanks!

fatih:
ping @mattn This is great if we could work on it. Let us know @mattn what your plans are, would be helpful to merge this into vim-go.

autoload/go/test.vim
```diff
@@ -38,7 +38,16 @@ function! go#test#Test(bang, compile, ...) abort
     call go#term#new(a:bang, ["go"] + args)
   endif
 
-  if go#util#has_job() || has('nvim')
+  if go#util#has_term()
+    " use nvims's job functionality
+    if get(g:, 'go_term_enabled', 0)
+      let id = go#term#new(a:bang, ["go"] + args)
+    else
+      let id = go#jobcontrol#Spawn(a:bang, "test", args)
```
fatih: The test is failing because there is no more `jobcontrol.vim`. We have a single `job.vim` that takes care of Vim8 and Nvim related async operations.

bhcleek:
Closing. A competing implementation was done in #2250 2639

PR #1390, opened 2017-07-27
add GoDebug

mattn:
Hi, I added debugger. Do you have interesting this?

godebug

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,7 @@ This plugin adds Go language support for Vim, with the following main features:
 * Go to symbol/declaration with `:GoDef`.
 * Look up documentation with `:GoDoc` or `:GoDocBrowser`.
 * Easily import packages via `:GoImport`, remove them via `:GoDrop`.
-* Automatic `GOPATH` detection which works with `gb` and `godep`. Change or
-  display `GOPATH` with `:GoPath`.
+* Precise type-safe renaming of identifiers with `:GoRename`.
 * See which code is covered by tests with `:GoCoverage`.
 * Add or remove tags on struct fields with `:GoAddTags` and `:GoRemoveTags`.
 * Call `gometalinter` with `:GoMetaLinter` to invoke all possible linters
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ This plugin adds Go language support for Vim, with the following main features:
   with `:GoTest`. Run a single tests with `:GoTestFunc`).
 * Quickly execute your current file(s) with `:GoRun`.
 * Improved syntax highlighting and folding.
+* Debug programs with integrated `delve` support with `:GoDebugStart`.
 * Completion support via `gocode`.
 * `gofmt` or `goimports` on save keeps the cursor position and undo history.
 * Go to symbol/declaration with `:GoDef`.
@@ -28,7 +29,6 @@ This plugin adds Go language support for Vim, with the following main features:
   errors, or make sure errors are checked with `:GoErrCheck`.
 * Advanced source analysis tools utilizing `guru`, such as `:GoImplements`,
   `:GoCallees`, and `:GoReferrers`.
-* Precise type-safe renaming of identifiers with `:GoRename`.
 * ... and many more! Please see [doc/vim-go.txt](doc/vim-go.txt) for more
   information.
 
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -180,7 +180,12 @@ function! go#debug#Stop() abort
 
   call s:stop()
 
-  wincmd p
+  let bufs = filter(map(range(1, winnr('$')), '[v:val,bufname(winbufnr(v:val))]'), 'v:val[1]=~"\.go$"')
+  if len(bufs) > 0
+    exe bufs[0][0] 'wincmd w'
+  else
+    wincmd p
+  endif
   silent! exe bufwinnr(bufnr('__GODEBUG_STACKTRACE__')) 'wincmd c'
   silent! exe bufwinnr(bufnr('__GODEBUG_VARIABLES__')) 'wincmd c'
   silent! exe bufwinnr(bufnr('__GODEBUG_OUTPUT__')) 'wincmd c'
@@ -319,14 +324,14 @@ function! s:starting(ch, msg) abort
   endif
 endfunction
 
-function! go#debug#Start() abort
+function! go#debug#Start(...) abort
   if has_key(s:state, 'job') && job_status(s:state['job']) == 'run'
     return
   endif
   try
     echohl SpecialKey | echomsg 'Starting GoDebug...' | echohl None
     let s:state['message'] = []
-    let job = job_start('dlv debug --headless --api-version=2 --log --listen=' . s:addr . ' --accept-multiclient', {
+    let job = job_start('dlv debug --headless --api-version=2 --log --listen=' . s:addr . ' --accept-multiclient -- ' . join(a:000, ' '), {
     \ 'out_cb': function('s:starting'),
     \ 'err_cb': function('s:starting'),
     \ 'exit_cb': function('s:exit'),
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -29,6 +29,7 @@ function! s:exit(job, status) abort
   if has_key(s:state, 'job')
     call remove(s:state, 'job')
   endif
+  call s:clearState()
   if a:status != 0
     echohl Error | echo join(s:state['message'], "\n") . "\n" | echohl None
 	call getchar()
@@ -161,7 +162,6 @@ function! s:show_variables() abort
 endfunction
 
 function! s:clearState() abort
-  let s:state['breakpoint'] = {}
   let s:state['currentThread'] = {}
   let s:state['localVars'] = {}
   let s:state['functionArgs'] = {}
@@ -508,7 +508,9 @@ function! s:stack_cb(ch, json) abort
   let s:stack_name = ''
   let res = json_decode(a:json)
   if type(res) == v:t_dict && has_key(res, 'error') && !empty(res.error)
-    echohl Error | res.error | echohl None
+    echohl Error | echomsg res.error | echohl None
+    call s:clearState()
+    call go#debug#Restart()
     return
   endif
   if empty(res) || !has_key(res, 'result')
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -10,6 +10,8 @@ if !exists('s:state')
   \}
 endif
 
+let s:addr = '127.0.0.1:8181'
+
 function! s:groutineID()
   return s:state['currentThread'].goroutineID
 endfunction
@@ -197,7 +199,7 @@ function! s:start_cb(ch, json)
   setlocal filetype=godebug-output
   nmap <buffer> q <Plug>(go-debug-stop)
 
-  silent leftabove 20vnew
+  silent leftabove 30vnew
   silent file `='__GODEBUG_VARIABLES__'`
   setlocal buftype=nofile bufhidden=wipe nobuflisted noswapfile nowrap nonumber nocursorline
   setlocal filetype=godebug-variables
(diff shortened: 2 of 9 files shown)


mattn:
Maybe, there are not enough commands, features, behaviors.

Carpetsmoker:
Previous PR: https://github.com/fatih/vim-go/pull/1198

mattn:
status update
terminal

fatih:
@mattn so happy that you working on this. I'm on road to NYC, but hopefully gonna take a look more in detail. 

I think this is possible now that vim-8 has term support. Do you know which patch version should be cap it? I think we should add something like this:

```
  " we use XXXtermXXX() which was introduce with 8.0.XXXXX, be sure we have
  " it: http://ftp.vim.org/vim/patches/8.0/8.0.XXX
  if !exists("*XXXtermXXX")
    call go#util#EchoError("GoDebugStart requires 'XXXtermXXX'. Update your Vim/Neovim version.")
    return
  endif
```

or better if you have the patch version that would be better. I think people should use an updated Vim version otherwise we have to deal with problems that are already fixed on HEAD in vim-dev.

mattn:
Ultimately, I did not use vim8 terminal feature for this debugger. So this should work on current vim8. And one more point is possible to port to neovim easily.

Carpetsmoker:
I haven't had time to look at this before, but I just tested it and this is
great stuff!

I made a PR on your fork with a bunch of minor changes; see the commit messages
for details on that. I also did some work on improving and expanding the
documentation a bit; I'll finish that up later.

----

Some further comments (Vim 8.0.1066, Go 1.9):

- The string `v := "hello"` is displayed as `v: [5]`, and expanding it gives me
  the byte values:

		v[0]: 104
		v[1]: 101
		v[2]: 108
		v[3]: 108
		v[4]: 111

  Is this intentional?

- Small thing, but you can't close expanded variables. I would expect a second
  `<CR>` to close it. I tried adding this in `s:expand_var()` but had some
  trouble grokking that function at a glance.

- Both the the struct:
  
		s := struct {
			a, b string
			u    int64
		}{"aasd", "zxc", 42}

  and:

		type named struct {
			a, b string
			u    int64
		}
		n := named{"zxc", "qwe", 42}

  Give an error on expanding:

		expression "s" (struct { main.a string; main.b string; main.u int64 }) does not support indexing
		expression "n" (main.named) does not support indexing

  Expanding slices seems to work fine.

Carpetsmoker:
Did some more testing, and found a few more small problems:

- String types now show up as `v: ""` with this example (breaking on the `fmt` line):

      v := "hello"
      fmt.Println("one", v)

- Using `vim dbg/debug.go` gives the error:

      2017/09/10 04:49:57 debugger.go:97: launching process with args: [/home/martin/go/src/a/debug ]
      Can not debug non-main package

   I have to use `:cd dbg` and then `:GoDebugStart` works.

- `:GoDebugRestart` doesn't seem to work. I added a breakpoint and the program is paused. I changed a variable declaration line (e.g. from `i := 42` to `i := 43`), and use `:GoDebugRestart`. I would expect the debugger to restart the program with the new source, but `i` still shows us as `42` instead of the expected `43`. I need to use `:GoDebugStop` and `:GoDebugStart` to get the new version.

   Also, the debugger doesn't restart with the new source if I restart the program with F5 after it has finished. I'm not sure if it *should* though!

- Pressing F5 after the program has exited messes up the window layout:
  2017-09-10-045658_906x532_scrot

  `:messages` has:

      API server listening at: 127.0.0.1:8181
      Process 19284 has exited with status 0
      Process 19284 has exited with status 0
      Process 19284 has exited with status 0

- The current directory now has a `debug` binary in it, presumably from the `debug.go` filename. This is because it's currently using `dlv debug ...`. Maybe it would be better to `:GoInstall` the binary and then use `dlv exec`? Or run it from a temp directory (like `go run`)?
  It's not a huge deal, but personally I'm not a huge fan of commands that produce unexpected side-effects like this on the filesystem.

- Come to think of it, maybe we want to have a way for the user to add flags to `dlv`? Especially `--build-flags` would be useful for some people I think. We cab also do this later maybe, if someone asks for it ("[YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it)")

autoload/go/debug.vim
```diff
@@ -381,6 +381,7 @@ function! go#debug#Start(...) abort
   try
     echohl SpecialKey | echomsg 'Starting GoDebug...' | echohl None
     let s:state['message'] = []
+    exe 'lcd' fnamemodify(bufname(''), ':p:h')
```
Carpetsmoker: Won't this cause the current working directory to change? `:pwd` gives me `/home/martin/go/src/a` when starting `vim dbg/debug.go`, but it's `/home/martin/go/src/a/debug` after running `:GoDebugStart`.

Also, I get an error using `:GoDebugStart` with the latest changes?

```
can't load package: package ['--']: cannot find package "['--']" in any of:
        /usr/lib/go/src/['--'] (from $GOROOT)
        /home/martin/go/src/['--'] (from $GOPATH)
        /home/martin/work/src/['--']
exit status 1
```

Maybe it's worth adding some tests for this? I never really looked in to testing Vimscript stuff so I don't know how hard that is.

mattn: fixed for `["--"]`.

`lcd` change current directory for current window.

Carpetsmoker:
What are your plans for supporting Neovim btw @mattn? It currently errors out with `Vim(let):E117: Unknown function: job_start`.

mattn:
Go debugger uses JSON-RPC to communicate with dlv debugger. neovim support TCP connection?

Carpetsmoker:
I don't know; never looked in to Neovim's implementation in detail.

Personally I'm fine with not adding support for that in this PR if it's hard, but then we should add a clear error message for that.

fatih:
We should not add NeoVim now. First, let finish making the UX and usability right. And then fix bugs and co. once it's in master. Neovim can wait as Vim-go aims to be a Vim first class plugin rather than supporting both nvim and vim.

Carpetsmoker:
This PR currently depends [on this delve PR](https://github.com/derekparker/delve/pull/1028), but after that's merged I think this PR is ready to be merged @mattn @fatih @bhcleek. Testing or feedback welcome!

There are some areas that might need some further improving – there always are – but overall it works well and I don't expect any further major changes to the `:GoDebug` commands.

It's currently a bit slow for larger packages because `dlv` builds the binary with `-a`, skipping the cache. This is so that it can add the `-gcflags "-N -l"` compile flags (the cached packages may not have that). Luckily this will be fixed for us in Go 1.10 which has the ability to keep different caches for different build flags.

Aside from the `:GoDebug*` commands this also adds a `g:go_debug` setting to debug vim-go itself. See [this commit](https://github.com/fatih/vim-go/pull/1390/commits/2a1822e453cdc5de52cfce913e28a3d59c546d46). Other than this, all changes are limited to `:GoDebug*` commands.

See `:help go-debug` for documentation.

fatih:
I want to test it before we merge it. Please don't merge it until then. I appreciate all the hard work btw, I think this is going to be one of the biggest addition in 2018 :)

README.md
```diff
@@ -28,7 +29,6 @@ This plugin adds Go language support for Vim, with the following main features:
   errors, or make sure errors are checked with `:GoErrCheck`.
 * Advanced source analysis tools utilizing `guru`, such as `:GoImplements`,
   `:GoCallees`, and `:GoReferrers`.
-* Precise type-safe renaming of identifiers with `:GoRename`.
```
fatih: Is this removed accidentally?

doc/vim-go.txt
```diff
@@ -1815,6 +1829,200 @@ The `gohtmltmpl` filetype is automatically set for `*.tmpl` files; the
...
+variable name to expand the values.
+
+When you're done use |:GoDebugStop| to close the debugging windows and halt
+the `dlv` process, or |:GoDebugRestart| to recompile the code.
+
+                                                           *go-debug-commands*
+DEBUGGER COMMANDS~
+
+Only |:GoDebugStart| and |:GoDebugBreakpoint| are available by default; the
+rest of the commands become available after starting debug mode.
```
fatih: `commands and mappings...` Worth noting that we also create default mappings

doc/vim-go.txt
```diff
@@ -1815,6 +1829,200 @@ The `gohtmltmpl` filetype is automatically set for `*.tmpl` files; the
...
+
+You can add breakpoints with |:GoDebugBreakpoint| (<F9>) and run your program
+with |:GoDebugContinue| (<F5>).
+
+The program will halt on the breakpoint, at which point you can inspect the
+program state. You can go to the next line with |:GoDebugNext| (<F10>) or step
+in with |:GoDebugStep| (<F11>).
+
+Struct values are displayed as `{...}`, array/slices as `[4]`. Use <CR> on the
+variable name to expand the values.
```
fatih: This is great, but I didn't know I had to go to the `GODEBUG_VARIABLES` window. Worth adding this little piece. Actually we should document more each window. We just mentioned them briefly above.

doc/vim-go.txt
```diff
@@ -1815,6 +1829,200 @@ The `gohtmltmpl` filetype is automatically set for `*.tmpl` files; the
...
+
+When you're done use |:GoDebugStop| to close the debugging windows and halt
+the `dlv` process, or |:GoDebugRestart| to recompile the code.
```
fatih: How can I clear previously set breakpoints? Even though I've stopped or restarted the server, the breakpoints are still intact. Is this something we can provide to the user? Seems like there is no way to remove all breakpoints right now and you have to manually remove them one by one.

fatih:
Thanks a lot for the hard work. There are couple of things that I think needs clarification, otherwise it looks good.

doc/vim-go.txt
fatih: Oh it seems like restarting Vim does remove all breakpoints

README.md
Carpetsmoker: If I remember correctly, at some point it was decided to keep the top 10 or 15 points from the help file here instead of all. So I removed the bottom one after I added the debug feature.

fatih: Can we remove the `automatic gopath` detection line and keep this? I think the GOPATH one can change as it's less relevant nowadays.

doc/vim-go.txt
Carpetsmoker: We could add a `:GoDebugReset` command maybe?

fatih: Yeap definitely, I think this is much needed otherwise one has to close and restart Vim (which is not something we want to do). Also another thing that I've just seen is, when you change or remove the lines, the breakpoint lines will be invalid. Not %100 related to this, but worth adding as well.

Carpetsmoker: > when you change or remove the lines, the breakpoint lines will be invalid. Not %100 related to this, but worth adding as well.

How do other debuggers handle this? Move the breakpoint, or just clear it?

doc/vim-go.txt
Carpetsmoker: Okay, I rephrased it a bit and added some more info.

README.md
Carpetsmoker: Sure, sounds good

doc/vim-go.txt
fatih: In VSCode it moves when you add newlines. So it's dynamic to the content of the line. But not sure if it's worth adding that kind of complexity. But one thing I appreciate is the way to disable/enable breakpoints. This allows to temporary run the whole stack without getting blocked. And they also have something to reset all the breakpoints.  So I think we need the following commands:

```
:GoDebugToggleBreakpoint
:GoDebugDisableBreakpoints
:GoDebugEnableBreakpoints
:GoDebugRemoveBreakpoints
```

O think `:GoDebugReset` would be confusing with `:GoDebugRestart`. So don't let us use `:GoDebugReset`, instead let's have these dedicated commands. We can add `:GoDisableBreakpoints` and `:GoEnableBreakpoints` later if it's to much work. Wdyt?

(Here is the Debug menu from VSCode)

screen shot 2018-02-27 at 6 08 20 pm

Carpetsmoker: I'm not so sure of adding four new commands, I find it more annoying to use as e.g. `:GoDebug<Tab>` now has many more commands that are pretty much the same.

I think a better way is to use e.g.:

```
:GoDebugBreakpoint 42         " Toggle breakpoint for a line (add/remove)
:GoDebugBreakpoint toggle     " Toggle all
:GoDebugBreakpoint disable    " Disable all
:GoDebugBreakpoint enable     " Enable all 
:GoDebugBreakpoint remove     " Remove all
```

That being said, I think we should make an issue for this after merging this. It definitely should be done one way or the other, but I don't think it's a critical feature to add *right now*.

Same applies to the other issue with the wrong line. It's definitely something that needs to be improved, but not necessarily in this PR. It's already a large PR, and tracking all comments on GitHub is kind of hard.

fatih: Agree with that. Let's talk about the UI changes later (command and co).

fatih:
@mattn @Carpetsmoker lgtm. This is ready to be merged. let's merge this and then open new PR's upon feedback and things we see later (such as the breakpoint improvements). Thanks again for all the hard work, much appreciated.

mattn:
Great. Thank you!

PR #1391, opened 2017-07-27
Add support for unite.vim about :GoDecls / :GoDeclsDir

delphinus:
[unite.vim][] is a powerful "listing & selecting" plugin that looks like ctrlp.vim.  I want to use `:GoDeclsDir` feature with unite.vim and have implemented it.

[unite.vim]: https://github.com/Shougo/unite.vim

It has almost like `:GoDeclsDir / :GoDecls`, but differs at some points below.

* `:Unite decls` has features for both `:GoDeclsDir` and `:GoDecls`, but the main target is `:GoDeclsDir` because `unite.vim` has already `:Unite outline` that is similar to `:GoDecls`.
* `:Unite decls` does not conflict with the existent `:GoDecls / :GoDeclsDir`, so both of them can be used if both unite.vim and ctrp.vim are installed.

The document about this on `vim-go.txt` is the one I am worrying about.  I placed it at the next of `:GoDeclsDir`, but does anyone have an idea to place somewhere better?

| `:Unite decls` |
|---|
|  |

| `:Unite decls:appengine.go` |
|---|
|  |

| FYI: existing function from unite.vim `:Unite outline` |
|---|
|  |

### DIFF
--- a/autoload/unite/sources/decls.vim
+++ b/autoload/unite/sources/decls.vim
@@ -1,0 +1,63 @@
+let s:save_cpo = &cpoptions
+set cpoptions&vim
+
+let s:source = {
+      \ 'name': 'decls',
+      \ 'description': 'GoDecls implementation for unite',
+      \ 'syntax': 'uniteSource__Decls',
+      \ 'action_table': {},
+      \ 'hooks': {},
+      \ }
+
+function! unite#sources#decls#define()
+  return s:source
+endfunction
+
+function! s:source.gather_candidates(args, context) abort
+  let l:bin_path = go#path#CheckBinPath('motion')
+  if empty(l:bin_path)
+    return []
+  endif
+
+  let l:path = expand(get(a:args, 0, '%:p:h'))
+  if isdirectory(l:path)
+    let l:mode = 'dir'
+  elseif filereadable(l:path)
+    let l:mode = 'file'
+  else
+    return []
+  endif
+
+  let l:include = get(g:, 'go_decls_includes', 'func,type')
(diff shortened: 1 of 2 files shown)


delphinus:
Fixed docs & pass the tests

fatih:
Thanks @delphinus I need to install unite.vim and see how it behaves and works. One question though, while quick skimming the code, I wonder what happens if `motion` returns an error. For example introduce a syntax error and call Unite. What happens? I think Unite takes care of it?

delphinus:
Ah, good point.  I see many unite addon have error handling from source command output.  I will add it.

delphinus:
done.  If `motion` returned errors, unite handles them.

autoload/unite/sources/decls.vim
```diff
@@ -0,0 +1,68 @@
...
+  if isdirectory(l:path)
+    let l:mode = 'dir'
+  elseif filereadable(l:path)
+    let l:mode = 'file'
+  else
+    return []
+  endif
+
+  let l:include = get(g:, 'go_decls_includes', 'func,type')
+  let l:command = printf('%s -format vim -mode decls -include %s -%s %s', l:bin_path, l:include, l:mode, l:path)
```
Carpetsmoker: You should `shellescape()` most or all the variables you're printf-ing in the command string; otherwise stuff will break if the path contains spaces or special shell characters (this is uncommon in Go, but not illegal either, and especially an issue on systems like Windows with stuff `Application Data` in the path).

For example, if I use `:Unite decls` on `~/go/src/a/space path/space name.go` I get `[unite.vim] [decls] command returned invalid response.`.

delphinus: Thx for reviewing 👍 I fixed it.

PR #1398, opened 2017-08-04
silent gofmt when autosave

tomwei7:
Hi, when i save file which has syntax error, the autosave fmt work like this:
vim-go-before
it will break  vim window so i think set it silent will be better

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -165,7 +165,7 @@ endfunction
 function! s:fmt_autosave()
   " Go code formatting on save
   if get(g:, "go_fmt_autosave", 1)
-    call go#fmt#Format(-1)
+    silent! call go#fmt#Format(-1)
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
I'm not sure ... it seems like a good idea to show a clear error message indicating there was an error.

You can use the [`'shortmess'`](http://vimhelp.appspot.com/options.txt.html#%27shortmess%27) setting to control some of the "Press enter to continue" prompts (although I *don't* get this prompt with the default).

tomwei7:
my mistake, it seem just on macvim

fatih:
Yeah. It's the most used commant, don't want to shallow the error for no reason. Also as Martin said you can avoid the prompt by changing the shortmess settings. Thanks for the feedback anyway.

PR #1400, opened 2017-08-04
Issue 1396 go#def#tab fails if file is already open

cfsalguero:
Setting switchbuf=usetab only is not enough since switchbuf only works for buffer related commands.
If mode is tab, I made it to use sbuf as cmd, instead of edit since edit will ignore the value of switbuf


Carpetsmoker:
I've also noticed that this is broken, but haven't looked into a fix yet. This seems to work well in my testing.

👍

fatih:
Thanks for the fix!

PR #1402, opened 2017-08-09
Fix GoMetaLinter file paths in quicklist window

jeromenerf:
### Behavior

Similar to #1376  

When using neovim, `GoMetaFilter` shows a stripped down path for files in sub packages in the quicklist window.

### Steps to reproduce:

1. nvim subpkg/file.go
2. Make sure there is a go error in the file.
3. Run :GoMetaLinter
4. The error(s) will be correctly reported, but the buffer is now completely blank(!)
5. This doesn't happen with cd subpkg; vim file.go.
6. The buffer's filename has also changed from subpkg/file.go to file.go.

For some reason, the problem doesn't exist in vim8.

Using the same approach as https://github.com/fatih/vim-go/pull/1381 seems to work: 

```
"  let out = go#tool#ExecuteInDir(meta_command)
  let out = go#util#System(meta_command . ' ' . go#util#Shellescape(go#package#ImportPath()))
```

I have seen some commits regarding `cwd` recently which might be related.

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -87,7 +87,7 @@ function! go#lint#Gometa(autosave, ...) abort
 
   let meta_command = join(cmd, " ")
 
-  let out = go#tool#ExecuteInDir(meta_command)
+  let out = go#util#System(meta_command . ' ' . go#util#Shellescape(go#package#ImportPath()))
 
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
Thanks! This is now closed with #1414.

PR #1405, opened 2017-08-11
Add support fot golang/dep

ricardoseriani:


ricardoseriani:
I'm not sure how can I add a option to use the command DepEnsure and passes multiple arguments, so I add multiple DepEnsure commands.
When I find a way I'll add another pr fixing it.

Carpetsmoker:
What would be the advantage over using `:GoDepEnsure ...` vs. `:!dep ...`? As far as I can see it doesn't really integrate with Vim (as e.g. `:GoBuild`, `:GoTest`, etc. do)?

I'm not sure if adding simple wrappers for Go commandline tools is really necessary to be honest, not unless vim-go really adds some value over just using the command-line tools on their own. Otherwise there will be a *a lot* of `:Go*` commands.

Just my 2c...

> I'm not sure how can I add a option to use the command DepEnsure and passes multiple arguments, so I add multiple DepEnsure commands.

You can use `:command! -nargs=*`, which you already use. Doing the logic and custom completion is some work, and would more or less be duplicating the `dep` code though (see above). You can use `-complete=custom` or `-complete=customlist` for the tab completion (some examples [here](https://github.com/fatih/vim-go/blob/master/ftplugin/go/commands.vim)).

ricardoseriani:
You're right, the only thing that it adds is that when you import a package you're not currently using (by using GoDepEnsureAdd) it will auto-import it, so you can already do auto-completions using vim.
I just do this PS because the golang/dep its trying to be the "official" way to import things in vendor, so I think that it maybe will be useful.

ricardoseriani:
I'm trying to make this PR take more advantage of vim, but I'm having a trouble.
The GoImport can't find any packages in my vendor folder, is that ok ?

fatih:
Thanks @ricardoseriani but `dep` is not a tool to be invoked via the editor. `:GoBuild` for example doesn't create any binary, that's by design as I don't want people to create their deployment artifacts as part of their editors. Also `dep` is till an ongoing project. Let's revisit later (if there is anything to revisit). Thanks!

Issue #1389, opened 2017-07-26
quickfix list clobbered on save


nyarly:
### Behavior

Having just updated vim-go, I've noticed that when I save codefiles, my quickfix window closes and the quickfix list is emptied. When correcting e.g. compile errors, this is irritating because 

### Steps to reproduce:

Please create a reproducible case of your problem. If this step is
not provided, the issue will be **closed**

Re produce it with a minimal `vimrc` with all plugins disabled and
only `vim-go` enabled:

1. open a Go project
2. :make
3. :cw
4. < quickfix window opens with errors >
5. :update
6 < quickfix window closes, :cw won't reopen >

### Configuration

Add here your current configuration and additional information that might be
useful, such as:

This is vim-go 76cd99db6a88e825f361df0043cbff777c4a14fb, under Neovim 0.1.7, and Go 1.7.4.

EddieFloresLive:
I had the same issue and had to roll back to go-vim v1.12 (26362e4bfcd7601da5b8fcf0041744962766ef0c). Now the quickfix window is being displayed with build errors again.

josephholsten:
Looks like this was introduced by https://github.com/fatih/vim-go/commit/e4e992a1cab7f707f1b3447157795885182235d1

Fortunately, it documents a workaround: 
```vim
    let g:go_list_type = "locationlist"
```

josephholsten:
I think https://stackoverflow.com/questions/20933836/what-is-the-difference-between-location-list-and-quickfix-list-in-vim sums up why `locationlist` is more appropriate anyway. TL;DR:

>  More generally: The quickfix list is best when your search or compile involves multiple files, the location list best when only a single file is involved. – Trebor Rude Aug 29 '14 at 17:40

josephholsten:
@fatih can we revert https://github.com/fatih/vim-go/commit/e4e992a1cab7f707f1b3447157795885182235d1 in light of the above?

fatih:
Not sure how to fix it. Quickfix is global. So is Locationlist as well. What if you use locationlist for both save and format as well? In that case this will still happen. 

This can be fixed by disabling the following line in `ftm.vim`

```
  " this closes the window if there are no errors or it opens
  " it if there is any
  call go#list#Window(l:listtype, len(a:errors))
```

luckily we have a new setting that prevents autoclosing it. So if you add the following you should be good to go:

```
let g:go_list_autoclose = 0
```

[...]

PR #1406, opened 2017-08-12
not execute cleaning up list when gofmt success with fail_silently option

hori-ryota:
[quickfix list clobbered on save · Issue #1389 · fatih/vim-go](https://github.com/fatih/vim-go/issues/1389)

I don't want to clean up quickfix when autosave.

As a solution, if you do not display an error with gofmt, why not stop clean up?


fatih:
Fixed with another solution. Thanks @hori-ryota anyway 👍

PR #1411, opened 2017-08-16
term: exit insert mode

kirstein:
When triggering commands in term mode we enter insert mode. This is really frustrating. This fixes this behaviour and quits insert mode after entering it.


fatih:
Thanks @kirstein 👍

PR #1414, opened 2017-08-18
Fix to fails of GoMetalinter result filepath in subdir. resolve #1413

### LINKED ISSUE #1413: GoMetalinter fails to parse file paths in subdir
### Behavior

When execute GoMetalinter in subdir, that is failed to parse file paths

### Steps to reproduce:

1. create subdir in repo (e.g. foo/)
2. create go source file (e.g. foo/bar.go)
3. execute :GoMetalinter 
4. parse failed (e.g. need to `foo/bar.go` but `bar.go` )

### Configuration

* vim version: NVIM v0.2.0
* vim-go version: df99b193e4469ecf3ad38ffd81374cc3572381b5
* go version: go1.8.3 darwin/amd64

I'll create pull-request.

hori-ryota:
Execute `:GoMetalinter` in subdir, failed to parse filepath.

e.g.

Execute to `foo/bar.go`.
Now execute gometalinter in `foo/` dir.
So, result path will `bar.go` but want `foo/bar.go`

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -87,7 +87,7 @@ function! go#lint#Gometa(autosave, ...) abort
 
   let meta_command = join(cmd, " ")
 
-  let out = go#tool#ExecuteInDir(meta_command)
+  let out = go#util#System(meta_command)
 
   let l:listtype = "quickfix"
   if go#util#ShellError() == 0
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -60,7 +60,7 @@ function! go#lint#Gometa(autosave, ...) abort
   " For async mode (s:lint_job), we want to override the default deadline only
   " if we have a deadline configured.
   "
-  " For sync mode (go#tool#ExecuteInDir), always explicitly pass the 5 seconds
+  " For sync mode (go#util#System), always explicitly pass the 5 seconds
   " deadline if there is no other deadline configured. If a deadline is
   " configured, then use it.
 
--- a/tags
+++ b/tags
@@ -1,0 +1,529 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
+!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
+!_TAG_PROGRAM_AUTHOR	Darren Hiebert	/dhiebert@users.sourceforge.net/
+!_TAG_PROGRAM_NAME	Exuberant Ctags	//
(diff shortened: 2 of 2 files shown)


fatih:
I think Martin changed this so it's consistent with others, adding him so he can chime as well. /cc @Carpetsmoker

Carpetsmoker:
I tested this, and both `go#tool#ExecuteInDir()` and `go#util#System()` seem to work fine for me. I made sure to comment out the async part, and tested both inside outside of GOPATH.

This is because the command that gets run is:

    /home/martin/go/bin/gometalinter --disable-all --enable=vet --enable=golint --enable=errcheck /home/martin/go/src/a/b --deadline=5s

Which will report back with the full path in the output.

I'm not sure why it's failing for you, but `ExecuteInDir()` is at best an unnecessary abstraction here, so removing it sounds like a good idea :+1:

> I think Martin changed this so it's consistent with others, adding him so he can chime as well. /cc @Carpetsmoker

Nope, didn't touch this :-) I did make a similar fix in #1381 though.

autoload/go/lint.vim
```diff
@@ -87,7 +87,7 @@ function! go#lint#Gometa(autosave, ...) abort
 
   let meta_command = join(cmd, " ")
 
-  let out = go#tool#ExecuteInDir(meta_command)
+  let out = go#util#System(meta_command)
```
Carpetsmoker: The function comment says:

   " For sync mode (go#tool#ExecuteInDir), always explicitly pass the 5 seconds
   " deadline if there is no other deadline configured. If a deadline is
   " configured, then use it.

But this is no longer accurate.

Could you also update the comment?

hori-ryota:
My gometalinter returns relative path.

Perhaps this behavior may have changed the behavior of metalinter

[Convert absolute paths to package paths for some linters by dnephin · Pull Request #319 · alecthomas/gometalinter](https://github.com/alecthomas/gometalinter/pull/319)

mirza-s:
I can confirm this fixes #1417 for me.

Carpetsmoker:
I just noticed that someone else made a PR for this previously: #1402; this PR is (slightly) better though, since it also updates a comment :-)

fatih:
@hori-ryota can you remove the accidentally added tags file please?

hori-ryota:
Oh, sorry and thank you for fixing it!

[Remove accidentally committed file; update CHANGELOG · fatih/vim-go@ea33466](https://github.com/fatih/vim-go/commit/ea3346620533dc61f7f3cf7fb9f1325587c40597)

PR #1415, opened 2017-08-18
support for customizable lists types

### LINKED ISSUE #1408: Make go_list_type configurable for each command independently
Currently this setting is global. However we could make it better by giving the user the ability what they want to use for each command. So instead of 

```
let g:go_list_type = "locationlist"
```

we could have:

```
let g:go_list_type_commands = {
  "GoFmt": "locationlist",
  "GoBuild": "quickfix"
}
```

This would override the command's default setting for the list.

bhcleek:
Add a new setting, `g:go_list_type_commands`: a dictionary
whose entries allows users to customize whether to use the quickfix list
or the location list for each command. The dictionary does not require
an entry for every command; only those the user wants to specify.

When g:go_list_type is set and g:go_list_type_commands contains an entry
for a command, go:go_list_type_commands takes precedence.

I'm not totally happy with the keys for the commands in the `g:go_list_type_commands` dictionary. Currently the entries map very closely to the commands (e.g. `:GoBuild`), providing an easy to understand set of keys for most users. Unfortunately, the `GoModifyTags` entry applies to both `:GoRemoveTags` and `:GoAddTags` commands, because the way those two commands are implemented makes it a little difficult to vary the key used and also because it seemed strange to allow those two commands to use different lists since they are each other's inverse.

I considered using a smaller set of keys that are more closely aligned with the general class of command that needs to use the key (e.g. `build`, `install`, `fmt`, `lint`, `modifytags`) instead of the specific command that needs to use the key. Such a scheme would use a single key, `lint`, for all linting commands (e.g. `:GoMetaLinter`, `:GoLint`, `:GoErrCheck`)...

I'm open to suggestions on the key names and how much documentation there should be for them.

I could also use some help testing this, since there's different code paths for Vim 7, Vim 8, and Neovim.

Fixes #1408

### DIFF
--- a/autoload/go/term.vim
+++ b/autoload/go/term.vim
@@ -104,7 +104,7 @@ function! s:on_exit(job_id, exit_status, event) dict abort
   endif
   let job = s:jobs[a:job_id]
 
-  let l:listtype = "locationlist"
+  let l:listtype = go#list#Type("_term", "locationlist")
 
   " usually there is always output so never branch into this clause
   if empty(job.stdout)
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1134,7 +1134,7 @@ The dictionary version allows you to define options for multiple binaries:
 <
                                                     *'g:go_fmt_fail_silently'*
 
-Use this option to disable showing a quickfix list when |'g:go_fmt_command'|
+Use this option to disable showing a location list when |'g:go_fmt_command'|
 fails. By default the location list is shown. >
 
   let g:go_fmt_fail_silently = 0
@@ -1418,11 +1418,25 @@ explicitly overrides this behavior. For standard Vim behavior, set it to 10.
                                                             *'g:go_list_type'*
 
 Specifies the type of list to use for command outputs (such as errors from
-builds, results from static analysis commands, etc...). The default value
+builds, results from static analysis commands, etc...) when a value is not
+specified for the command in g:go_list_type_commands. The default value
 (empty) will use the appropriate kind of list for the command that was called.
 Supported values are "", "quickfix", and "locationlist". >
 
   let g:go_list_type = ""
+<
+
+                                                   *'g:go_list_type_commands'*
+
+Specifies the type of list to use for command outputs (such as errors from
+builds, results from static analysis commands, etc...). The default value
+(an empty dictionary) will use the appropriate kind of list for the command
+that was called. Supported keys are "GoBuild", "GoErrCheck", "GoFmt",
+"GoInstall", "GoLint", "GoMetaLinter", "GoModifyTags", "GoRun", and "GoTest".
+Supported values for each command are "quickfix" and "locationlist".
+
+>
+  let g:go_list_type_commands = {}
 <
 
                                                        *'g:go_list_autoclose'*
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1432,8 +1432,9 @@ Specifies the type of list to use for command outputs (such as errors from
 builds, results from static analysis commands, etc...). The default value
 (an empty dictionary) will use the appropriate kind of list for the command
 that was called. Supported keys are "GoBuild", "GoErrCheck", "GoFmt",
-"GoInstall", "GoLint", "GoMetaLinter", "GoModifyTags", "GoRun", and "GoTest".
-Supported values for each command are "quickfix" and "locationlist".
+"GoInstall", "GoLint", "GoMetaLinter", "GoModifyTags" (used for both
+:GoAddTags and :GoRemoveTags), "GoRename", "GoRun", and "GoTest".  Supported
+values for each command are "quickfix" and "locationlist".
 
 >
   let g:go_list_type_commands = {}
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1418,24 +1418,23 @@ explicitly overrides this behavior. For standard Vim behavior, set it to 10.
                                                             *'g:go_list_type'*
 
 Specifies the type of list to use for command outputs (such as errors from
-builds, results from static analysis commands, etc...) when a value is not
-specified for the command in g:go_list_type_commands. The default value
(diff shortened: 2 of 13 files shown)


Carpetsmoker:
I stuck this in my vimrc:

    let g:go_list_type = "locationlist"
    let g:go_list_type_commands = {"GoBuild": "quickfix"}


But when using `:GoBuild` error still open in the location list? I would expect the quickfix list?

Am I doing it wrong, or did I discover a problem?

Using Vim 8.0.987 on Linux and vim-go master.

Carpetsmoker:
> .

I considered using a smaller set of keys that are more closely aligned with the general class of command that needs to use the key (e.g. build, install, fmt, lint, modifytags) instead of the specific command that needs to use the key. Such a scheme would use a single key, lint for all linting commands (e.g. :GoMetaLinter, :GoLint, :GoErrCheck)...

Personally I think this approach makes more sense; is there ever a situation where you want to open `:GoLint` output in the quickfix and `:GoMetaLinter` in the location list? Probably not.

That being said, I'm not "against" the current approach.

fatih:
@bhcleek Why not make the `for` variable a list? That way it could match multiple things, so we could add `["GoAddTags", "GoRemoveTags"]` for the case where it's only in one place and in other places it would be `["GoBuild"]`.

bhcleek:
@Carpetsmoker vim-go master doesn't have this change yet, so `g:go_list_type` is the only variable that's being checked in your test.

bhcleek:
Good idea @fatih. I'll make that change. It will simplify a few things, too.

Carpetsmoker:
Sorry, I'm not sure why I said in my comment that I was using master, I was using this PR (do'h).

bhcleek:
@fatih On second thought, I don't think it would be possible to make `for` a list. Imagine `
let g:go_list_type_commands = ["GoAddTags", "quickfix", "GoRemoveTags": "locationlist"]`

What would `go:list:Type(["GoAddTags", "GoRemoveTags"], "locationlist")` return?

bhcleek:
@Carpetsmoker looks like a bug. I know the source of it, and will fix it tonight.

bhcleek:
@Carpetsmoker I believe I've resolved the issue you discovered. Can you verify that it's resolved for you, too?

autoload/go/term.vim
```diff
@@ -104,7 +104,7 @@ function! s:on_exit(job_id, exit_status, event) dict abort
   endif
   let job = s:jobs[a:job_id]
 
-  let l:listtype = "locationlist"
+  let l:listtype = go#list#Type("_term", "locationlist")
```
Carpetsmoker: The `_term` value isn't documented; is that intentional?

doc/vim-go.txt
```diff
@@ -1418,13 +1418,28 @@ explicitly overrides this behavior. For standard Vim behavior, set it to 10.
...
 
   let g:go_list_type = ""
 <
 
+                                                   *'g:go_list_type_commands'*
+
+Specifies the type of list to use for command outputs (such as errors from
+builds, results from static analysis commands, etc...). The default value
+(an empty dictionary) will use the appropriate kind of list for the command
+that was called. Supported keys are "GoBuild", "GoErrCheck", "GoFmt",
```
Carpetsmoker: I think that "The default value (an empty dictionary) will use the appropriate kind of list for the command that was called." isn't 100% true, as `g:go_list_type` can override it.

Perhaps it's an idea to add a single help entry for both `g:go_list_type` and `g:go_list_type_commands` which describes how vim-go figures out which list type to use? For example something along the lines of:

```
                                                            *'g:go_list_type'* 
                                                   *'g:go_list_type_commands'*

Specifies the type of list to use for command outputs (such as errors from
builds, results from static analysis commands, etc.)
With `g:go_list_type` you can specify which list to open by default. This
default value can be overridden for specific commands with
`g:go_list_type_commands`, which is a dictionary to map a command to a value
("quickfix" or "locationlist").

Supported keys are "GoBuild", "GoErrCheck", "GoFmt", "GoInstall", "GoLint",
"GoMetaLinter", "GoModifyTags" (used for both :GoAddTags and :GoRemoveTags),
"GoRename", "GoRun", and "GoTest".

If neither options are set then vim-go will use the most appropriate kind of
list for the command that was called.

For example to use the location list by default but the quickfix list for
|:GoInstall|: >
  let g:go_list_type = "quickfix"
  let g:go_list_type_commands = {"GoBuild": "locationlist"}
<
```

Carpetsmoker:
Yup, seems to work well now :+1:

autoload/go/term.vim
bhcleek: Yes. From the most recent commit message:
> When displaying the results of a command that was run in a terminal,
pass "_term" to go#list#Type.  _term should not be considered one of the
supported values of g:go_list_type_commands.

doc/vim-go.txt
bhcleek: That's possible, but note that the documentation for `g:go_list_type` does document that it is used when `g:go_list_type_commands` does not contain an entry for the command:
> Specifies the type of list to use for command outputs (such as errors from
builds, results from static analysis commands, etc...) when a value is not
specified for the command in g:go_list_type_commands. The default value
(empty) will use the appropriate kind of list for the command that was called.

bhcleek:
I edited the doc to mention the interplay of `g:go_list_type` and `g:go_list_type_commands`.

PTAL

doc/vim-go.txt
Carpetsmoker: Yeah, but if you use `:help 'g:go_list_type_commands'` then you won't see that part, because it's (probably) off-screen.

Your last commit looks good :+1:

bhcleek:
@Carpetsmoker does this need your LGTM on the PR as opposed to an in-line comment for visibility?

fatih:
Thanks @bhcleek 👍

bhcleek:
🙇

PR #1420, opened 2017-08-29
Add support of g:go_doc_command and g:go_doc_options parameters

nk2ge5k:


Carpetsmoker:
Personally I think it might be a good idea store this as a list, instead of two separate options:

    g:go_doc_command = ["godoc"]

You can then set this to `["go", "doc"]`.

Or perhaps even store all commands as a dict, instead of many options:

```
g:go_commands = {
    \ "go": ["go"],
    \ "godoc": ["godoc"],
    \ "gometalinter": ["gometalinter"],
    " ..etc..   
\}

I don't know what Fatih thinks, though.

**Comment by fatih** [comment]
Yeah, a list is better. We can then join later. For `go_commands` I think that's something we can tackle later as it'll require a lot of work, don't want to add it now at least.

**Comment by nk2ge5k** [comment]
Sorry for delay. 
I made g:go_doc_command type of list and removed g:go_doc_options. Hope its better now.

File: autoload/go/doc.vim [hunk 1]
```diff
@@ -60,12 +56,11 @@ endfunction
...
   " With argument: run "godoc [arg]".
   if len(a:000)
-    let bin_path = go#path#CheckBinPath('godoc')
-    if empty(bin_path)
+    if empty(go#path#CheckBinPath(g:go_doc_command[0]))
       return
     endif
 
-    let command = printf("%s %s", go#util#Shellescape(bin_path), join(a:000, ' '))
+    let command = printf("%s %s", join(g:go_doc_command, ' '), join(a:000, ' '))
```
Carpetsmoker: This removes the `go#util#Shellescape()` part. You should probably map that to every entry in `g:go_doc_command`?

(This should probably be done by the `System()` function, but changing that would require careful testing of all the `System()` invocations and may break people's configs).

Carpetsmoker:
Seems to work well in my testing, but there is a caveats with this: the `go_doc_command` setting is *only* used when invoked as `:GoDoc packagename`. If you use `:GoDoc` or `K` to get the documentation under the cursor, vim-go will always use [`gogetdoc`](https://github.com/zmb3/gogetdoc).

It would be good to update the documentation as well, because this confused me rather a lot and it took me a trip to the source code to figure out why setting `g:go_doc_command` doesn't work.

Currently it reads:

	Use this option to define which tool is used to godoc. By default `godoc` is
	used

IMHO something like this would be better:

	Command to use for |:GoDoc|; only used when it's invoked with a package
	name. The `gogetdoc` command is always used when |:GoDoc| is used to get the
	documentation of the identifier under the cursor.

Carpetsmoker:
Thank you @nk2ge5k.

nk2ge5k:
Thank you

PR #1426, opened 2017-09-05
Vim syntax match for variable definitions

tdewolff:
This adds a match for variable definitions so that they can be highlighted using a color scheme. Variable definitions are the `foo` and `bar` in:

    foo := ...
    foo, bar := ...

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1332,6 +1332,13 @@ By default it's enabled. >
 
   let g:go_highlight_format_strings = 1
 <
+                                      *'g:go_highlight_variable_declarations'*
+
+Highlight variable declarations (`:=`). Variable assignments (`=`) are not
+highlighted. By default it's disabled. >
+
+  let g:go_highlight_variable_declarations = 0
+<
                                                     *'g:go_autodetect_gopath'*
 
 Automatically modifies GOPATH for certain directory structures, such as for
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -389,6 +389,7 @@ hi def link     goDeclType          Keyword
 " Variable Declarations
 if g:go_highlight_variable_declarations != 0
   syn match goVarDefs /\v\w+(,\s*\w+)*\ze(\s*:\=)/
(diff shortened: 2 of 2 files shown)


Carpetsmoker:
Two comments:

1. It only seems to work for the last variable? Using this:

       :syn match goVarDefs /\w\+\(,\s*\w+\)*\s*\ze:=/
       :hi def link goVarDefs Statement

   Gives me (the brownish colour being `Statement`):

     2017-09-05-220910_254x127_scrot

   Is this intentional? It seems strange to me (I would expect both `a` and `err` to be highlighted).

2. The syntax groups isn't highlighted by default, but will still be applied, which will be a performance hit. You should probably put this behind a variable (like the existing `g:go_highlight_*` ones).

tdewolff:
Hmm that is strange, it works for multiple variable for me (that's the `(,\s* ... )` part). I'll do some testing

Carpetsmoker:
Just tested again; it really doesn't work for me in Vim 8.0.987 with the current vim-go mater.

It looks like the problem is you forgot to add a `\` to the `+` inside the capture group; if I add that it works:

    syn match goVarDefs /\w\+\(,\s*\w\+\)*\s*\ze:=/

I don't know why it works for you...

Also, to avoid highlighting spaces between the last variable and `:=` (only shows if you set the highlight to something with a different background colour):

    syn match goVarDefs /\w\+\(,\s*\w\+\)*\ze\(\s*:=\)/

And with `\v` added to cure the Leaning Toothpick Syndrome and make it a wee bit more readable:

    syn match goVarDefs /\v\w+(,\s*\w+)*\ze(\s*:\=)/

tdewolff:
Ah I see you are well versed in Vim regexes. I've changed it.

fatih:
@Carpetsmoker I've added you to the repo. From now on please feel free to merge & review syntax based PR's as I think you're the best fit. Just don't forgot to update the Changelog.

Carpetsmoker:
Thanks @fatih :+1:

PR #1428, opened 2017-09-06
folding improvements

hexasoftware:
Improved fold syntax to include comments above, currently only working on funcs

open
closed

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -247,6 +247,7 @@ else
 endif
 
 if s:fold_comment
+	" changed this to comments instead of declaration_comment
   syn region    goComment           start="/\*" end="\*/" contains=@goCommentGroup,@Spell fold
   syn match     goComment           "\v(^\s*//.*\n)+" contains=goGenerate,@goCommentGroup,@Spell fold
 endif
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -93,7 +93,7 @@ let s:fold_block = 1
 let s:fold_import = 1
 let s:fold_varconst = 1
 let s:fold_package_comment = 1
-let s:fold_comment = 1
+let s:fold_comment = 0
 
 if exists("g:go_fold_enable")
   if index(g:go_fold_enable, 'block') == -1
@@ -170,7 +170,7 @@ if s:fold_comment
   syn region    goComment           start="/\*" end="\*/" contains=@goCommentGroup,@Spell fold
   syn match     goComment           "\v(^\s*//.*\n)+" contains=goGenerate,@goCommentGroup,@Spell fold
 else
-	syn region    goComment           start="/\*" end="\*/" contains=@goCommentGroup,@Spell
+  syn region    goComment           start="/\*" end="\*/" contains=@goCommentGroup,@Spell
 endif
 
 hi def link     goComment           Comment
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -92,7 +92,8 @@ endif
 let s:fold_block = 1
 let s:fold_import = 1
 let s:fold_varconst = 1
-let s:fold_func = 1
+let s:fold_doc = 1
+
 if exists("g:go_fold_enable")
   if index(g:go_fold_enable, 'block') == -1
     let s:fold_block = 0
@@ -103,8 +104,8 @@ if exists("g:go_fold_enable")
   if index(g:go_fold_enable, 'varconst') == -1
     let s:fold_varconst = 0
   endif
-  if index(g:go_fold_enable, 'func') == -1
-    let s:fold_func = 0
+  if index(g:go_fold_enable, 'doc') == -1
+    let s:fold_doc = 0
   endif
 endif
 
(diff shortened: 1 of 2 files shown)


hexasoftware:
Actually there is some issue with syntax coloring and I notice the default folds seems odd

normal-behaviour 
On the left i have the foldcolumn and with normal behaviour seems to be wrong. Now I'm not sure if this is a vim-go issue or if i have any other plugin performing this


disabled-behaviour
When the fold on varconst is disabled the folds seems proper

Carpetsmoker:
FYI, I have a related PR to fold package-level coments: #1377. I've been waiting for that to be merged to do more work on the folding.

Carpetsmoker:
I'm afraid I don't quite understand what you mean with your last comment btw. The screenshots are kind of confusing since there is so much going on, and I'm not sure which colour stuff *ought* to be.

---

At any rate, I tested this PR. I noticed that it also folds the function or type declaration. Is this intentional? Especially with var/const/type blocks (`var ( ... )`) you lose hat line, so you no longer know if it's a `var` or `const`.
But I would personally also consider it annoying with just functions. The function signature is kind of important :-)

---

It doesn't seem to work with block comments btw:

```
/* asd
* asd
*/
func main() {
```

I get `E490: No fold found` when using `zc`.

---

It also noticed it doesn't work nested stuff like so:

```
// asd
// zxczxc
type (
    // asdf
    // zxcv
    foo interface {
        ASD()
    }
)
```

hexasoftware:
And seems that fatih  merged couple minutes ago, I will try to solve the conflicts and merge into mine

Carpetsmoker:
Thanks @hexasoftware; my (now merged) PR added `package_comment` to the list in `g:go_fold_enable`; it's probably clearer if you rename your variable to something like `declaration_comment` or some such? Just to make it clearer what applies where.

Let me know if you need help. It's a good PR to add :+1:

hexasoftware:
I agree should be renamed, I've seen some references to same blocks as 'textobjs', refering to a func including its doc, but 'textobjs' wouldn't be intuitive

hexasoftware:
@Carpetsmoker what do you think, should I create these per types? as in func_comment, type_comment varconst_comment? or for all? right now anything with a comment in a text next to it would fold

Carpetsmoker:
> what do you think, should I create these per types? as in func_comment, type_comment varconst_comment? or for all?

I think both approaches are fine.

I added the package comment separately because personally I *don't* want to fold function comments (as they're often useful), but *do* want to fold package comments (as they're often an introduction, and not useful to me).

So the question is, are there people who want to fold type comments but not function comments? Maybe? Maybe not? I don't know what the answer is to that :-)

Just go with what works well for you and/or is easiest to implement. It can always be changed later :-)

> right now anything with a comment in a text next to it would fold

Yeah, that's probably not great :-) You can see in my PR on how to fix that: https://github.com/fatih/vim-go/pull/1377/files#diff-6bbebb95ce25cd40d0072544e2dad91eR418 – this will fold comments *only* if it's followed by the text `package `.

Another way to do it would be to use `nextgroup`; I'm not sure why I didn't use that there.

hexasoftware:
My original goal was indeed to to fold both comment and func as a single fold, as I've seen in other languages some pretty folds with only descriptions but I wasn't able to solve the highlight problems as a 'func' contained keyword shows differently than a goDeclaration.

for this PR I just added folding for any kind of repeated comment either /*...*/ or repeated //.. which I've seen requested in other situations.

I would like the fold to include declarations in the future(in other option), I will check better on that when I have the time, as is not a 'conventional' fold

Carpetsmoker:
> for this PR I just added folding for any kind of repeated comment either /.../ or repeated //.. which I've seen requested in other situations.

That sounds like a reasonable place to start with.

> I would like the fold to include declarations in the future(in other option)

I attempted to do that [with this](https://github.com/fatih/vim-go/compare/master...Carpetsmoker:fold-func), but it doesn't quite work in all situations (open the rest file).

syntax/go.vim
```diff
@@ -240,6 +246,13 @@ else
                         \ contains=ALLBUT,goParen,goBlock,goFunction,goTypeName,goReceiverType,goReceiverVar
 endif
 
+if s:fold_comment
+	" changed this to comments instead of declaration_comment
+  syn region    goComment           start="/\*" end="\*/" contains=@goCommentGroup,@Spell fold
+  syn match     goComment           "\v(^\s*//.*\n)+" contains=goGenerate,@goCommentGroup,@Spell fold
+endif
```
Carpetsmoker: Looks good and works well in my testing, but could you move this next to the regular `goComment` lines on line 168? I think that makes more sense and would make it easier to maintain later on.

Would also be nice if you could remove the extra blank lines you added :-)

syntax/go.vim
```diff
@@ -240,6 +246,13 @@ else
                         \ contains=ALLBUT,goParen,goBlock,goFunction,goTypeName,goReceiverType,goReceiverVar
 endif
 
+if s:fold_comment
+	" changed this to comments instead of declaration_comment
```
Carpetsmoker: I think this comment can be removed?

syntax/go.vim
```diff
@@ -452,4 +465,5 @@ syn sync minlines=500
 
 let b:current_syntax = "go"
 
-" vim: sw=2 ts=2 et
```
Carpetsmoker: I think we want to keep the modeline; you you add this back please?

hexasoftware: Oh, sry I guess I removed modeline within my tests down there and didn't realize, I was also trying to create a FoldText() function so in the cases we fold the declarations the text would contain the subject (func/type/const/var),

I made the suggested changes

hexasoftware:
> I attempted to do that with this, but it doesn't quite work in all situations (open the test file).

I tried, and this is meant to fold only the comments above the functions? the multiline comments would serve that purpose as well, unless indeed users don't want to fold all but only descriptions of certain types

the very first screenshots I posted was kind of my intention of fold funcs and was working except for the highlight (and probably other) issues

hexasoftware:
Keep in mind by leaving s:fold_comment = 0 I think there is no way for user to turn on?

Carpetsmoker:
Ugh. Looks like I forgot to test that feature after I added it @hexasoftware >_< Good catch :+1: I fixed it now.

PR #1433, opened 2017-09-11
CHANGELOG.md: fix markdown syntax to link to GH issues/pr's

fatih:


fatih:
Alright, seems like I have to remove the brackets so it's valid. Will open another PR that fixes it

Issue #1234, opened 2017-03-08
More linkage to doc/vim-go.txt


rigelrozanski:
There are very few links which lead to the amazing document you folks have created under doc/vim-go.txt
I've been using this repo for quite some time and wish I had known about this doc earlier on. 
It would be great to see references to this doc under:
 - Readme.md
 - Wiki pages, usage and home
   - https://github.com/fatih/vim-go/wiki
   - https://github.com/fatih/vim-go/wiki/Usage

The only link in the entire repo I saw which led to this was in the contributing guidelines for submitting this issue (and that link was broken!). Only came across that doc for the first time when googling "vim-go change GoGuruScope". Maybe I'm just a dumb user, but I'd imagine others are in my boat. 

Will gladly write a few lines and PR if there is general agreement. Thoughts?

rigelrozanski:
Update: just saw the reference in the readme (under usage). but it's just code... maybe use a hyperlink too?

fatih:
Haha :) You're definitely right @rigelrozanski Eventually I should get rid of the whole readme and just use `docs/` folder, because that's the way documentation is written in Vim. And then you can for example type `:help :GoGuruScope` and it shows you the documentation, or `:help vim-go` shows the beginning of the documentation. Let's have this open so I'll fix this later. Thanks!

fatih:
This is fixed now. Thanks for the feedback. See latest readme.md: https://github.com/fatih/vim-go :) Also populated doc/vim-go.txt with all the information. I think it's much more better now.

PR #1434, opened 2017-09-11
CHANGELOG.md: fix GH links syntax

fatih:
@Carpetsmoker seems like the `#1234` syntax doesn't work. Do we have to use the URL link syntax? is this a bug?


fatih:
Merged to see what's wrong with the rendering.

Carpetsmoker:
I thought this worked, but guess not? I can't get it to work no matter what I do in my test repo :-(
I can't find anything in the docs either ... maybe support for it was removed together with support for `[gh-1]` style references?

fatih:
Seems like Github uses different Markdown renderers for their repo, issues and wiki's. Lesson learned. Let me convert them to standard markdown syntax so we have at least full URL's.

PR #1435, opened 2017-09-11
Fix go#package#FromPath()

Carpetsmoker:
This didn't work at all on my system. Turned out the problem was that on
my system (Linux) some paths didn't have a trailing slash, so
`workspace` would be `/home/martin/gosrc`.

This makes sure that it always works, no matter if a trailing slash is
present or not.

Also `break` from the loop on the first match, so it selects the
*first* package found in GOPATH rather than the *last*. This is also how
imports in Go behave.


fatih:
lgtm

PR #1436, opened 2017-09-11
Fix ChangeLog links

Carpetsmoker:
:%s,\(#\d\+\),[\1](https://github.com/fatih/vim-go/pull/\1),


Carpetsmoker:
Seems to be okay: https://github.com/Carpetsmoker/vim-go/blob/2df7ee9d8cd5163b3a4757b41aaf55d06e9e4c0c/CHANGELOG.md

I can also change the link text to the old format of `[gh-n]`? I don't really care and it's a quick change.

fatih:
if you can do `GH-` it would be perfect. Thanks!

PR #1437, opened 2017-09-11
Support fzf on GoDecls[Dir]

fsouza:
Revisiting #976.

This introduces a new configuration option (``g:go_decls_mode``). It's still missing documentation, but I wanted to open the PR anyway so I can get feedback on the implementation before moving forward with the documentation.

### DIFF
--- a/autoload/fzf/decls.vim
+++ b/autoload/fzf/decls.vim
@@ -45,7 +45,7 @@ function! s:sink(str)
     let cmd = get({'ctrl-x': 'split',
           \ 'ctrl-v': 'vertical split',
           \ 'ctrl-t': 'tabe'}, a:str[0], 'e')
-    execute cmd filepath
+    execute cmd fnameescape(filepath)
     call cursor(line, col)
     silent! norm! zvzz
   finally
--- a/autoload/fzf/decls.vim
+++ b/autoload/fzf/decls.vim
@@ -1,0 +1,154 @@
+function! s:code(group, attr)
+  let code = synIDattr(synIDtrans(hlID(a:group)), a:attr, "cterm")
+  if code =~ '^[0-9]\+$'
+    return code
+  endif
+endfunction
+
+function! s:color(str, group)
+  let fg = s:code(a:group, "fg")
+  let bg = s:code(a:group, "bg")
+  let bold = s:code(a:group, "bold")
+  let italic = s:code(a:group, "italic")
+  let reverse = s:code(a:group, "reverse")
+  let underline = s:code(a:group, "underline")
+  let color = (empty(fg) ? "" : ("38;5;".fg)) .
+            \ (empty(bg) ? "" : (";48;5;".bg)) .
+            \ (empty(bold) ? "" : ";1") .
+            \ (empty(italic) ? "" : ";3") .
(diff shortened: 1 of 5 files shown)


fsouza:
I just added some documentation, please also review that :D

Carpetsmoker:
Thanks :+1: Seems to work well in my testing.

---

Using `:GoDecls` gives me:

    vim-go: ctrlp.vim plugin is not installed. Please install from: https://github.com/ctrlpvim/ctrlp.vim

This message is somewhat confusing, since it should be more like "please install ctrlp.vim, fzf, or unite.vim" (unite.vim support was added in #1391, would be nice if `g:go_decls_mode` would also support unite.vim btw).

Actually, the best would be to autodetect one of those plugins and just use it, and have `g:go_decls_mode` only if you want to force using one of those plugins (in case you have multiple installed).

fsouza:
@Carpetsmoker Unite is supported through a different command, ``GoDecls`` can only use ``fzf`` or ``ctrlp.vim``.

I'll update the PR so it detects the tool and use it, preferring ctrlp.vim so we can keep the current behavior.

Carpetsmoker:
> Unite is supported through a different command, GoDecls can only use fzf or ctrlp.vim.

Wouldn't it be possible to have `:GoDecls` call `:Unite decls` if unite.vim is installed or if the `g:go_decls_mode` is set to `unite`?

I have no idea if this makes sense; I don't use any of these plugins myself :-) But it seems to make that this would make this command easier to use for people who have unite.vim installed.

autoload/fzf/decls.vim
```diff
@@ -0,0 +1,154 @@
...
+    let line =  vals[2]
+    let col =  vals[3]
+
+    " i.e: /Users/fatih/vim-go/main.go
+    let filepath =  fnamemodify(filename, ":p")
+
+    let cmd = get({'ctrl-x': 'split',
+          \ 'ctrl-v': 'vertical split',
+          \ 'ctrl-t': 'tabe'}, a:str[0], 'e')
+    execute cmd filepath
```
Carpetsmoker: I think you want to escape this with `fnameescape(filepath)`, otherwise it will break if the path contains characters (see `:help fnameescape()`).

fsouza: Ohh, nice to know, thank you!

I've updated the code.

fsouza:
@Carpetsmoker I think the integration with Unite exists to support people that use ``Unite`` (and therefore ``Unite decls``), it's "the Unite way", so I don't think it makes sense to make GoDecls call Unite, but I can do it if you think that's the best approach.

I just finished the logic for automatically picking between fzf and ctrlp. Please let me know what you think.

Thanks for the feedback!

Carpetsmoker:
Okay, did some more testing and works well; I did still encounter some problems when using paths with spaces in it. I fixed that in the above commit. I also tweaked the docs a wee bit in that.

> `Unite decls` is "the Unite way", so I don't think it makes sense to make GoDecls call Unite, but I can do it if you think that's the best approach.

If that's the standard way then it's fine :+1:

fsouza:
@Carpetsmoker awesome, thank you very much!

PR #1442, opened 2017-09-13
Load ftplugin/html.vim for gohtmltmpl

Carpetsmoker:
This will make `%` with the `matchit` plugin loaded work as expected,
and also adds some other features.

### DIFF
--- a/ftplugin/gohtmltmpl.vim
+++ b/ftplugin/gohtmltmpl.vim
@@ -1,8 +1,7 @@
 if exists("b:did_ftplugin")
   finish
 endif
-let b:did_ftplugin = 1
 
-setlocal commentstring=<!--\ %s\ -->
+runtime! ftplugin/html.vim
 
 " vim: sw=2 ts=2 et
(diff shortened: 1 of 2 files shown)


ftplugin/gohtmltmpl.vim
```diff
@@ -1,8 +1,7 @@
 if exists("b:did_ftplugin")
   finish
 endif
-let b:did_ftplugin = 1
 
-setlocal commentstring=<!--\ %s\ -->
+runtime! ftplugin/html.vim
```
fatih: Is this available by default for all Vim versions?

Carpetsmoker: The `git log` for `runtime/ftplugin/html.vim` goes back to 7.0001, so it seems so

fatih:
lgtm

Carpetsmoker:
Ugh, I added another commit with the CHANGELOG entry and now your review got cleared @bhcleek >_<

PR #1443, opened 2017-09-13
Implement :GoFillStruct

### LINKED ISSUE #1425: feature-request: Initialize struct instance command
It would be nice to have something like `:GoInitStruct`.
### Proposed workflow.
Consider we have this struct:
```go
type Car struct {
    Name string
    Color string
    Weight int
}
```
And when I create instance of it
```go
car := &Car{}
```
I place a cursor at `Car{}` and run `:GoInitStruct`, which converts the code to:
```go
car := &Car{
    Name: "",
    Color: "",
    Weight: 0,
}
```
[ideally] it also moves cursor to the first field value assignment [where I can start editing it immediately], and when I press Tab (for example), it moves cursor to next field assignment.

Carpetsmoker:
Just the bare basics; fixes #1425.

### DIFF
--- a/autoload/go/fillstruct.vim
+++ b/autoload/go/fillstruct.vim
@@ -1,0 +1,57 @@
+function! go#fillstruct#FillStruct() abort
+  let binpath = go#path#CheckBinPath('fillstruct')
+  if empty(binpath)
+    return
+  endif
+
+  let l:cmd = [binpath,
+      \ '-file', bufname(''),
+      \ '-offset', go#util#OffsetCursor()]
+
+  " Read from stdin if modified.
+  if &modified
+    call add(l:cmd, '-modified')
+    let l:out = go#util#System(go#util#Shelljoin(l:cmd), go#util#archive())
+  else
+    let l:out = go#util#System(go#util#Shelljoin(l:cmd))
+  endif
+
+  if go#util#ShellError() != 0
+    call go#util#EchoError(l:out)
+    return
+  endif
+
+  try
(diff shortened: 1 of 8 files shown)


Carpetsmoker:
This is now done. It depends on #1456, otherwise errors don't look very good.

I haven't implemented the proposed `:GoCode` from #1445; can do that in a future PR, after we've decided how this will look.

fatih:
This looks good. Wish it could work when you put it everywhere on the line instead of putting it on the struct literal. I think this is something @davidrjenni can fix though. Adding him if he can chime on that.

fatih:
@Carpetsmoker can you add docs please for this

autoload/go/fillstruct.vim
```diff
@@ -0,0 +1,57 @@
...
+    let l:code[len(l:code)-1] .= getline('.')[col('.'):] 
+
+    " Indent every line except the first one; makes it look nice.
+    let l:indent = repeat("\t", indent('.') / &ts)
+    for i in range(1, len(l:code)-1)
+      let l:code[l:i] = l:indent . l:code[i]
+    endfor
+
+    " Out with the old ...
+    exe 'normal! ' . l:json['start'] . 'gov' . l:json['end'] . 'gox'
```
fatih: What is gov and gox here?

Carpetsmoker: It's:

- `123go` – go to offset/
- `v132` – go to offset.
- `x` – delete.

PR #1449, opened 2017-09-14
Consolidate list type command keys

bhcleek:
Refactor go#list#Type for maintainability. Instead of having the
supported keys (both internally and externally supported) spread
throughout the code, use a single map that contains the keys and their
defaults in autoload/go/list.vim.

Refactor go#list#Type so that it is aware of the map of default values
and does not need its second parameter.

After refactoring, I realized that the default list type for `GoAddTags` and `GoRemoveTags` should be `locationlist` instead of `quickfix`, so I made that change, too.

### DIFF
--- a/autoload/go/list.vim
+++ b/autoload/go/list.vim
@@ -128,8 +128,37 @@ function! s:listtype(listtype) abort
   return a:listtype
 endfunction
 
-function! go#list#Type(for, default) abort
-  let l:listtype = s:listtype(a:default)
+" s:default_list_type_commands is the defaults that will be used for each of
+" the supported commands (see documentation for g:go_list_type_commands). When
+" defining a default, quickfix should be used if the command operates on
+" multiple files, while locationlist should be used if the command operates on a
+" single file or buffer. Keys that begin with an underscore are not supported
+" in g:go_list_type_commands.
+let s:default_list_type_commands = {
+  "GoBuild": "quickfix",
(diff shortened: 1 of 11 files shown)


autoload/go/list.vim
```diff
@@ -128,8 +128,37 @@ function! s:listtype(listtype) abort
...
+" in g:go_list_type_commands.
+let s:default_list_type_commands = {
+  "GoBuild": "quickfix",
+  "GoErrCheck": "quickfix",
+  "GoFmt": "locationlist",
+  "GoGenerate": "quickfix",
+  "GoInstall": "quickfix",
+  "GoLint": "quickfix",
+  "GoMetaLinter": "quickfix",
+  "GoModifyTags": "locationlist",
```
fatih: Maybe we should rename this to `_modifytags`, how we use `_guru` that represents multiple vim commands?

bhcleek: Keys that begin with `_` don't mean that they represent multiple commands, it means that they're not supported to be set in `g:go_list_type_commands`.  Think of keys that start with `_` as private-by-convention to `vim-go`.

If someone were to set them (which is *possible*), vim-go would pick up their overrides, but we are making no guarantees that they will continue to be supported; we may decide there there's a better way to support the commands that currently use `_term` and `_guru`, and I wanted to give us the freedom to do that.

fatih:
lgtm

PR #1450, opened 2017-09-14
Fix :GoBuild shell escaping

### LINKED ISSUE #1404: GoBuild overquotes it's arguments
### Behavior

When I run `GoBuild -tags=foo` strace shows the vim-go running `execve("go", [ "go", "'tags=foo'" ])` (note the extra quoting introduced in the tags argument).

Commenting out the lines in `cmd.vim`:

    " escape all shell arguments before we pass it to make
    if !has('nvim')
      let goargs = go#util#Shelllist(goargs, 1)
    endif

fixes this. It looks like the macro is expecting the command to be run via a shell (except for nvim?) when it is being directly executed.

This is with go-vim 1.1.14 (pulled from github juts now), go 1.8, and vim 8.0.823 using Fedora 26.

Carpetsmoker:
There was a stray `go#util#Shelllist()` in there. As near as I can tell
this isn't needed at all. Probably a left-over from some earlier
version?

Also move the non-async code in an `else`, I think this makes it a lot
clearer which code path does what. The two early returns were kinda
confusing.

Tested with Vim 8, 7.4, and Neovim, and seems to work well.

Fixes #1404

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ BUG FIXES:
 * The `gohtmltmpl` filetype now sources the `html` ftplugin, so that `matchit`,
   completion, and some other things work better.
   [[GH-1442]](https://github.com/fatih/vim-go/pull/1442)
+* Fix `:GoBuild` shell escaping [[GH-1450]](https://github.com/fatih/vim-go/pull/1450).
 
 BACKWARDS INCOMPATIBILITIES:
 
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -9,20 +9,17 @@ endfunction
 " default it tries to call simply 'go build', but it first tries to get all
 " dependent files for the current folder and passes it to go build.
 function! go#cmd#Build(bang, ...) abort
-  " expand all wildcards(i.e: '%' to the current file name)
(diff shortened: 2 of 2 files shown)


fatih:
At some point we should ditch supporting 7.4 Not sure when, probably not in near future but we should eventually.

lgtm

Carpetsmoker:
> At some point we should ditch supporting 7.4 Not sure when, probably not in near future but we should eventually.

I was thinking the same; Ubuntu 16.04 (the most recent LTS) has Vim 7.4.1689; the next LTS release (18.04) will have Vim 8, so that might be a good time?

I also saw that some functions don't actually have async support, for example for :GoInstall it's implemented for just Vim but not Neovim. It's annoying that the APIs are different :-/

fatih:
Yeah :/ The reason is that I'm not using neovim. Initially when it was released I've added some async capabilities for neovim. But then vim-go grew and grew,so I couldn't keep up with neovim support. I've decided after that vim-go should have first class support for Vim, but for Neovim it shouldn't have it.  So whenever there was an issue for Neovim and I had time, I would add it. That's why the API is so different. 

> the next LTS release (18.04) will have Vim 8, so that might be a good time? 

Yeap definitely!

PR #1461, opened 2017-09-16
disable g:go_autodetect_gopath by default

bhcleek:
Disable g:go_autodetect_gopath by default since first class support for
vendoring was enabled in Go by default a year and a half ago in Go 1.6
and most vendoring tools use it. At this point, there is a greatly
reduced need to detect whether the old method used by Godep and gb is in
effect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ IMPROVEMENTS:
   gets pre-filled can be configured with `g:go_gorename_prefill` option.
   In addition `:GoRename <Tab>` now lists some common options.
   [[GH-1465]](https://github.com/fatih/vim-go/pull/1465).
+* Disable `g:go_autodetect_gopath` by default. [[GH-1461]](https://github.com/fatih/vim-go/pull/1461).
 
 ## 1.15 - (October 3, 2017)
 
--- a/autoload/go/path.vim
+++ b/autoload/go/path.vim
@@ -73,7 +73,7 @@ function! go#path#Detect() abort
   let gopath = go#path#Default()
 
   " don't lookup for godeps if autodetect is disabled.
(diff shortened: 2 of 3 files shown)


Carpetsmoker:
Personally, I think that the fewer tricks we play with `GOPATH` the better, so
:+1:.

There is the additional issue that right now `go#path#Detect()` is applied
fairly inconsistently; it's called only for some commands not others.

so I would propose going a step further and removing the `go#path#Detect()`
calls from the commands and instead hooking in to the `FileType` autocommand
instead when `g:go_autodetect_gopath` is enabled.

bhcleek:
I agree that using `FileType` autocommands offers a better method to detect `GOPATH`, but I think that's a separable task, and I'd prefer to tackle it separately (I'll do that as my next contribution).

bhcleek:
@Carpetsmoker are you on board with merging this as is?

Carpetsmoker:
> @Carpetsmoker are you on board with merging this as is?

Yeah, sure. It would be good to get @fatih's input on this too though.

fatih:
While at this, we should also remove the autodetection completely. The whole logic around godep and gb is not needed in my opionion. With `vendor/` people know have the ability of Project based Go applications.

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ IMPROVEMENTS:
 
         let g:go_list_type_commands = {"GoBuild": "quickfix", "GoTest": "locationlist"}
 
+* Disable `g:go_autodetect_gopath` by default [[GH-1461]](https://github.com/fatih/vim-go/pull/1461).
```
fatih: Can you move this up to the latest changelog? This seems to be in the old place as we released a new version.

fatih:
lgtm after the changelog fix 👍

PR #1463, opened 2017-09-16
teach goVarAssign about all assignment operators

bhcleek:
Add regular expression branches so that all assignment operators will be
highlighted (e.g. +=, |=, et al.).

Add a match for character other than '=' at the end of the goVarAssign
match expression so that equality comparision (i.e. '==') won't match.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -410,7 +410,7 @@ hi def link     goDeclType          Keyword
 
 " Variable Assignments
 if g:go_highlight_variable_assignments != 0
-  syn match goVarAssign /\v[_.[:alnum:]]+(,\s*[_.[:alnum:]]+)*\ze(\s*(\+|\-|\||\^|\*|\/|\%|\<\<|\>\>|\&|\&\^)?\=[^\=])/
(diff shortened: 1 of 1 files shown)


syntax/go.vim
```diff
@@ -410,7 +410,7 @@ hi def link     goDeclType          Keyword
 
 " Variable Assignments
 if g:go_highlight_variable_assignments != 0
-  syn match goVarAssign /\v[_.[:alnum:]]+(,\s*[_.[:alnum:]]+)*\ze(\s*\=)/
+  syn match goVarAssign /\v[_.[:alnum:]]+(,\s*[_.[:alnum:]]+)*\ze(\s*(\+|\-|\||\^|\*|\/|\%|\<\<|\>\>|\&|\&\^)?\=[^\=])/
```
Carpetsmoker: Using a bracketed list is probably a bit more readable than a subgroup with a lot of `|`s and `\`s?

For example (only tested minimally):

    \v[_.[:alnum:]]+(,\s*[_.[:alnum:]]+)*\ze(\s*[-+%<>!&|^*=]?\=[^\=])

bhcleek: That particular example wouldn't quite work, but it's close.  Just for completeness, the `!=` isn't a valid assignment, so there's no ned for `!` in the collection, and that example also doesn't account for `<<=`, `>>=`, and `$^=`.  RE: https://golang.org/ref/spec#Assignments.

However, I was able to reduce some of the branches to a collection.

bhcleek:
PTAL

PR #1468, opened 2017-09-21
Support matchit in Go templates

Carpetsmoker:
So you can use `%` go from `{{if ..}}` to `{{end}}`.

Almost works well,`{{else if ...}}` parts are still causing problems though since it gets confused with `{{if ...}}`. My brain has turned to jelly from 40 minutes of Vim regexps.


Carpetsmoker:
I give up getting this to work with both Go template tags and HTML tags :-(

PR #1469, opened 2017-09-21
Documentation for syntax highlighting

### LINKED ISSUE #1403: No documentation for gotexttmpl or gohtmltmpl filetypes
https://github.com/cespare/vim-go-templates says it was merged into vim-go, but there is no documentation in `:help vim-go`.  Searching for templates only returns the skeleton-file-on-new-file feature.

I expected there to be at least some minimal documentation on how to use these filetypes.

Carpetsmoker:
Also moves all of this to a new section. Fix #1403

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1765,8 +1765,10 @@ binaries).
 
 Vim becomes slow while editing Go files~
 
-Don't enable the |'g:go_highlight_types'| and |'g:go_highlight_operators'|
-options.
+This is usually caused by `g:go_hilight_*` options. Try disabling them if
+you've enabled some of them.
+
+Other common culprits are |'g:go_auto_sameids'| and |go#statusline#Show()|.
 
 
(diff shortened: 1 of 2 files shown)


doc/vim-go.txt
```diff
@@ -1762,11 +1765,9 @@ binaries).
 
 Vim becomes slow while editing Go files~
 
-Don't enable these options:
->
-  let g:go_highlight_types = 0
-  let g:go_highlight_operators = 0
-<
+Don't enable the |'g:go_highlight_types'| and |'g:go_highlight_operators'|
```
bhcleek: While we're at it, can the language be changed to a positive assertion?
> Disable the ...

Please include `g:go_highlight_assignments` in this list.

Carpetsmoker: I think there are actually a bunch of other things that can cause performance problems too; I've updated the entry.

bhcleek:
LGTM

PR #1471, opened 2017-09-21
Add go#util#Exec()

Carpetsmoker:
This is intended as a replacement for `go#util#System()`. The main
difference is that this will accept a list instead of a string as the
command, and that every item in the list will be automatically
shell-escaped.

This means you can just do:

	echo go#util#Exec(['printf', 'look "quotes" and spaces and stuff!'])

and not worry about escaping and stuff.

This approach makes much more sense IMHO, and is also what Go's os/exec
and Python subprocess do. It's especially useful with the jobs API,
since that *doesn't* need escaping, so it makes things easier all round.
Vim still needs to use the shell unfortunately; it can't run programs
directly.

I added a new function instead of modifying `System()` so we can migrate
slowly and carefully; double escaping will break stuff! I just converted
`FillStruct()` as an test/example.

While I'm making a new function anyway I also changed the signature a
bit to something that seems a bit easier to use. I can change that back
if people don't like it or prefer something else.

### DIFF
--- a/autoload/go/fillstruct.vim
+++ b/autoload/go/fillstruct.vim
@@ -1,22 +1,17 @@
 function! go#fillstruct#FillStruct() abort
-  let binpath = go#path#CheckBinPath('fillstruct')
-  if empty(binpath)
-    return
-  endif
-
-  let l:cmd = [binpath,
+  let l:cmd = ['fillstruct',
       \ '-file', bufname(''),
       \ '-offset', go#util#OffsetCursor()]
 
   " Read from stdin if modified.
   if &modified
     call add(l:cmd, '-modified')
-    let l:out = go#util#System(go#util#Shelljoin(l:cmd), go#util#archive())
+    let [l:out, l:err] = go#util#Exec(l:cmd, go#util#archive())
   else
-    let l:out = go#util#System(go#util#Shelljoin(l:cmd))
+    let [l:out, l:err] = go#util#Exec(l:cmd)
   endif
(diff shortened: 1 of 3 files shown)


bhcleek:
I like this :-) Exec'ing seems much more natural and will allow callers to focus on what's to be called instead of how to escape arguments for the shell.

I don't see how the signature changed. What am I missing?

There's a good amount of repeated code in `go#util#Exec` and `go#util#System`. Why not just have `go#util#Exec` called `go#util#System` after `go#util#Exec` shell escapes its arguments? Then once all external uses of `go#util#System` are removed we can reduce its scope `autoload/go/util.vim` by renaming it `s:System`.

Carpetsmoker:
> I don't see how the signature changed. What am I missing?

It now returns `[output, err]`, instead of just the output. The `err` is a boolean to indicate there was an error.

> Why not just have go#util#Exec called go#util#System after go#util#Exec shell escapes its arguments?

I originally did this, but I moved it to the current solution because this effectively deprecates `System()`, and is something that should be removed in the not-so-distant future. Having the code independent means we avoid refactoring when we finished moving everything over.

bhcleek:
Keeping the functions separate will allow for better focus within the function; each function can do one thing and do it well. The only refactor when we're finished would be to rename the function....

Carpetsmoker:
Alright @bhcleek; I've refactored per your suggestion. Let me know what you think.

PR #1474, opened 2017-09-24
ensure fmt list gets closed when title cannot be checked

### LINKED ISSUE #1423: Quickfix window is no longer closed on Vim 7.4
After 9a433e5db072e767d10d7b66cf4807e1a10ba897 and the follow-up fix 52c5b1f74dade7ef21b34675751cd4a7b3ee2d03, the quickfix window is no longer closed in Vim 7.4.

To reproduce:

1. Open some go file.
2. Make it invalid by introducing a syntax error.
3. Save, so that a gofmt error appears in the quickfix window.
4. Fix the syntax error.
5. Save again, and observe that the quickfix window doesn't go away.

The particular vim I was using gave this for `vim --version`:

```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Nov 24 2016 16:44:48)
Included patches: 1-1689
Extra patches: 8.0.0056
<lots of other junk>
```

The particular behaviors I observed were:

1. Before 9a433e5db072e767d10d7b66cf4807e1a10ba897: the quickfix window is closed when I expect.
2. After 9a433e5db072e767d10d7b66cf4807e1a10ba897: there's some crash regarding getqflist when the quickfix window is supposed to be closed.
3. After 52c5b1f74dade7ef21b34675751cd4a7b3ee2d03: there's no crash, but the quickfix window isn't closed.

I upgraded to vim 8.0.1033 and then this fixed the problem (now the quickfix window is closed as expected).

So basically, I think that 9a433e5db072e767d10d7b66cf4807e1a10ba897 needs to be adjusted for vim 7 compatibility.

bhcleek:
Fixes #1423

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -127,19 +127,22 @@ function! go#fmt#update_file(source, target)
   let &fileformat = old_fileformat
   let &syntax = &syntax
 
+  let l:listtype = go#list#Type("GoFmt")
+
   " the title information was introduced with 7.4-2200
   " https://github.com/vim/vim/commit/d823fa910cca43fec3c31c030ee908a14c272640
-  if !has('patch-7.4-2200')
-    return
+  if has('patch-7.4-2200')
+    " clean up previous list
+    if l:listtype == "quickfix"
+      let l:list_title = getqflist({'title': 1})
+    else
+      let l:list_title = getloclist(0, {'title': 1})
+    endif
+  else
+    " can't check the title, so assume that the list was for go fmt.
+    let l:list_title = {'title': 'Format'}
   endif
 
-  " clean up previous list
-  let l:listtype = go#list#Type("GoFmt")
-  if l:listtype == "quickfix"
-    let l:list_title = getqflist({'title': 1})
-  else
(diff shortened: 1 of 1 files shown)


autoload/go/fmt.vim
```diff
@@ -129,17 +129,19 @@ function! go#fmt#update_file(source, target)
...
+    let l:listtype = go#list#Type("GoFmt")
+    if l:listtype == "quickfix"
+      let l:list_title = getqflist({'title': 1})
+    else
+      let l:list_title = getloclist(0, {'title': 1})
+    endif
   else
-    let l:list_title = getloclist(0, {'title': 1})
+    " can't check the title, so assume that the list was for go fmt.
+    let l:list_title = {'title': 'Format'}
```
jthurman42: This doesn't work if you you do not have `patch-7.4-2200`, as `l:listtype` is unset, and is referenced on lines 146 / 147 below.

Add: `let l:listtype = "locationlist"` here to resolve the issue.

bhcleek: Thanks @jthurman42. It should be resolved now.

PR #1476, opened 2017-09-25
Better testing

Carpetsmoker:
- Support multiple Vim versions (currently 7.4, 8.0, and Neovim).

- Make sure that we always run Vim from a temporary installation in
  `/tmp/vim-go-test` without loading `~/.vim`. This makes it a lot
  easier to run on people's computers.

- Also add a handy `./script/run-vim` script to run the installed Vim
  from the temp directory. Useful for testing, debugging, etc. without a
  (potentially large) ~/.vim/ dir.

- Previously the tests weren't actually being run correctly; see:
  https://travis-ci.org/fatih/vim-go/builds/279277579

- Format the output of the tests a wee bit nicer, roughly similar to the
  `go test` output.

- Expand docs on testing a bit.

I also attempted to integrate code coverage support with
https://github.com/Vimjas/covimerage (which was the reason I started
working on this), but couldn't really get that to work. Need to look in
to that later.


scripts/install-vim
```diff
@@ -0,0 +1,92 @@
...
+# This should work on both Travis and people's desktop computers, and be 100%
+# independent from any system installed Vim.
+#
+# It will echo the full path to a Vim binary, e.g.:
+#   /some/path/src/vim
+
+set -euC
+cd "$(dirname "$0")"
+
+vimgodir="$(dirname "$(readlink -f .)")" # TODO: doesn't work on OSX I think.
```
fatih: Can you add an if clause that exits the script when run on OSX (checking with uname for example). I'm using macOS and don't want to blow my setup because of this. We can later add homebrew or a different kind of installation step here.

scripts/test
```diff
@@ -0,0 +1,75 @@
...
+
+vim=$1
+vimdir=$(echo "$out" | tail -n1)
+vimgo="$(dirname "$(readlink -f .)")" # TODO: doesn't work on OSX I think.
+
+# Make sure all Go tools and other dependencies are installed.
+./run-vim $vim +':silent :GoInstallBinaries' +':qa'
+[ -d "$vimdir/share/vim/vimgo/pack/vim-go/start/vim-vimhelplint" ] || \
+  git clone --quiet https://github.com/machakann/vim-vimhelplint \
+    "$vimdir/share/vim/vimgo/pack/vim-go/start/vim-vimhelplint"
```
fatih: Don't think these should be part of the test. Without this I should still run the tests on macOS.

scripts/install-vim
Carpetsmoker: I just need to fix it; `readlink -f` doesn't work on OSX but there's a workaround (I just couldn't remember what it was exactly yesterday).

scripts/test
Carpetsmoker: I'm afraid that I don't quite follow your comment. What do you mean with "this"? Cloning vimhelplint or the `:GoInstallBinaries` (or both?)

fatih: This is the whole section above that installs Vim on the machine. Test should just run the tests, it shouldn't try to install anything every time I execute it.

Carpetsmoker: Okay; how would you do it? Have the script fail if the vim in `/tmp/vim-go-test/...` doesn't exist?

Carpetsmoker: One of the reasons I like this approach is that it's very easy to run tests: just run `./bin/test vim-7.4` and presto! This makes it easy for people to actually start *writing* some tests.

It will cache the results so only the first time takes a few minutes, and outside of the `:GoInstallBinaries` it shouldn't modify anything outside of the `/tmp/vim-go/test/` directory.

Maybe we could split it up in to `./bin/setup` and `./bin/run-test` or some such? Personally I don't really care as such, as long as running tests remains easy and "fool-proof" with just a few commands.

fatih: I'm in favor of splitting it. Maybe `./bin/install` is a better name. We have makefile where we can combine both.

Carpetsmoker: Alright, I'll work on that later :+1:

bhcleek: Perhaps we should run the tests in a Docker container so that the user's machine/environment is never polluted with what tests do?

scripts/install-vim
bhcleek: Is `$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" > /dev/null && pwd )` what you're thinking of @Carpetsmoker ?

scripts/test
Carpetsmoker: > Perhaps we should run the tests in a Docker container so that the user's machine/environment is never polluted with what tests do?

Yeah, could add that. It would largely be the same scripts, but with a Docker wrapper.
Personally I've had a lot of problems running Docker on my desktop machine and would prefer to avoid it.

Carpetsmoker:
Okay, I updated the PR with the requested changes. It seems to work well both on my Linux machine and Travis.

scripts/test
```diff
@@ -0,0 +1,75 @@
...
+if [ -f "/tmp/vim-go-test/FAILED" ]; then
+  echo 2>&1 "Some tests FAILED"
+  exit 1
+fi
+echo 2>&1 "All tests PASSED"
+
+
+### Run vimhelplint.
+####################
+# TODO: Doesn't work in nvim? Not sure why, just don't run it for now.
```
bhcleek: This doesn't seem to work in any Vim.... Every time I try, it's unable to find `VimhelpLintEcho`.

scripts/install-vim
```diff
@@ -0,0 +1,106 @@
...
+  ln -s "$vimgodir" "$installdir/share/vim/vimgo/pack/vim-go/start/vim-go"
+fi
+
+# Make sure all Go tools and other dependencies are installed.
+export GOPATH=$installdir
+export PATH=${GOPATH}/bin:$PATH
+"$vimgodir/scripts/run-vim" $vim +':silent :GoUpdateBinaries' +':qa'
+
+[ -d "$installdir/share/vim/vimgo/pack/vim-go/start/vim-vimhelplint" ] || \
+  git clone --quiet https://github.com/machakann/vim-vimhelplint \
```
bhcleek: There's no need for the full history; a shallow clone will be fine.... (e.g. `--depth 1`).

fatih:
Btw, before merging let me also know as I want to test it on my local Macbook (unless you both tested it) :)

bhcleek:
Yeah, I'm testing it on my Macbook and also in a Docker container. I'll put up a PR to this branch to bring the Dockerfile in....

bhcleek:
@fatih https://github.com/Carpetsmoker/vim-go/pull/1

scripts/install-vim
```diff
@@ -0,0 +1,106 @@
...
+# Install and setup a Vim or Neovim for running tests.
+# This should work on both Travis and people's desktop computers, and be 100%
+# independent from any system installed Vim.
+#
+# It will echo the full path to a Vim binary, e.g.:
+#   /some/path/src/vim
+
+set -euC
+
+vimgodir=$(cd "$(dirname "$0")/.." && pwd)
```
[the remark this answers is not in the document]
Carpetsmoker: Setting `cdpath` for non-interactive shells is probably not really a good idea btw. Pretty sure that more shell scripts will subtly break with that.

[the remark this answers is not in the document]
bhcleek: Agreed, but it is something that happens. I always think of it from the perspective of being a good script writer *and* being an environment author. Assume the worst, but adopt the best practice :-)

.travis.yml
```diff
@@ -1,21 +1,14 @@
...
-  cd vim
-  ./configure --prefix=$DEPS --with-features=huge --disable-gui
-  make
-  make install
-  cd -
-
-script: ./scripts/test.sh
+cache:
+  directories:
+    - /tmp/vim-go-test
```
bhcleek: How will this affect the builds when we decide to update to a newer patch version of the installed versions?

Carpetsmoker: The cache is per-pr, so we should be fine.

scripts/test
Carpetsmoker: Not sure how to fix that to be honest. Does it work for you in master?

I split out the linting stuff to a `./scripts/lint` script. That makes it look nicer in Travis as well, and makes sure the tests at least run.

.travis.yml
bhcleek: Travis docs say the cache for a branch/pr may come from master, too. Couldn't that be a problem for us?

Carpetsmoker: I don't know, but it's never has been for me in the past. But if it is, we can manually delete it in the Travis UI. Updating patch-levels is not going to be a super-frequent event I think.

Carpetsmoker: Could also add patchlevels to the paths in `/tmp/vim-go`. I think there was a reason I didn't, but I forgot what it was (been a few weeks).

bhcleek: We shouldn't have to go change Travis configuration because of a change in which Vim versions we support.

I understand the draw of caching this, but long term I think it will cause us to spend cycles debugging what we could solve now by not caching (though at the cost of longer build times).

I'll digress, but please consider (maybe even test to try to invalidate each of our understandings) the concerns I raised.

bhcleek:
Because of the nvim tarball, testing against nvim fails on Mac. That's mitigated by the Dockerfile, though. @fatih, is docker acceptable to you, or do you want the nvim tests to work on Mac?

doc/vim-go.txt
```diff
@@ -566,13 +566,13 @@ CTRL-t
                                                                  *:GoFreevars*
 :GoFreevars
 
-    Enumerates the free variables of the selection. “Free variables” is a
+    Enumerates the free variables of the selection. "Free variables" is a
```
Carpetsmoker: You're not smart quote hip :-(

bhcleek: I'm hip, but vimhelplint wasn't in Linux without some particular encodings installed.

.travis.yml
Carpetsmoker: I don't think it'll be an issue, but OTOH builds are already very fast so removing the cache isn't going to hurt us very much (so I've just removed it).

bhcleek:
LGTM

fatih:
This is great. Thanks @Carpetsmoker @bhcleek 

How do I run the tests locally now?  Do I have to install each Vim version via install-vim first?  When I call `make test` it fails as it can't find Vim (though I have it on my machine). I wonder what the preferred workflow here is.

Carpetsmoker:
Yeah, you'll have to run `install-vim [version]` first @fatih. If you use `make all` then it should do that for you for all three Vim versions (although running neovim tests doesn't work on macOS yet).

We can tweak the workflow a bit if we want.

> it fails as it can't find Vim (though I have it on my machine).

Tests are *always* run with a fresh Vim version now, and never with the version that may or may not be installed on your machine.

fatih:
Thanks @Carpetsmoker I'm flying in couple of days to Japan for VimConf. I'll take a look when I'm back to add a Docker image that makes it easy to run (for those who use macOS)

Carpetsmoker:
Billy already added a `Dockerfile` @fatih; I also wrote a wrapper script to make that a bit easier to use, but that doesn't quite work for some reason (I'll open a PR in a bit).

fatih:
Nice! Seems like I've totally missed it 🤦‍♂️

bhcleek:
@fatih to test in a clean environment, just:
```
docker build -t vim-go-test .
docker run --rm vim-go-test
```

PR #1488, opened 2017-09-30
[GoDoc] Fix: don not resize GoDoc window if its still visible

nk2ge5k:
Hello. First of all sorry for my english.

I am encounter an issue.

The steps to recreate it:
1. Call `:GoDoc os`
2. Move GoDoc window to the right
3. And then call `:GoDoc io`

Result looks like this:
vim


Carpetsmoker:
This behaviour seems to make sense, thanks! :+1:

PR #1495, opened 2017-10-05
Clarify documentation for g:go_play_browser_command

### LINKED ISSUE #1487: vim go default browser
### Behavior

I try to launch every vim-go related command to a browser different to the default one. First I used : 
`let g:go_play_browser_command = 'chromium'`

There are two problems with that : 
1. If chromium is not already opened, it will froze neovim
2. It does not pass the adress to chromium correctly, and instead display the default starting page

To avoid problem 1, I can do that instead : 
`let g:go_play_browser_command = '(chromium & &> /dev/null)'`

So that chromium runs in its own subshell when launched.
But problem n° 2 is still there.

Is there a way to pass the correct address to chromium without having to manually paste it in the address bar ?

### Steps to reproduce:

Modify the default `g:go_play_browser_command` value, and try to use the `:GoDocBrowser` or `:GoPlay` command

### Configuration

OS : Linux (Archlinux)
Window Manager : i3

vim-go version : 1.14
go version : 1.9

vim version : 8.0
neovim version : 0.2.0

Carpetsmoker:
It never mentioned that you need to use `%URL%` or that you're expected
to return the shell yourself.

While I'm here also add `chromium` to list (the binary for Chrome proper
seems to differ a lot per Linux distro) and move the macOS detection to
a function.

Fixes #1487

### DIFF
--- a/autoload/go/play.vim
+++ b/autoload/go/play.vim
@@ -75,14 +75,16 @@ endfunction
 function! s:get_browser_command() abort
   let go_play_browser_command = get(g:, 'go_play_browser_command', '')
   if go_play_browser_command == ''
-    if has('win32') || has('win64')
+    if go#util#IsWin()
       let go_play_browser_command = '!start rundll32 url.dll,FileProtocolHandler %URL%'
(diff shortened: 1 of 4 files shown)


bhcleek:
LGTM

PR #1497, opened 2017-10-06
 Suppress formatting error messages on quitting

igor-mikushkin:
Please note that I am absolute vim noob, so these changes may be far from perfect.
The issue I experience is following.

Steps to reproduce:
1. Open go file in vim,
2. Make it invalid by introducing a syntax error,
3. `:wq`.

This results in opened window with errors.
But this window is totally unusable.
If you try to type a command it is shown just over a text, etc.

So I'm just suppressing this window.
Please note that it is implemented only for go files not asm (I didn't check the issue for them).

Environment:
Vi IMproved 8.0 (Included patches: 1-95),
Ubuntu 17.04.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -75,7 +75,7 @@ function! go#fmt#Format(withGoimport) abort
 
   if go#util#ShellError() == 0
     call go#fmt#update_file(l:tmpname, expand('%'))
-  elseif g:go_fmt_fail_silently == 0
+  elseif g:go_fmt_fail_silently == 0 && !(exists('g:go_fmt_suppress_errors_quitting') && g:go_fmt_suppress_errors_quitting == 1)
     let errors = s:parse_errors(expand('%'), out)
     call s:show_errors(errors)
   endif
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -188,6 +188,11 @@ function! s:auto_sameids()
   endif
 endfunction
 
+function! s:fmt_quit()
+  " Setting a flag for suppressing formatting errors on quitting
+  let g:go_fmt_suppress_errors_quitting = 1
+endfunction
+
 function! s:fmt_autosave()
   " Go code formatting on save
   if get(g:, "go_fmt_autosave", 1)
@@ -228,6 +233,7 @@ augroup vim-go
     autocmd CompleteDone *.go call s:echo_go_info()
   endif
 
+  autocmd QuitPre *.go call s:fmt_quit()
   autocmd BufWritePre *.go call s:fmt_autosave()
   autocmd BufWritePre *.s call s:asmfmt_autosave()
   autocmd BufWritePost *.go call s:metalinter_autosave()


bhcleek:
Thank you for the PR @igor-mikushkin . I'm reluctant to merge this; it introduces an inconsistency in the way formatting on write works depending on whether you're quitting or not. If you want to quit even if the write fails, then you `:wq!` will do the job for you natively already.

@fatih @Carpetsmoker what do you think?

Carpetsmoker:
Using `:wq` *does* kinda of brick your Vim at the moment if you save a buffer with compile errors, so *something* should be fixed. Never noticed this behaviour before, so I think it's new(-ish)?

I agree that this patch is probably not the best way to fix it.

bhcleek:
I don't see it being any different, really, than trying to write to a readonly directory. If you really don't care about errors, then `:wq!` is the right thing to do.

Carpetsmoker:
Sorry, I should have clarified; the problem isn't really about seeing errors, it's that `:wq` won't actually quit Vim, and leaves the terminal in a very weird state where input isn't properly echo'd etc. You have to "blindly" type `:wq!` again to actually quit Vim.

Maybe it only happens with a certain combination of settings? Didn't investigate too deeply yet, but can reproduce with my settings, and there seems to be a bug here.

bhcleek:
Oh, I see. Are you ending up with a quickfix window only?

igor-mikushkin:
You are ending up with a locationlist window only. The file you was editing is closed. So you should type blindly `:wq!` or just `:q`. There is no difference at this point.

Carpetsmoker:
I was looking at how Syntastic fixes this problem, and it turns out it doesn't; it has the same issue when you use:

```
let g:syntastic_auto_loc_list = 1
let g:syntastic_check_on_wq = 1
let g:syntastic_always_populate_loc_list = 1
```

Carpetsmoker:
There are two issues here:

1. Vim is left in an unusable state.
2. Vim-go happily exits the buffer on `:wq` even if gofmt has errors, which it
   probably shouldn't.

The first issue is a bug in Vim, not Vim-go. I've made a patch to fix that: https://github.com/vim/vim/pull/2205

The second issue is a bit trickier, since Vim doesn't have a way to abort the `:wq` command from the `BufWritePre` auto command. As far as I know, the only way to do that is to hook in to the `BufWriteCmd` autocommand. Doing that can be a bit tricky, as getting it wrong may result in the buffer not being written at all(!)

For now, the best solution may be to `echo` the errors if Vim is quitting with
`go#util#EchoError()` instead of using the location list? That way people still
get to see the errors, but it won't cause any confusing behaviour. We can make
an issue for aborting the write on gofmt errors.

bhcleek:
@Carpetsmoker good find in Vim!

 I propose that we close this PR without merging; it has a wider scope than just the `:GoFmt` case.

But I don't understand why you say Vim is bricked and unusable. When I try doing `:wq` on a file that is missing a curly brace:
```
package main
import "fmt"

func main() {
    fmt.Println("bricked?")
```

I'm left with only the location list, but Vim is still usable, and I can get back to the file easily with `:e #` or `ctrl-^`, which isn't a state that I would describe as bricked or unusable. Is your experience different?

Carpetsmoker:
Do you use a GUI @bhcleek? It's only a problem with terminal Vim; [`not_exiting()`](https://github.com/vim/vim/blob/2c33d7bb69c4c2c5b0e39b03cc4b0c04cfdfbb0b/src/ex_docmd.c#L7188) sets `settmode(TMODE_RAW);`, so Vim left the terminal in the input behaviour (`TMODE_COOK` or `TMODE_SLEEP`).

In GUIs you don't have this mucking about with terminal modes, so not calling `not_exiting()` is not a big deal.

Carpetsmoker:
Here's a gif btw (GitHub's upload feature seems broken at the moment :-/) http://tmp.arp242.net/a.gif

Carpetsmoker:
I've opened #1511 with a possible different fix. I'll close this PR.

Thanks for reporting the issue!

Issue #1483, opened 2017-09-28
GoBuild fails with E714: List required


turkenh:
### Behavior

Running :GoBuild with simple [main.go](https://github.com/fatih/vim-go-tutorial/blob/master/main.go) in the tutorial errors instead of success.

### Steps to reproduce:

1. Just follow the [tutorial](https://github.com/fatih/vim-go-tutorial)
2. Run :GoBuild as in [build-it step](https://github.com/fatih/vim-go-tutorial#build-it)

### Configuration

Add here your current configuration and additional information that might be
useful, such as:

* `vimrc` you used to reproduce
call plug#begin()
Plug 'fatih/vim-go', { 'do': ':GoInstallBinaries' }
call plug#end()

* vim version: 
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Apr  4 2017 18:14:54)
Included patches: 1-898, 8056
Compiled by root@apple.com

* vim-go version: master branch
* go version: go version go1.8.3 darwin/amd64

Only occurs from tip of master, not reproducible in v1.14
Instructing to use latest stable version in the tutorial might be a good idea.

Carpetsmoker:
Strict type checking was removed in Vim 7.4.1546, but you're using Vim 7.4.898, which is almost two years old.

Easier way to fix it is to update Vim to a newer version. The lowest version I test stuff with is Vim 7.4.1689 (but I would recommend you use Vim 8).

Carpetsmoker:
Fixed with #1498.

I would strongly recommend updating to a newer Vim version though.

PR #1498, opened 2017-10-06
Fix util echo function: ensure l:msg is a list after split()

kleww:
There was a bug on `GoBuild` with vim74. The variable l:msg in util echo function is a string even after calling split() that causes a crash on the map() call.
Issue #1483

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -287,9 +287,11 @@ endfunction
 "
 " The message can be a list or string; every line with be :echomsg'd separately.
 function! s:echo(msg, hi)
-  let l:msg = a:msg
-  if type(l:msg) != type([])
-    let l:msg = split(l:msg, "\n")
+  let l:msg = []
+  if type(a:msg) != type([])
+    let l:msg = split(a:msg, "\n")
+  else
+    let l:msg = a:msg
   endif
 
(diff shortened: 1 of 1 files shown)


bhcleek:
@kleww thank you for the contributation. #1483 was filed against an old version of Vim. What is the lowest version of Vim where you've seen `l:msg` be a string instead of list?

kleww:
I have seen this bug on this vim version:

> VIM - Vi IMproved 7.4
> Included patches: 1-488, 576
> Extra patches: 8.0.0322, 8.0.0056

I know it's a bit old but I'm stuck with it at work... I didn't encounter any problem on last vim version.
I am not an expert at all in vimL but currently:

```
function! s:echo(msg, hi)
 let l:msg = a:msg
    if type(l:msg) != type([])
    let l:msg = split(l:msg, "\n")
endif
```
If `a:msg` is a string `l:msg` will be a string then a list (which old vim version doesn't allow), isn't this could be error-prone ?

Carpetsmoker:
Thanks, seems to work well :+1:

I would strongly recommend updating to a newer Vim version though. You should be able to just install it to `~/vim` or some such, don't need root permissions (`./configure --prefix=$HOME/vim`). This is what I used to do when I was stuck on some ancient CentOS box years ago :-)

PR #1502, opened 2017-10-08
edit issue template

bhcleek:
Edit the issue template to more clearly walk issue submitters through
the process of providing a good report.

* Clearly prompt the submitter for what they did, what they expeted, and
  what they saw.
* Added an entry to the issue template configuration section to prompt
  submitters for their operating system.
* Added a parenthetical to let submitters know how to get their current
  Go version.
* Moved the vimrc prompt to the end of the configuration list since it
  is usually larger and spans multiple lines.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,25 +1,15 @@
-### Behavior
+### What did you do? (required. The issue will be **closed** when not provided.)
 
-Write here what's happening and what you're expecting instead of...
+Please provide a recipe to replicate your problem with a minimal `vimrc` with all plugins other than vim-go disabled.
 
-### Steps to reproduce:
+### What did you expect to happen?
 
-Please create a reproducible case of your problem. If this step is
-not provided, the issue will be **closed**
-
-Re produce it with a minimal `vimrc` with all plugins disabled and
-only `vim-go` enabled:
-
-1.
-2.
-3.
+### What happened instead?
 
 ### Configuration
 
-Add here your current configuration and additional information that might be
-useful, such as:
-
-* `vimrc` you used to reproduce
-* vim version:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,16 @@
...
-3.
+### What happened instead?
 
 ### Configuration
 
-Add here your current configuration and additional information that might be
-useful, such as:
-
-* `vimrc` you used to reproduce
 * vim version:
```
Carpetsmoker: Maybe mention that this includes the first *two* lines from `:version`? I've seen some people post their version just as the first line, which includes only the version number (e.g. 7.4 or 8.0) and not the patch level (which is kinda important).

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,16 @@
-### Behavior
+### What did you do? (required. The issue will be **closed** when not provided.)
 
-Write here what's happening and what you're expecting instead of...
+If possible, please provide a recipe to replicate your problem with a minimal
+`vimrc` with all plugins other than vim-go disabled.
```
Carpetsmoker: I would personally prefer to drop the "If possible" part here. I don't know about you, but I don't feel like testing stuff with people's huge ~/.vim directories and 37 plugins. Besides, debugging vimrc files is as easy as commenting stuff out, and something anyone can do.

For example:

> Please make sure you're able to replicate your problem with a minimal `vimrc` with all plugins other than vim-go disabled.

Could maybe also link to [this](https://vi.stackexchange.com/q/2003/51); I think there's also a help page with more or less the same info, but I can't seem to find that at the moment.

bhcleek:
All good suggestions. Thanks @Carpetsmoker.

PTAL

PR #1504, opened 2017-10-08
edit documentation

bhcleek:
Edit the INTRO and INSTALL sections of the documentation. Edits mostly
consist of rewording some sentences and paragraphs to read more
naturally.

Consistently capitalize vim-go, even if it's the first word in a
sentence. This might not seem right to you at first, but think about how
other proper nouns that begin with a lower case letter (e.g.  iPhone)
are capitalized when used as the the first word in a sentence.

Make analogous edits to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,11 @@ This plugin adds Go language support for Vim, with the following main features:
 
 ## Install
 
-Master branch is a **development** branch. Please use with caution.
-I recommend to use the [**latest stable release**](https://github.com/fatih/vim-go/releases/latest)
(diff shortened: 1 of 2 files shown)


fatih:
SGTM. But seems like vimhelplint is not happy:

```
 ../doc/vim-go.txt:35:1:Warning:1:The width of a line should be no longer than 78.
```

Also, not related to this PR, we need to fix the test for `fmt_test.vim`.

bhcleek:
Thank you for the reminder about vimhelplint. I fixed and amended the commit.

PR #1511, opened 2017-10-14
Don't open the {quickfix,location}list when quitting

Carpetsmoker:
In Vim versions before [8.0.1190][1] it would open the new window *and*
leave Vim in the wrong terminal mode (making it unusable).

In later Vim versions it would close the window with the Go file, only
to have a new list popup. Whether this is the desired behaviour is a
matter of opinion, but considering the above bug I think it's better to
*not* open the list for now and simple echo any errors.

This continues from #1497.

[1]: https://github.com/vim/vim/commit/2c33d7bb69c4c2c5b0e39b03cc4b0c04cfdfbb0b

### DIFF
--- a/autoload/go/list.vim
+++ b/autoload/go/list.vim
@@ -12,6 +12,11 @@ endif
 " If no or zero height is given it closes the window by default.  
 " To prevent this, set g:go_list_autoclose = 0
 function! go#list#Window(listtype, ...) abort
+  " Don't open the list when quiting; prevents orphaned window and Vim bug #1497.
+  if exists('g:_go_quiting') && g:_go_quiting
+    return
+  endif
+
   " we don't use lwindow to close the location list as we need also the
   " ability to resize the window. So, we are going to use lopen and lclose
   " for a better user experience. If the number of errors in a current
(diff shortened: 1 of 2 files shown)


bhcleek:
When I test this with a single window open, the error messages go by quickly, and then the window closes without a chance for me to even read them. I don't think that's what you intended, though.... (note that when I have multiple windows open and I `:wq` in the window with a syntax error, I do see the messages). Perhaps you're seeing slightly different behavior because you're using your 1190 patch?

 I agree that this solution will be better than the current state of things, but I think the echoing should happen in `go#list#Window` instead of when the list is populated so that the separation of when to display the window and what to put into the list to be displayed remain separated. `getqflist()` and `getlocllist()` will give you the lists....

Instead of a global variable to indicate that the window is quitting, a window variable is probably more appropriate so other windows are not also forced into this behavior.

@Carpetsmoker RE: your responses to my most recent comments on #1497: I don't use a GUI.  I suspect we're just dithering on what the threshold for "unusable" is... Here's what happens on my machine:

image

Carpetsmoker:
After thinking about this some more, I don't think this is a good approach after all.

- It won't work for people with GUIs; the messages will be lost for them.
- It may not even work for all people in terminals, as some clear the screen output when Vim exits.

I also can't think of a better way to do this, other than making `:wq` not quit when gofmt has errors (as [mentioned here](https://github.com/fatih/vim-go/pull/1497#issuecomment-336607860)).

PR #1516, opened 2017-10-17
short circuit CursorHold event commands in fugitive buffers

### LINKED ISSUE #1399: go#template#create error when using vim-fugitive :Gdiff
### Behavior

See below.

### Steps to reproduce:

* vim-fugitive commit https://github.com/tpope/vim-fugitive/commit/913fff1cea3aa1a08a360a494fa05555e59147f5
* vim-go commit 76cd99db6a88e825f361df0043cbff777c4a14fb
* `NVIM v0.2.1-635-geb40b7ec4`
* `go version go1.9beta2 linux/amd64`

1. Open a `.go` file in a git controlled repo
2. Run `:Gdiff` to view the differences in that file between working tree and staged version
3. See the following error:

```
Error detected while processing function <SNR>45_template_autocreate[3]..go#template#create:
line    6:
E472: Command failed
```

and the side-by-side diff fails to show.

Instead we should see the normal `:Gdiff` output.

This error goes away if `g:go_template_autocreate = 0` (but then so does the templating of new `.go` files)

### Configuration

As above.

bhcleek:
Do not try to show same id or type info when in fugitive buffers; doing
so causes errors, because the buffer filename is not a real file.

Fixes #1399


fatih:
Is there a way we can call these functions only in Go files (or filetypes)? That would automatically exclude fugitive and any other files. Maybe the fix is just simply moving them to ftplugin/go.vim ? (didn't tried it yet). This would be a better solution instead of adding custom logic for third party plugins.

bhcleek:
These functions *are* only called for Go files. Fugitive doesn't change the filetype at all, but it does use a filename that starts with `fugitive://`. This is really just a quickfix. There's probably a bigger issue where we need to handle guru errors better. I'll leave this open in case you want to merge it before I can get to better handling for the guru errors that occur when doing something like `:Gdiff` with `g:go_auto_sameids =1`.

bhcleek:
I'm closing for now, because I believe that there's a better solution and that the problems here can be worked around in two ways until we have a more holisitic fix: using autocmd events or simply temporarily disabling auto sameids while looking at diffs with fugitive.

PR #1519, opened 2017-10-18
do not try to run tests for :GoTestCompile

### LINKED ISSUE #1515: When ensuring if a test compiles, `TestMain()` and any `init()` functions are executed
When using `go#test#Test` to test if a test file compiles, it runs the test setup. This looks deliberate from the given [commit](https://github.com/fatih/vim-go/commit/21376e689d6b701b4367cf121da4cfcf17e5fa19)

If the tests are intended to run with some form of configuration, then the developer can not use the function with their tests.

cc @enocom

bhcleek:
Do not try to run a non-existent test for :GoTestCompile. Instead,
compile tests to a temporary file with `go test -c` in order to avoid
problems that may be caused by runtime errors in an init function when
trying to run a non-existent test.

To avoid the problems that 21376e689d6b701b4367cf121da4cfcf17e5fa19
resolved, this change relies on Vim's managing of the temporary
directory instead.

Fixes #1515

### DIFF
--- a/autoload/go/test.vim
+++ b/autoload/go/test.vim
@@ -6,9 +6,8 @@ function! go#test#Test(bang, compile, ...) abort
 
   " don't run the test, only compile it. Useful to capture and fix errors.
   if a:compile
-    " we're going to tell to run a test function that doesn't exist. This
-    " triggers a build of the test file itself but no tests will run.
-    call extend(args, ["-run", "499EE4A2-5C85-4D35-98FC-7377CD87F263"])
+    let testfile = tempname() . ".vim-go.test"
+    call extend(args, ["-c", "-o", testfile])
   endif
 
   if a:0


fatih:
What happens to testfile?  The change previously wouldn't create any binary. Imho if we decide to get rid of it (due the problem outlined in #1515) we should also delete the created `testfile`

bhcleek:
As it stands right now, the OS will clean up the file if/when it needs to, and Vim will also clean it up when Vim exits, but I can add back the callbacks to try to clean it up after the `go test -c` completes.

FWIW, I vascillated on whether to delete it or not, because it won't _for sure_  get deleted for the reasons outlined in `21376e689d6b701b4367cf121da4cfcf17e5fa19`, but we can give it a best effort; if the callback to delete doesn't get run, then Vim will clean up the file when it exits.

fatih:
I'm not sure how Vim deletes it if we quit it. Can you show the source code in vim-go (if it's done here) or the doc in Vim for it? I believe the binary will be still in CWD.

bhcleek:
Yeah, sorry about that; I had made some changes, then altered those just a bit to verify the file location, and then when I only undid a portion of the debugging changes, which resulted in a commit that wasn't what I originally intended.

I've amended the commit to show what I'd originally intended to submit. As you can see, the filename comes from `tempname()`, which provides a temporary filename. The directory that it's created it is deleted by Vim automatically when Vim exits. And of course, operating systems are free to clean up the temp directories as they see fit.

Let me know if you still want to proactively delete the file.

autoload/go/test.vim
```diff
@@ -6,9 +6,8 @@ function! go#test#Test(bang, compile, ...) abort
 
   " don't run the test, only compile it. Useful to capture and fix errors.
   if a:compile
-    " we're going to tell to run a test function that doesn't exist. This
-    " triggers a build of the test file itself but no tests will run.
-    call extend(args, ["-run", "499EE4A2-5C85-4D35-98FC-7377CD87F263"])
+    let testfile = tempname(). ".test"
```
fatih: can you change this to `.vim-go.test`. At least it wouldn't be a surprise for some if they somehow delve /tmp and see tens of these files

fatih:
lgtm, just a small comment.

PR #1521, opened 2017-10-18
update changelog

bhcleek:


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ BUG FIXES:
   [[GH-1498]](https://github.com/fatih/vim-go/pull/1498).
 * Don't resize godoc window if it's already visible
   [[GH-1488]](https://github.com/fatih/vim-go/pull/1488).
+* `:GoTestCompile` produces a test binary again. [[GH-1519]](https://github.com/fatih/vim-go/pull/1519)
```
fatih: I think we should add more context. People might expect it creates a binary in CWD, which is not the case. Maybe adding `... produces a test binary in tmp folder again` Feel free to add your own wording.

fatih:
lgtm 👍

Issue #1518, opened 2017-10-17
:GoSameIds "next" command


hjma29:
I use GoSameIDs often but hope to have a "next"/"previous" command so I can jump to next SameIDs quickly. From doc command, I don't see such function exists. Is there any way that I can jump to next SameID?

fatih:
I actually implemented something where you could use `*/#` to select the underlying identifier and jumpy via `n/N` You can try it by appending the following piece of code to the file `ftplugin/go.vim`:

```
if get(g:, "go_sameid_search_enabled", 0)
  autocmd FileType go nnoremap <buffer> <silent> * :<c-u>call Sameids_search(0)<CR>
  autocmd FileType go nnoremap <buffer> <silent> # :<c-u>call Sameids_search(1)<CR>
  autocmd FileType go nnoremap <buffer> <silent> n :<c-u>call Sameids_repeat(0)<CR>
  autocmd FileType go nnoremap <buffer> <silent> N :<c-u>call Sameids_repeat(1)<CR>
  autocmd FileType go cabbrev nohlsearch <C-r>=Sameids_nohlsearch()<CR>
endif

" " mode 0: next 1: prev
function! Sameids_repeat(mode)
  let matches = getmatches()
  if empty(matches)
    return
  endif
  let cur_offset = go#util#OffsetCursor()

  " reverse list to make it easy to find the prev occurrence
  if a:mode
   call reverse(matches)
  endif

  for m in matches
    if !has_key(m, "group")
      return
    endif

    if m.group != "goSameId"
      return
    endif

    let offset = go#util#Offset(m.pos1[0], m.pos1[1])

    if a:mode && cur_offset > offset
      call cursor(m.pos1[0], m.pos1[1])
      return
    elseif !a:mode && cur_offset < offset
      call cursor(m.pos1[0], m.pos1[1])
      return
    endif
  endfor

  " reached start/end, jump to the end/start
  let initial_match = matches[0]
  if !has_key(initial_match, "group")
    return
  endif

  if initial_match.group != "goSameId"
    return
  endif

  call cursor(initial_match.pos1[0], initial_match.pos1[1])
endfunction

function! Sameids_search(mode)
  call go#guru#SameIds()
  call Sameids_repeat(a:mode)
endfunction

function! Sameids_nohlsearch()
  call go#guru#ClearSameIds()
  return "nohlsearch"
endfunction
```

[...]

PR #1522, opened 2017-10-18
Basic version of GoNextId and GoPrevId commands #1518

kleww:
I started a basic version of what could be commands to jump between ids, I'm not sure about the command names or the way I manage to get the current id under the cursor so feedbacks are welcome.

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -483,6 +483,51 @@ function! go#guru#SameIds() abort
   call s:run_guru(args)
 endfunction
 
+function! go#guru#NextId() abort
+  let same_ids = getmatches()
+  if len(same_ids) < 2
+    return
+  endif
+
+  let curpos = getcurpos()
+  let current_idx = 0
+  for item in same_ids
+    let itempos = item['pos1']
+    if curpos[1] == itempos[0] && curpos[2] >= itempos[1] && curpos[2] < itempos[1] + itempos[2]
+      break
+    endif
+    let current_idx += 1
+  endfor
+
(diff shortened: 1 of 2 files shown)


fatih:
Hm not sure these are useful for us. Can you share a screencast/gif on how it look like? A better way would be having text objects, such as `w`, `b`, etc.. so we can could jump between id's instead of using commands. But before that please share a screencast so know how it looks like.

autoload/go/guru.vim
```diff
@@ -483,6 +483,51 @@ function! go#guru#SameIds() abort
...
+function! go#guru#NextId() abort
+  let same_ids = getmatches()
+  if len(same_ids) < 2
+    return
+  endif
+
+  let curpos = getcurpos()
+  let current_idx = 0
+  for item in same_ids
+    let itempos = item['pos1']
```
Carpetsmoker: Aside from Fatih's comment, you need to make sure you're looping only over `sameIds` match groups. People can have more matches defined from their vimrc or other plugins (check the `group` attribute).

For example, for me it will break because I have [this in my vimrc](https://github.com/Carpetsmoker/dotfiles/blob/master/vim/softwrap.vim):

    :call matchadd('SoftWrap', '\%82v')
    :call matchadd('HardWrap', '\%122v')

Carpetsmoker:
I make a general plugin to jump between matches added with `:matchaddpos()`: https://github.com/Carpetsmoker/matchjump.vim

I think that is probably a better approach than putting this in vim-go? Otherwise we can copy the code from there I suppose.

PR #1524, opened 2017-10-18
Display warning when people use an older Vim

Carpetsmoker:
This shows a warning when people use an older Vim version. You can still set
`g:go_no_version_warning` to override this, if people really want to continue. I
don't think we need to bother documenting that setting in the help page.

The version 7.4.1689 was chosen because that's what the latest Ubuntu LTS
(16.04) uses, which seems like a reasonable choice. I don't mind using an older
version though.

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -11,7 +11,7 @@ let g:go_loaded_install = 1
 " Version 7.4.1689 was chosen because that's what the most recent Ubuntu LTS
 " release (16.04) uses.
 if
-      \ !exists('g:go_no_version_warning') &&
+      \ get(g:, 'go_no_version_warning', 0) != 1 &&
       \ (v:version < 704 || (v:version == 704 && !has('patch1689')))
       \ && !has('nvim')
   echohl Error
@@ -19,7 +19,7 @@ if
   echom "Please update your Vim for the best vim-go experience."
   echom "If you really want to continue you can set this to make the error go away:"
   echom "    let g:go_no_version_warning = 1"
-  echom "Note that some features may error out or behave incorrect."
+  echom "Note that some features may error out or behave incorrectly."
   echom "Please do not report bugs unless you're using Vim 7.4.1689 or newer."
   echohl None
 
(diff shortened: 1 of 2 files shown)


plugin/go.vim
```diff
@@ -4,6 +4,29 @@ if exists("g:go_loaded_install")
...
+if
+      \ !exists('g:go_no_version_warning') &&
+      \ (v:version < 704 || (v:version == 704 && !has('patch1689')))
+      \ && !has('nvim')
+  echohl Error
+  echom "vim-go requires Vim 7.4.1689 or Neovim, but you're using an older version."
+  echom "Please update your Vim for the best vim-go experience."
+  echom "If you really want to continue you can set this to make the error go away:"
+  echom "    let g:go_no_version_warning = 1"
+  echom "Note that some features may error out or behave incorrect."
```
bhcleek: incorrectly instead of incorrect.

plugin/go.vim
```diff
@@ -4,6 +4,29 @@ if exists("g:go_loaded_install")
...
 let g:go_loaded_install = 1
 
+" Not using the has('patch-7.4.1689') syntax because that wasn't added until
+" 7.4.237, and we want to be sure this works for everyone (this is also why
+" we're not using utils#EchoError()).
+"
+" Version 7.4.1689 was chosen because that's what the most recent Ubuntu LTS
+" release (16.04) uses.
+if
+      \ !exists('g:go_no_version_warning') &&
```
bhcleek: I wouldn't be surprised is someone sets this to `0` to try to re-enable the warning. Maybe just check that it's not `1`?

nqv:
@Carpetsmoker @bhcleek 
Do we really need this feature at all? I'm on macOS 10.12.4 and this is my default vim:
```
$ /usr/bin/vim --version
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Feb  4 2017 00:48:24)
Included patches: 1-898, 8056
Compiled by root@apple.com
```

fatih:
@nqv yes we do. Unfortunately Vim is moving fast and with every release, old releases tend to work less. You have the option to disable it as you know, so I think that would be the best option. Thanks for your feedback :)

Carpetsmoker:
Adding to that, v1.15 **is** actually broken in versions before Vim 7.4.1546; see #1483.

PR #1525, opened 2017-10-19
Remove g:go_autodetect_gopath

Carpetsmoker:
Follow-up to #1461.

Needs a bit more testing.


autoload/go/complete.vim
```diff
@@ -19,13 +19,6 @@ function! s:gocodeCommand(cmd, preargs, args) abort
     return
   endif
 
-  " we might hit cache problems, as gocode doesn't handle well different
-  " GOPATHS: https://github.com/nsf/gocode/issues/239
-  let old_gopath = $GOPATH
-  let old_goroot = $GOROOT
-  let $GOPATH = go#path#Detect()
-  let $GOROOT = go#util#env("goroot")
```
Carpetsmoker: Not 100% sure about removing this particular case though. Need to carefully read that issue.

bhcleek:
I like getting rid of all the code that this eliminates, and I agree that setting $GOPATH using events is better. But I'm not sure we should break people like this. Why not go ahead and add support for the events?

We'd need to handle a number of events:
* BufNewFile
* BufReadPre
* BufFilePre
* BufEnter

Carpetsmoker:
Wouldn't just the `FileType` autocommand be enough?

bhcleek:
No, because environment variable are global within a Vim session. If two files with different detected gopaths were open, the last one to be opened would set the gopath that would take effect when running commands in the first unless $GOPATH were set when changing windows and buffers.

Carpetsmoker:
Oh, right. That makes sense :+1:

autoload/go/complete.vim
fatih: this is needed for Go upgrades. Unfortunately `gocode` embeds the goroot into the binary instead of looking from env, after a Go upgrade `gocode` can be broken if the user has a `version` based path. i.e: `/usr/lib/go/1.9.1/src`.

fatih: We should keep the goroot behavior but remove the gopath change.

Carpetsmoker:
Okay, I added it back as autocommands.

I think just the `BufEnter` and `BufLeave` should be enough? That seems to work well in my tests anyway.

autoload/go/complete.vim
```diff
@@ -19,25 +19,25 @@ function! s:gocodeCommand(cmd, preargs, args) abort
...
+    let socket_type = get(g:, 'go_gocode_socket_type', s:sock_type)
+    let cmd = printf('%s -sock %s %s %s %s',
+          \ go#util#Shellescape(bin_path),
+          \ socket_type,
+          \ join(a:preargs),
+          \ go#util#Shellescape(a:cmd),
+          \ join(a:args)
+          \ )
+
+    let result = go#util#System(cmd)
```
bhcleek: This seems like a good time to convert this to `go#util#Exec`, but I won't object if you don't want to do it now.

Carpetsmoker: I'd rather not, since much of this isn't tested and can introduce regressions. Unless we're very careful we may have issues with double shell quoting for example.

I want to add a test for every tool which tests stuff like filenames with spaces, directories that don't exist, etc. and also fix whatever is needed (add build tags, use `#Exec()`, and whatever else might be needed). This is why I made the changes to the testing script and removed all the GOPATH detection.

Carpetsmoker:
> I've started looking at this and testing it. I want to spend a bit more time with it, especially making sure that the autocmd events are complete.

Don't want to push you, but did you have a chance to look at this?

I tested with a whole bunch of different scenarios, and it seemed to work well in all of them.

bhcleek:
Not yet. I should be able to get to it by next week (though probably this weekend).

plugin/go.vim
```diff
@@ -260,6 +260,16 @@ augroup vim-go
...
+  autocmd BufEnter *.go
+        \  if get(g:, 'go_autodetect_gopath', 0)
+        \|   let b:old_gopath = $GOPATH
+        \|   let $GOPATH = go#path#Detect()
+        \| endif
+  autocmd BufLeave *.go
+        \  if get(g:, 'go_autodetect_gopath', 0)
+        \|   let $GOPATH = b:old_gopath
+        \| endif
+augroup end
```
bhcleek: In the rare case when a user changes `g:go_autotdetect_gopath`, `BufLeave` will not behave as expected. Instead of `if get(g: 'go_autodetect_gopath', 0)`, `if exists(b:old_gopath)` will guard restoring `GOPATH` correctly. We should also unset `b:old_gopath`...

plugin/go.vim
```diff
@@ -260,6 +260,16 @@ augroup vim-go
...
-augroup END
+
+  autocmd BufEnter *.go
+        \  if get(g:, 'go_autodetect_gopath', 0)
+        \|   let b:old_gopath = $GOPATH
+        \|   let $GOPATH = go#path#Detect()
+        \| endif
+  autocmd BufLeave *.go
+        \  if get(g:, 'go_autodetect_gopath', 0)
+        \|   let $GOPATH = b:old_gopath
```
bhcleek: (nit) When the old `GOPATH` was not set, restoring `GOPATH` to an empty string does not faithfully restore the state to what it was. As a pragmatic matter, all the go tooling will handle this case correctly, but users may have other scripts, etc. that do not. As I see it, vim-go should restore `GOPATH` to its state prior to entering the buffer.

plugin/go.vim
```diff
@@ -260,6 +260,16 @@ augroup vim-go
   " in the same window doesn't highlight the most recently matched
   " identifier's positions.
   autocmd BufWinEnter *.go call go#guru#ClearSameIds()
-augroup END
+
+  autocmd BufEnter *.go
```
bhcleek: I saw some cases where `BufEnter` gets fired twice, but I was unable to duplicate it when I tried to.

In any event, it caused me to realize that if vim fires `BufEnter` twice, the second event will cause `b:old_gopath` to be overwritten with the current `$GOPATH`, and then when the buffer is left, `BufLeave` cannot assign the original value of `GOPATH`.  Can we add protection for this case by only checking `g:go_autodetect_gopath` when `if !exists('b:old_gopath')`

plugin/go.vim
Carpetsmoker: Good point; I've updated the PR with this issue and the other small issues you found.

bhcleek:
Did you mean to duplicate the changelog message in `BACKWARDS INCOMPATIBILITIES` and `IMPROVEMENTS`?

Carpetsmoker:
Oops, no; that was just a merge error.

bhcleek:
Don't forget to `unlet b:old_gopath` in the `BufLeave` after restoring `$GOPATH, too.

And this still didn't handle the situation I mentioned in https://github.com/fatih/vim-go/pull/1525#discussion_r151852055

Carpetsmoker:
> Don't forget to unlet b:old_gopath in the BufLeave after restoring `$GOPATH, too.

Ah, right. Fixed.

> And this still didn't handle the situation I mentioned in #1525 (comment)

I believe it should, as it checks for `if get(b:, 'old_gopath', '') isnot ''`.

bhcleek:
> I believe it should, as it checks for if get(b:, 'old_gopath', '') isnot ''.

On my system, `GOPATH` is not set, so in `BufEnter`, `let b:old_gopath = $GOPATH` will set `b:old_gopath` to the empty string., then in `BufLeave`, instead of `unlet $GOPATH`, `GOPATH` is set to the empty string using `let $GOPATH = b:old_gopath`. Unfortunately, Vim doesn't seem to have a way to unset an environment variable: https://github.com/vim/vim/issues/1116.

Can you add a comment in the `BufLeave`, though, documenting this edge case?

Also, now that `unlet b:old_gopath` is used in `BufLeave`, its guard should use `if exists(b:old_gopath)`, otherwise for people whose GOPATH is the empty string (which will be everyone after the `BufLeave` executes after `BufEnter` where `GOPATH` was not set!), `GOPATH` will never be set to the empty string.

Carpetsmoker:
> On my system, GOPATH is not set, so in BufEnter, let b:old_gopath = $GOPATH will set b:old_gopath to the empty string
> GOPATH is set to the empty string using let $GOPATH = b:old_gopath

How would this happen? Right now the code is:

```
  autocmd BufLeave *.go
        \  if get(b:, 'old_gopath', '') isnot ''
        \|   let $GOPATH = b:old_gopath
        \| endif
```

So I don't see how it will ever set `$GOPATH` to an empty string?

bhcleek:
Load vim with GOPATH unset or null and enable GOPATH autodetection:
```
$> GOPATH= vim
:let g:go_autodetect_gopath = 1
```
Then edit two files:
```
:e  $HOME/src/gb-project/bar.go
:e $HOME/go/src/quux/foo.go
```

When bar.go was entered, its `b:old_gopath` was set to the empty string and `GOPATH` was set to the detected value. When bar.go was left, it should have restored GOPATH to the empty string, but did not, because `b:old_gopath` is empty. The mere existence of `b:old_gopath` is what should trigger the restoration of GOPATH, not the value of `b:old_gopath`.

Carpetsmoker:
Okay; I pushed an update which should handle all cases I think. We can't `unlet $GOPATH` but `exists('$GOPATH')` does work correct.

bhcleek:
LGTM, but setting `b:old_gopath` to -1 and checking for that value is extraneous.

Carpetsmoker:
> LGTM, but setting b:old_gopath to -1 and checking for that value is extraneous.

Without that it will run `let $GOPATH = go#path#Detect()` on a second `BufEnter` event as `exists('b:old_gopath')` is false. Not a huge deal, but better to not do it, I think.

PR #1527, opened 2017-10-20
eliminate errors trying to run a command in a directory that does not exist

### LINKED ISSUE #1399: go#template#create error when using vim-fugitive :Gdiff

bhcleek:
Fixes #1399

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -162,6 +162,18 @@ function! go#tool#FilterValids(items) abort
 endfunction
 
 function! go#tool#ExecuteInDir(cmd) abort
+  " Verify that the directory actually exists. If the directory does not
+  " exist, then assume that the a:cmd should not be executed. Callers expect
+  " to check v:shell_error (via go#util#ShellError()), so execute a command
+  " that will return an error as if a:cmd was run and exited with an error.
+  " This helps avoid errors when working with plugins that use virtual files
+  " that don't actually exist on the file system (e.g. vim-fugitive's
+  " GitDiff).
+  if !isdirectory(expand("%:p:h"))
+    let [out, err] = go#util#Exec(["false"])
+    return ''
+  endif
+
   let old_gopath = $GOPATH
   let old_goroot = $GOROOT
   let $GOPATH = go#path#Detect()
--- a/autoload/go/tool_test.vim
+++ b/autoload/go/tool_test.vim
@@ -1,0 +1,23 @@
+func! Test_ExecuteInDir() abort
+  let l:tmp = gotest#write_file('a/a.go', ['package a'])
+  try
+    let l:out = go#tool#ExecuteInDir("pwd")
+    call assert_equal(l:tmp . "/src/a\n", l:out)
+  finally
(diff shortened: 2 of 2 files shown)


autoload/go/tool.vim
```diff
@@ -162,6 +162,18 @@ function! go#tool#FilterValids(items) abort
...
 function! go#tool#ExecuteInDir(cmd) abort
+  " Verify that the directory actually exists. If the directory does not
+  " exist, then assume that the a:cmd should not be executed. Callers expect
+  " to check v:shell_error (via go#util.ShellError()), so execute a command
+  " that will return an error as if a:cmd was run and exited with an error.
+  " This helps avoid errors when working with plugins that use virtual files
+  " that don't actually exist on the file system (e.g. vim-fugitive's
+  " GitDiff).
+  if !isdirectory(expand("%:p:h"))
+    let out = go#util#System("exit 1")
```
Carpetsmoker: Could you use `false`, instead of `exit`? The former is in POSIX, the latter isn't. Also `exit` is usually a shell built-in, so this code depends on the shell.

You should also use `go#util#Exec()` for all new code, the `System()` function will be removed once all calls are rewritten to use that.

bhcleek:
@Carpetsmoker  PTAL

Carpetsmoker:
I just added pushed a basic test for this.

I do wonder if silently doing nothing is really the best behaviour though. For example `Errcheck` will now simply do nothing. I wonder if that's really the best thing to do?

Carpetsmoker:
Meh, also seems that Neovim behaves slightly different

bhcleek:
If `go#lint#Errcheck` doesn't check the `ShellError` or is hiding errors for some other reason, then that's a completely separate problem. This change merely makes sure to return early and avoid a bunch of useless, noisy messages when the directory that `go#util#ExecuteInDir` is supposed to execute in does not exist, but I don't see that behavior. I see a message, "package is not inside GOPATH src"

I'm working on a fix for the nvim behavior. Thanks for adding some tests!

bhcleek:
PTAL.

I'll rebase before merging to apply the fixups correctly.

bhcleek:
I'll rebase now and then merge.

PR #1538, opened 2017-10-28
check group in when toggling same ids

tmatias:
Currently, `ToggleSameIds` doesn't check the group of the match and, as I set up vim to highlight characters that go beyond column 80 (just find it a mostly helpful heuristic), when I open a file that contains longer lines, `ToggleSameIds` always try to clear highlighted ids, because `getmatches()` return all highlights and not just the `GoSameId` ones.

So, as already done in `ClearSameIds`, I think we should also test the group here.

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -548,11 +548,14 @@ function! go#guru#ClearSameIds() abort
 endfunction
 
 function! go#guru#ToggleSameIds() abort
-  if len(getmatches()) != 0
-    call go#guru#ClearSameIds()
-  else
-    call go#guru#SameIds()
-  endif
+  let m = getmatches()
+  for item in m
+    if item['group'] == 'goSameId'
+      call go#guru#ClearSameIds()
+      return
+    endif
+  endfor
+  call go#guru#SameIds()
 endfunction
 
 function! go#guru#AutoToogleSameIds() abort
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -534,10 +534,13 @@ function! s:same_ids_highlight(exit_val, output) abort
 endfunction
 
 function! go#guru#ClearSameIds() abort
+  let l:cleared = 0
+
   let m = getmatches()
   for item in m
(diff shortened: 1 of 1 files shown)


bhcleek:

`go#guru#ClearSameIds`, which is called by `go#guru#ToggleSameIds`, already checks the group and only clears matches that are in the `goSameId` group. See https://github.com/fatih/vim-go/blob/master/autoload/go/guru.vim#L537-L542.

tmatias:
@bhcleek you are correct about that. 

However, `ToggleSameIds` itself does not and so, if there are any matches, even if not in the `goSameId` group, it will always decide that it should call `GoClearSameIds` and therefore the toggling won't work as expected.

(see PR description :P)

bhcleek:
I see. In that case, instead of repeating the logic for iterating through the matches, `go#guru#ClearSameIds` should return a non-zero value when no matches were found and that value should be checked in `go#guru#ToggleSameIds`, which should call go#guru#SameIds` if no `goSameId` groups where cleared. Will you refactor to do that?

tmatias:
That sounds like a better approach. Thank you. 

I'll do that.

tmatias:
Returned a positive value for success and zero for the other case, but I can change that to return a negative value if it would make things better. (Not knowing much vimscript, it just seemed nicer.)

autoload/go/guru.vim
```diff
@@ -534,23 +534,26 @@ function! s:same_ids_highlight(exit_val, output) abort
...
 function! go#guru#ClearSameIds() abort
+  let l:cleared = 0
+
   let m = getmatches()
   for item in m
     if item['group'] == 'goSameId'
       call matchdelete(item['id'])
+      let l:cleared = 1
     endif
   endfor
```
bhcleek: If no matches were in the `goSameId` group, then return early here so that the `BufWinEnter` events won't be removed.

bhcleek:
Thank you!

bhcleek:
Thank you for contributing. I squashed your commits and merged them into master.

PR #1540, opened 2017-10-29
update CHANGELOG.md

bhcleek:


CHANGELOG.md
```diff
@@ -1025,3 +1028,7 @@ the community contributions).
 ## 0.0 (Mar 24, 2014)
 
 Initial commit: https://github.com/fatih/vim-go/commit/78c5caa82c111c50e9c219f222d65b07694f8f5a
+
+<!--
+ vim: et ts=2 sw=2
+-->
```
fatih: I think this should be removed?

bhcleek: I added it intentionally, because I find myself having to mess with some of the tab options every time I edit CHANGELOG.md. Because it's within an HTML comment, it won't be rendered when CHANGELOG.md is rendered to HTML (e.g. when viewing on GitHub).

PR #1546, opened 2017-10-30
add modeline to CHANGELOG.md

bhcleek:


Carpetsmoker:
I added an `.editorconfig` a while ago btw, it's nice be cause it also works for new files and not just existing ones :-)

PR #1548, opened 2017-10-30
Expand test framework

Carpetsmoker:
A lot of tests are in the form of:

1. Set up some buffer.
2. Run a vim-go command.
3. Check if the output matches what we expect.

This adds `autoload/gotest.vim`, which includes two helper functions to
make this easy:

- `gotest#loadFile()` sets up the buffer state and some other stuff such
  as GOPATH.
- `gotest#assert_buffer()` checks if the current buffer contents match
  what's being expected.

This should make writing many tests easier.

Unfortunately Vim's syntax makes this look a wee bit awkward; but
personally I prefer this method over external fixture files as it's
easier to see what the input and output is on a single screen.
We can use the 'fixture style' if people prefer that. Or another
alternative would be to use Python to make multi-line statements look
better (...but then tests depend on +python).

I added a `fillstruct_test.vim` as a test, and also converted some
existing tests to make sure this works well.

In addition, there are also some related changes:

- Make sure invalid syntax etc. fails the test.
- Make sure `runtest.vim` doesn't set any global variables (only `s:`).
- Support adding `abort` to test funtions.
- Reset GOPATH etc. after each test.


fatih:
Personally I like test fixtures better. There are sometimes cases where strings are better, but test fixtures allows us to express any kind of Go source code without fiddling with Vim syntax. I would say maybe we should create wrapper to make loading and writing the test fixtures to a temp file easier maybe? I just think we shouldn't abandon it.

Carpetsmoker:
Alright; I added `gotest#load_fixture()` and `gotest#assert_fixture()`, which allows easy loading and comparing to fixture files.

Carpetsmoker:
I'm going to merge this now so I can continue working on some other stuff; it doesn't change anything user-visible, so that should be fine.

If people want changes then I'll be happy to make them later.

PR #1549, opened 2017-10-30
Add scripts/docker-test

Carpetsmoker:
To make running tests in Docker easier.

### DIFF
--- a/scripts/docker-test
+++ b/scripts/docker-test
@@ -1,0 +1,24 @@
+#!/bin/sh
+#
+# Run all tests inside a Docker container
+#
+
+set -uC
+vimgodir=$(cd -P "$(dirname "$0")/.." > /dev/null && pwd)
+cd "$vimgodir"
+
+if [ $(docker container ls -a --filter name=vim-go-test | wc -l) -eq 1 ]; then
+    docker build --tag carpetsmoker/vim-go:test .
+    docker run --detach --name vim-go-test carpetsmoker/vim-go:test
+else
+    docker container start vim-go-test
+fi
+
+# Make sure we stop/destroy the container.
+cleanup() {
+  docker container stop vim-go-test
+}
+trap cleanup EXIT
+
+
+# vim:ts=2:sts=2:sw=2:et
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,3 +11,7 @@ trim_trailing_whitespace = true
 [*.go]
 indent_style = tab
 indent_size = 4
+
+[Makefile]
+indent_style = tab
+indent_size = 8
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,13 @@ lint:
 	@echo "==> Running linting tools"
 	@./scripts/lint vim-8.0
 
+docker:
+	@echo "==> Building/starting Docker container"
+	@./scripts/docker-test
+
 clean:
 	@echo "==> Cleaning /tmp/vim-go-test"
 	@rm -rf /tmp/vim-go-test
 
 
-.PHONY: all test install clean lint
+.PHONY: all test install clean lint docker


Carpetsmoker:
Doesn't quite work for me though:

```
[~/.vim/pack/plugins/start/vim-go](docker-test)% ./scripts/docker-test
Sending build context to Docker daemon  1.392MB
Step 1/7 : FROM golang:1.9.1
 ---> 99e596fc807e
Step 2/7 : RUN apt-get update -y &&   apt-get install -y build-essential curl git libncurses5-dev &&   apt-get clean &&   rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 ---> Using cache
 ---> 46ccb008089f
Step 3/7 : RUN useradd -ms /bin/bash -d /vim-go vim-go
 ---> Using cache
 ---> a6e92398a625
Step 4/7 : USER vim-go
 ---> Using cache
 ---> 721b1aa9cab4
Step 5/7 : WORKDIR /vim-go
 ---> Using cache
 ---> f79a3ff79a50
Step 6/7 : COPY . /vim-go/
 ---> 4e9ae7f0fd8f
Step 7/7 : ENTRYPOINT ["make"]
 ---> Running in d8bcc860383e
 ---> 7f370fb830f9
Removing intermediate container d8bcc860383e
Successfully built 7f370fb830f9
Successfully tagged carpetsmoker/vim-go:test
b01007aa51afa7c7d1b60aadc7da96ad240a02414a16de52cd94016a12afa312

vim-go-test
```

```
[~]% docker logs vim-go-test
==> Installing Vims
Cloning v7.4.1689 from https://github.com/vim/vim
Cloning into '/tmp/vim-go-test/vim-7.4-src'...
Makefile:4: recipe for target 'install' failed
make: *** [install] Terminated
```

Any ideas @bhcleek?

scripts/docker-test
```diff
@@ -0,0 +1,24 @@
...
+#
+# Run all tests inside a Docker container
+#
+
+set -uC
+vimgodir=$(cd -P "$(dirname "$0")/.." > /dev/null && pwd)
+cd "$vimgodir"
+
+if [ $(docker container ls -a --filter name=vim-go-test | wc -l) -eq 1 ]; then
+    docker build --tag carpetsmoker/vim-go:test .
```
bhcleek: Just tag this as `vim-go-test` instead of `carpetsmoker/vim-go:test`.

scripts/docker-test
```diff
@@ -0,0 +1,24 @@
...
+    docker run --detach --name vim-go-test carpetsmoker/vim-go:test
```
bhcleek: Don't `--detach` here, but do `--rm`.

scripts/docker-test
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+#
+# Run all tests inside a Docker container
+#
+
+set -uC
+vimgodir=$(cd -P "$(dirname "$0")/.." > /dev/null && pwd)
+cd "$vimgodir"
+
+if [ $(docker container ls -a --filter name=vim-go-test | wc -l) -eq 1 ]; then
```
bhcleek: This really shouldn't be guarded. Docker will cache the images and will only build layers as needed.

scripts/docker-test
```diff
@@ -0,0 +1,24 @@
...
+    docker build --tag carpetsmoker/vim-go:test .
+    docker run --detach --name vim-go-test carpetsmoker/vim-go:test
+else
```
bhcleek: This `else` block isn't needed.

scripts/docker-test
```diff
@@ -0,0 +1,24 @@
...
+    docker container start vim-go-test
+fi
+
+# Make sure we stop/destroy the container.
+cleanup() {
```
bhcleek: I don't think there's any need for cleanup here if you `docker run --rm` above.

scripts/docker-test
Carpetsmoker: But then we would have to destroy and re-create the container on every test run?

fatih:
While we're at. We need to somehow bake the `install` part into Dockerfile, otherwise every Docker run will install it, which is not efficient. Not something for this PR, just something to be noted.

bhcleek:
That's easy enough to do. I'll work up a PR and submit it.

scripts/docker-test
bhcleek: Yes, but creating a container from an image is cheap. What's your concern?

bhcleek:
@fatih PR to install the three Vim versions in the image instead of when the container is run: #1551

bhcleek:
@Carpetsmoker Here's an alternative implementation (it's basically all the things I suggested + `set -e`): #1550.

scripts/docker-test
Carpetsmoker: I don't know ... this is just how I've always done things.

bhcleek: Just because a container with the expected names exists, doesn't mean the container shouldn't be built. Let docker do the work of figuring out whether the image layers are valid or not....

PR #1551, opened 2017-10-30
install vim in the Docker image

bhcleek:


Dockerfile
```diff
@@ -9,5 +9,6 @@ RUN useradd -ms /bin/bash -d /vim-go vim-go
 USER vim-go
 WORKDIR /vim-go
 COPY . /vim-go/
+RUN make install
 
 ENTRYPOINT ["make"]
```
fatih: We need to change this to `make test lint` probably as `make` also includes install

bhcleek: The default target, `all`, does include `install`, but there's no harm in that and acts a good safety valve in case you do something like `docker run --rm -v my-empty-dir:/tmp vim-go-test`.

Carpetsmoker: The `install-vim` script doesn't do anything if this version is already installed:
https://github.com/fatih/vim-go/blob/master/scripts/install-vim#L48

fatih: the default target has `install` unfortunately: https://github.com/fatih/vim-go/blob/master/Makefile#L1 Maybe we should remove it? would be better imho, anyone is free to install via `make install` later

fatih: @Carpetsmoker fair enough :)

bhcleek: I think `install` definitely belongs in `all`; I want to be able to `make` locally and have it do all the right things.

Carpetsmoker: It does seem to re-install Vim on the second run btw. I tested with both this PR and #1550 applied (probably better to do that in the same PR).

bhcleek: I'm done it several times and do not see vim getting installed when the container is run.

bhcleek:
Are we agreed that this is ready to merge?

fatih:
@bhcleek can you please hold, I've still didn't tested it. I'll post more on this today.

fatih:
Now that I've tested, I think instead of `RUN make install` we should use the following lines:

```
RUN scripts/install-vim vim-7.4
RUN scripts/install-vim vim-8.0
RUN scripts/install-vim nvim
```

I just added a simple "echo" line to `Makefile` and had to rebuilt the image again and it started again installing all Vim versions because of `RUN make install` which picked up the change to `Makefile`. The above three lines prevents it.

bhcleek:
@fatih PTAL

fatih:
lgtm

PR #1552, opened 2017-10-30
Run gometalinter --install on :GoInstallBinaries

Carpetsmoker:
This will install the linting tools from the gometalinter vendor (no
need to add all those tools to the "go get" loop).

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -153,6 +153,19 @@ function! s:GoInstallBinaries(updateBinaries, ...)
     endif
   endfor
 
+  " Install gometalinter tools.
+  let cmd = ['gometalinter', '--install']
+  if a:updateBinaries == 2
+    let cmd = add(cmd, '--update')
+  endif
+
+  echo "vim-go: Running " . join(l:cmd, " ")
+  let [out, err] = go#util#Exec(cmd)
+  if err
+    call go#util#EchoError("vim-go: gometalinter --install failed:")
+    call go#util#EchoError(out)
+  endif
+
   " restore back!
   let $PATH = old_path
   if resetshellslash
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -154,16 +154,18 @@ function! s:GoInstallBinaries(updateBinaries, ...)
   endfor
 
   " Install gometalinter tools.
-  let cmd = ['gometalinter', '--install']
-  if a:updateBinaries == 2
-    let cmd = add(cmd, '--update')
-  endif
+  if has_key(l:packages, 'gometalinter')
+    let cmd = ['gometalinter', '--install']
+    if a:updateBinaries == 2
+      let cmd = add(cmd, '--update')
+    endif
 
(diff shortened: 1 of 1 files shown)


plugin/go.vim
```diff
@@ -153,6 +153,19 @@ function! s:GoInstallBinaries(updateBinaries, ...)
...
+  if a:updateBinaries == 2
+    let cmd = add(cmd, '--update')
+  endif
+
+  echo "vim-go: Running " . join(l:cmd, " ")
+  let [out, err] = go#util#Exec(cmd)
+  if err
+    call go#util#EchoError("vim-go: gometalinter --install failed:")
+    call go#util#EchoError(out)
+  endif
```
fatih: We should make this part of the list (with arguments). Because now this will be called for cases like `:GoInstallBinaries guru`

Carpetsmoker: Oh, right :+1:

fatih:
Is there a way we can make this part of the list instead of having it a separate logic? This doesn't seems to be right somehow, feels like a hack.

Carpetsmoker:
I could, but I don't know how much sense that would make. As I understand it `gometalinter --install` will only `go install` from their vendor directory, and it won't `go get` anything. I can't really think of a scenario where you update `gometalinter` and *don't* want to run `gometalinter --install`.

bhcleek:
I agree with @fatih; this seems like a hack, because it's treating `gometalinter` specially. We could add a another value to the packages list to do follow-on tasks after the binary is installed.

Not everyone wants all the `gometalinter` tools installed; I don't use them all and I don't really want them all installed just because I chooser to use some of them.

Carpetsmoker:
Adding another value will still give you all the gometalinter tools if you type `:GoInstallBinaries` or `:GoUpdateBinaries`, so I don't really see the point to be honest?

bhcleek:
I was thinking something like:
```
\ 'gometalinter':  ['github.com/alecthomas/gometalinter', {}, ['gometalinter', '--install'],
```

So that users have a way to do some kind of post-install task. But again, I don't think this should be the default, because gometalinter doesn't need all the tools installed to function, and not everyone would want them all (I'd argue not *many* people actually want all the tools).

Carpetsmoker:
> I don't think this should be the default, because gometalinter doesn't need all the tools installed to function, and not everyone would want them all 

In that case this entire PR is rather pointless, because if you have to type  second command you might as well just type `gometalinter --install`.

bhcleek:
Fair point, but I do wonder if we couldn't expose the post-install in some way that users could leverage. A dictionary whose key is the binary name and whose value is an array that will be input to `go#util#Exec`. By default, the dictionary would be empty (i.e. `let g:go_install_binaries_after = {}`)

But could be customized by users to run commands after the binary is installed:
```
let g:go_install_binaries_after = {'gometalinter': ['gometalinter', '--install']}
```

Carpetsmoker:
I'll just close this for now; we can discus how we want to handle this in ##1555.

PR #1556, opened 2017-11-04
Remove :GoDefStack

### LINKED ISSUE #1431: Independent stacks for each vim window when using GoDef
### Behavior

When using the `go to definition` `:GoDef` feature to navigate the source code with multiple vim windows open, each window will operate on the same `:GoDefStack`. This means that if you navigated to one area of the code in one window, all the other windows opened in the same vim process will see the same `GoDef` stack and subsequently push and pop on the same stack. This becomes a bit confusing as you move back and forth between windows pushing and popping as needed.

### Steps to reproduce:

There are all sorts of combinations of this behavior that are cumbersome but here is a trivial example to reproduce it using **version 1.14**:

1. Open a Go file for editing inside of a project that you can `GoDef` in.
2. Split the window however you'd like (vertical or horizontal) so that you have 2 windows operating on the same buffer for that Go file.
3. Then go to a spot in the code where you can use `GoDef` e.g. function, type, etc. and execute `:GoDef` or your equivalent hot-key.
4. Move to the other window and run `:GoDefStack` and observe how you're using the same stack as the first window.
5. Now go ahead and pop by using `:GoDefPop` or your equivalent hot-key and observe how you'll go back to the spot that you previously ran `:GoDef` from in the first window. This is confusing because you never navigated anywhere in this window.
6. Move back to the first window and try to pop there using `:GoDefPop` or your equivalent hot-key. Observe how you don't go anywhere because you've already popped in the other window. So you're sort of stuck without a trail back home. :( Now you have to `:GoDefStack` to try and figure out where you started from to get back home in this window.

While in this trivial example it's not too big a deal, as the levels of the stack grow and you fork to new areas of the code, this begins to get really hairy. Instead, the `GoDef` feature would be greatly enhanced if we had independent stacks for each vim window.

### Configuration

* vim version: 8.0
* vim-go version: 1.14

Carpetsmoker:
Fixes #1431


bhcleek:
Since `ctrl-]` is mapped to `GoDef`, I would also like to see `CTRL-t` mapped to `CTRL-O` instead of having the custom mapping for `CTRL-t` completely removed (personally, I use `ctrl-t` a lot)...

font:
I'm not sure that mapping `ctrl-t` to `ctrl-o` is going to work like one would expect. Specifically, `ctrl-o` tracks a lot more movements in the stack than just jumps via `:GoDef`. See my [comment in #1431](https://github.com/fatih/vim-go/issues/1431#issuecomment-341938005).

fatih:
@font is right. We need to come up with a solution that only tracks file jumps, but not mark jumps. I think there was a special command where you could feed the jumps yourself.

Carpetsmoker:
Note I'm not going to work on this in the foreseeable future; I just thought it would be a quick change, but looks like a but more work is needed here. If someone wants to base their changes off this PR then they're more than welcome :-)

PR #1558, opened 2017-11-04
Add some VimScript lint tools

Carpetsmoker:
https://github.com/syngan/vim-vimlint
https://github.com/Kuniwak/vint

Also fix some errors. I disabled checks for all errors that would result in a large number of changes or non-trivial changes. We can do that later when we have better tests.

### DIFF
--- a/scripts/lint
+++ b/scripts/lint
@@ -25,9 +25,45 @@ if [ ! -f "$vimdir/bin/vim" ]; then
 	exit 1
 fi
 
+### Run vint
+############
+failed=0
+printf "Running vint ... "
+if [ -x "$(command -v vint)" ]; then
+  lint=$(vint "$vimgodir" ||:)
+  if [ -n "$lint" ]; then
+    echo "FAILED"
+    echo "$lint"
+    failed=6
+  else
+    echo "PASSED"
+  fi
+else
+  echo "SKIPPED"
+  echo "'vint' binary not found; use 'pip install vim-vint' to install it."
+fi
+
+### Run vim-vimlint
+###################
+printf "\n\nRunning vim-vimlint ... "
+lint=$(sh "$vimdir/share/vim/vimgo/pack/vim-go/start/vim-vimlint/bin/vimlint.sh" \
+  -p "$vimdir/share/vim/vimgo/pack/vim-go/start/vim-vimlparser" \
+  -l "$vimdir/share/vim/vimgo/pack/vim-go/start/vim-vimlint" \
+  -u \
+  -c func_abort=1 \
+  -e EVL110=1 -e EVL103=1 -e EVL104=1 \
+  "$vimgodir" ||:)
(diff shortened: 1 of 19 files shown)


bhcleek:
LGTM

scripts/lint
```diff
@@ -37,12 +75,13 @@ lint=$($vimdir/bin/vim -esNR \
   +"e $vimgodir/doc/vim-go.txt" \
   +'verbose VimhelpLintEcho' \
   +q \
-  2>&1)
+  2>&1 ||:)
```
bhcleek: Why is this change being made?

Carpetsmoker: To make sure the command is considered "tested" and the script doesn't exit with `set -e`.

bhcleek: I assumed that :-). But why *wouldn't* we want it to exit non-zero in that case? Appending `||:` here will make it always exit with a successful zero status code, and if for some reason there weren't any output either, the script would exit zero and there would be no indication that something was wrong.

Carpetsmoker: The issue was that the script would stop if one linter fails, rather than continuing to the next.

To do this prettier is kind of where you hit the limits of shell scripting :-/ I can fix it if you really think it's worth it, but IMHO it's not.

bhcleek: I don't understand what suddenly made `vim` exit non-zero, though; before these changes, the exit code of vim wasn't dependent on whether or not `vimhelplint` found problems.

Carpetsmoker: Maybe this specific linter never exits with `>0` on failures actually; I just copy/pasted this from the other two I added.

scripts/lint
```diff
@@ -25,6 +25,44 @@ if [ ! -f "$vimdir/bin/vim" ]; then
 	exit 1
 fi
 
+### Run vint
+############
+failed=0
+printf "Running vint ... "
+if [ -x "$(command -v vint)" ]; then
+  lint=$(vint "$vimgodir" ||:)
```
bhcleek: Does these new linters need `2>&1`?

Carpetsmoker: Won't hurt to add that

bhcleek:
LGTM

I have a weakly held opinion that we shouldn't `||:` unless it's absolutely necessary, but feel free to merge if you don't want to revert the `||:` on `vimlinthelp`.

fatih:
When I call `make docker` I see this still:

```
==> Running linting tools
Running vint ... SKIPPED
'vint' binary not found; use 'pip install vim-vint' to install it.
Running vim-vimlint ... PASSED
Running vimhelplint ... PASSED
```

Somehow this line in Dockerfile seems to be have no effect:

```
RUN pip3 install vim-vint
```

Any ideas?

Carpetsmoker:
I suspect it may be a `PATH` issue?

bhcleek:
See #1592

PR #1559, opened 2017-11-04
Use new 'goimport' tool instead of VimScript for :GoImport and :GoDrop

### LINKED ISSUE #1534: GoImport broken when adding to empty imports.
### What did you do?

``` go
package main

import ()
```

Execute: `:GoImport time`

``` go
package main

import (
	"time"
)
```

### What happened instead?

``` go
package main

import ()
	"time"
```

### Configuration

* vim version: 8.0 Included patches: 1-1207
* vim-go version: 0878992106b60210722d202164e8ba8bf710e5d2 (HEAD)
* go version (`go version`): 1.9
* operating system: Ubuntu 16.04

``` vim
set nocompatible

" Plugins

Plug 'fatih/vim-go'

call plug#end()
```

Carpetsmoker:
Instead of a large regexp-based VimScript to parse imports,
it now uses [`goimport`](https://github.com/Carpetsmoker/goimport), which parses the ast.

Fixes #1534 and deals with various other edge cases too. The behaviour is mostly the same, except in cases where it didn't make too much sense to keep the old behaviour.

For example, `:GoImport errors` will now error out if `errors` is already imported, and `:GoImport github.com/pkg/errors` will now replace any existing `errors` package instead of just adding it (this is actually the reason I started on this, because it's pretty annoying otherwise).

### DIFF
--- a/autoload/go/import.vim
+++ b/autoload/go/import.vim
@@ -1,213 +1,51 @@
-" Copyright 2011 The Go Authors. All rights reserved.
-" Use of this source code is governed by a BSD-style
-" license that can be found in the LICENSE file.
+" Change imports of the current buffer.
 "
-" Check out the docs for more information at /doc/vim-go.txt
-"
-function! go#import#SwitchImport(enabled, localname, path, bang) abort
-  let view = winsaveview()
-  let path = substitute(a:path, '^\s*\(.\{-}\)\s*$', '\1', '')
-
-  " Quotes are not necessary, so remove them if provided.
-  if path[0] == '"'
-    let path = strpart(path, 1)
-  endif
-  if path[len(path)-1] == '"'
-    let path = strpart(path, 0, len(path) - 1)
-  endif
-
-  " if given a trailing slash, eg. `github.com/user/pkg/`, remove it
-  if path[len(path)-1] == '/'
-    let path = strpart(path, 0, len(path) - 1)
-  endif
-
-  if path == ''
-    call s:Error('Import path not provided')
+" add:   add import; if 0 it will remove the import
+" alias: import alias
+" path:  import path
+" bang:  go get if packages don't exist
+function! go#import#SwitchImport(add, alias, path, bang) abort
+  let l:path = a:path
+  if l:path is# ''
+    call go#util#EchoError('import path not provided')
     return
   endif
 
-  if a:bang == "!"
-    let out = go#util#System("go get -u -v ".shellescape(path))
-    if go#util#ShellError() != 0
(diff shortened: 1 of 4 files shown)


Carpetsmoker:
I maintained the current commands for compatibility btw, but we could remove `:GoImportAs` and simply make the second argument for `:GoImport` optional.

bhcleek:
Using the AST is going to fail when the go file is not yet parseable; e.g. if someone is in the middle of editing it and they `:GoImport time`. Consider trying to use `:GoImport` when the go file is
```
package main

func main() {
   showTime()
}

func showTime() error {
    fmt.Println(time.Now())
```

Given that `GoImport` is intended to help *while* editing, I don't think `:GoImport` should require a file that is syntactically correct...

Carpetsmoker:
> Using the AST is going to fail when the go file is not yet parseable

Not exactly; we pass the [`parser.ImportsOnly`](https://github.com/Carpetsmoker/goimport/blob/master/goimport.go#L159) flag, which means it will only parse the import block. There are probably some limits to what it can and can't parse, but in general it works fairly well with invalid syntax.

fatih:
Albeit less people use this because of goimports, I still think this is valuable. I'm +1 on this 

> For example, :GoImport errors will now error out if errors is already imported,

I don't think this should  show an error, instead it should be a noop. Just like :GoFmt, :GoAddTags, etc..

bhcleek:
@Carpetsmoker testing reveals `:GoImport io` will panic when run 

```
package foo

import()

func bar() {
    println("quux")
}
```

And if it will remove the package declaration if run in
```
package foo

func bar()
    println("quux")
}
```

bhcleek:
I haven't tested it yet, but I'd expect this solution to work with files that have build tags in them....

Carpetsmoker:
I fixed the panic; I'm not 100% sure what you mean with "And if it will remove the package declaration if run in"; is that related to the panic or another issue?

It works fine with build tags, as it operates on files rather than packages. I added a test for it to make sure.

bhcleek:
It's related to another issue. When the file does not have an import statement, I was seeing the `:Goimport`  cause the package declaration to be elided.

Carpetsmoker:
Oh right; that should be fixed too now.

bhcleek:
I found another scenario where this fails. I created two new tests: refs/pull/1559/head...bhcleek:pr-1559/more-tests

The Test_import_noemptyline one fails.

fatih:
@Carpetsmoker Martin, what's the state of this? Let us know how you feel about this.

PR #1562, opened 2017-11-05
Make :GoTestCompile respect build tags

washtubs:
I have a test that uses build tags: `// +build integration`.

I found I wasn't able to build it in vim, so I added this. Seems correct to me.


bhcleek:

LGTM.

PR #1563, opened 2017-11-05
Pass --tests flag to gometalinter so that we lint test files

horgh:
Previously we'd lint test files. In the recent PR
alecthomas/gometalinter#377, gometalinter stopped asking errcheck to
lint tests. This change makes us again lint test files which I think is
sensible.

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -48,19 +48,19 @@ function! go#lint#Gometa(autosave, ...) abort
       let cmd += ["--exclude=".exclude]
     endfor
 
+    " gometalinter has a --tests flag to tell its linters whether to run
+    " against tests. While not all of its linters respect this flag, for those
+    " that do, it means if we don't pass --tests, the linter won't run against
+    " test files. One example of a linter that will not run against tests if
+    " we do not specify this flag is errcheck.
+    let cmd += ["--tests"]
+
     " path
     let cmd += [expand('%:p:h')]
   else
     " the user wants something else, let us use it.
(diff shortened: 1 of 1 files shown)


horgh:
I thought about adding a new option for this, but I have a hard time seeing someone wanting to turn it off. Also, this apparently changed recently in the PR I mentioned. I was used to having tests linted and I thought something was broken. But I'm certainly open to adding an option for this if you think that's better!

autoload/go/lint.vim
```diff
@@ -55,6 +55,13 @@ function! go#lint#Gometa(autosave, ...) abort
     let cmd += split(g:go_metalinter_command, " ")
   endif
 
+  " gometalinter has a --tests flag to tell its linters whether to run against
+  " tests. While not all of its linters respect this flag, for those that do,
+  " it means if we don't pass --tests, the linter won't run against test
+  " files. One example of a linter that will not run against tests if we do
+  " not specify this flag is errcheck.
+  let cmd += ["--tests"]
```
Carpetsmoker: I think it might be better to put this in the `if a:autosave || empty(g:go_metalinter_command)` block above.

That way people who have `g:go_metalinter_command` defined – which is documented as "for users who want to have a complete control over how gometalinter should be executed" – won't have this extra flag, which also addresses your comment and gives people a way to disable it if they really want to.

horgh:
Great point! I moved it as suggested. Thanks for looking at this.

bhcleek:
LGTM

Thank you for contributing @horgh!

PR #1577, opened 2017-11-21
update CHANGELOG.md

bhcleek:


Carpetsmoker:
As far as I'm concerned you can just commit changelog entries to master btw

fatih:
> As far as I'm concerned you can just commit changelog entries to master btw 

yeap, for small changes don't open any PR's. No need to decrease the velocity. Feel free to ping each other in case we need it @bhcleek @Carpetsmoker

PR #1578, opened 2017-11-21
prepend test name to test errors

### LINKED ISSUE #1262: Add support for testing.T.Run() metadata to ParseError()
Table-driven tests iterate over a table of inputs to permute a single test to cover multiple sub-cases.  To accomplish table-drive tests, tests use t.Run() to execute tests; the pattern is:
```golang
t.Run(subcase_name, func(t *testing.T) {
   t.Errorf("sample error")
})
```
In this pattern, testing produces output that includes `subcase_name` in the FAIL: message:
```
--- FAIL: TestEven/test_even (0.00s)
```

The output of these tests causes `ParseError()` to misbehave in interesting ways. 

### Behavior

On the command line:
```
➜  go test .
--- FAIL: TestEven (0.00s)
    --- FAIL: TestEven/sub-case_a (0.00s)
        foobar_test.go:8: this is an error
    --- FAIL: TestEven/sub-case_b (0.00s)
        foobar_test.go:8: this is an error
FAIL
FAIL    _/home/ser/test/src/github.com/foobar   0.002s
```
Note that `testing.T.Run()` is adding the test name to the FAIL message.

`ParseError()` adds the context for every error except for the first, including when there is only a single error.  It may be undesirable that the context is added in the first case, but exclusion of the test name metadata makes errors untraceable.  Copied from within nvim:
```
foobar_test.go|8| this is an error
||     --- FAIL: TestEven/sub-case_b (0.00s)
foobar_test.go|8| this is an error
```
If there are many sub-cases, it is not possible to tell which sub-case caused the first failure -- including when only a single sub-case fails.

Ideally, `ParseErrors()` would:
1. preserve the test name (e.g., "sub-case_b")
2. strip the FAIL messages (as it does with common test error messages)
3. include the name in the errors.  E.g.:
   `foobar_test.go|8|sub-case_b this is an error`

### Steps to reproduce:

1.  Copy this into a file in an empty directory:
```golang
package foobar
import "testing"
func TestEven(t *testing.T) {
   for _, v := range []string{"a", "b"} {
      t.Run("sub-case "+v, func(t *testing.T) {
         t.Errorf("this is an error")
      })
   }
}
```
2. Open this in vim
3. Run `:GoTest`

### Configuration

I can provide if necessary; however, I'm pretty sure this is because `ParseErrors()` was not written with this use case of `testing` in mind.

bhcleek:
Add an option to prepend the test name to test errors and log messages.

Fixes #1262


Carpetsmoker:
Works well in my testing

Carpetsmoker:
We could maybe consider enabling this by default by the way.

bhcleek:
I considered it, but often times the test name is long enough that it distracts from the message.

fatih:
I really liked this new addition 👍

PR #1580, opened 2017-11-23
Highlighting function in and out parameters

igor-mikushkin:
I personally like to make input parameters distinguishable from their types.
This pull request defines `goParameterName` syntax group to achieve it.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -229,14 +229,14 @@ endif
 " var, const
 if s:fold_varconst
   syn region    goVar               start='var ('   end='^\s*)$' transparent fold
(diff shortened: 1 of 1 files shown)


igor-mikushkin:
I need to add it highlights output named parameters as well.

Carpetsmoker:
Thanks!

Would you mind maybe adding a "before" and "after" screenshot? It's not entirely clear to me what a "in" and "out" parameter are, and figuring out "what is this supposed to do?" from a Vim regexp is kind of hard :-)

igor-mikushkin:
Sure.

Before:
before

After:
after

Thanks.

igor-mikushkin:
Actually it doesn't work for nested functions.
It is the case when parameter type is a function.
Let me close it and create a new one with fix later.

PR #1581, opened 2017-11-24
Fix :GoImpl completion

### LINKED ISSUE #1554: :GoImpl completion seems to break after an error
:GoImpl <Tab>
    [completes as expected]
    :GoImpl asd
    vim-go: unrecognized interface: asd
    :GoImpl <Tab>
    [nothing happens]
    :w
    :GoImpl <Tab>
    [completes as expected]

(While editing `middleware/auditMiddleware/audit_test.go`; I'll investigate later).

Carpetsmoker:
Fixes #1554

The problem is that `:GoImpl somethingInvalid` would set
`v:shell_error`, and on the next `:GoImpl <Tab>` the `go#util#env()`
call in `s:root_dirs()` would use the cache, so the `v:shell_error` is
"stuck".

Fix it by not checking this here; `go#util#env()` already shows an error
so that should be fine.

Long-term fix is to use `go#util#Exec()` everywhere which returns
`v:shell_error` instead of relying on a global value.

While I'm here also add a basic test and use `Exec()` in a few places.

### DIFF
--- a/autoload/go/impl.vim
+++ b/autoload/go/impl.vim
@@ -1,16 +1,11 @@
 function! go#impl#Impl(...) abort
-  let binpath = go#path#CheckBinPath('impl')
-  if empty(binpath)
-    return
-  endif
-
   let recv = ""
   let iface = ""
   let interactive = 0
 
   let pos = getpos('.')
 
-  if a:0 == 0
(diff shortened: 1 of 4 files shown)


bhcleek:
LGTM

PR #1583, opened 2017-11-25
Add tests for :GoTest

bhcleek:
Add tests for `:GoTest`. While adding tests, it became clear that the
parsing of `go test` errors was mostly correct, but there were a few
edge cases that made writing the tests difficult (namely race conditions
between stdout and stderr when `go test` encountered a panic) that made
the tests flaky. So in addition to adding tests, the parsing of `go
test` output was refactored to use Vim's 'errorformat' option.

Refactor test output parsing to use Vim's 'errorformat' option. Using
'errorformat' makes the parsing of messages more reliable as it keeps
the internal state of whether a multi-line match is being used.
Although the docs for 'errorformat' can be slightly obtuse, they're
actually quite complete and each entry in test.vim's 's:errorformat' is
well-commented now. Using 'errorformat' also provides consistency with
other error parsing used by vim-go (e.g. guru).

Instead of prepending the test name to the test message when
'g:go_test_prepend_name' is set, make the test name its own quickfix
entry, because there's no clear way to sensibly capture the test name
and then use it in the error message. But this is actually quite nice,
because it now includes the test status (e.g. FAIL), and we may be able
to build on this later to also include the passed tests in the error
messages when `:GoTest -v` is used. It also solves the problem of long
test names obscuring the test messages, so it may be more feasible to
set 'g:go_test_prepend_name' by default.

Stack trace produced by panics are now mostly elided in favor of a
single line whose message is the panic message and whose filename and
line number correctly identify the line that produced the panic.

* Use a Make variable, VIMS, to specify which versions of Vim should be
  installed and tested.
* Include Vim version in test report to make correlation easier.

With these two changes, it's much easier to test only a single version
of Vim by running something like

    docker run -e VIMS=vim-8.0 -v $PWD:/vim-go vim-go-test


autoload/go/test_test.vim
```diff
@@ -0,0 +1,109 @@
...
+    " are slightly out of order due to the way it interleaves stderr and
+    " stdout lines from async jobs, and it doesn't implement support for
+    " 'errorformat' correctly; multi-line errors get handled differently than
+    " they do for Vim, and the messages produce by neovim include characters
+    " that get trimmed out of Vim's messages.
+    return
+  endif
+
+  let $GOPATH = fnameescape(expand("%:p:h")) . '/test-fixtures/test'
+  silent exe 'e ' . $GOPATH . '/src/play/play_test.go'
```
Carpetsmoker: There is the `gotest#load_fixture()` helper which you should be able to use for this:

```
try
  let l:tmp = gotest#load_fixture('test/src/play/play_test.go')
finally
  call delete(l:tmp, 'rf')
endtry
```

This will copy the fixture to a tmp dir, and sets GOPATH. This means it will never write to the vim-go source directory, which is better when running tests, I think.

autoload/go/test_test.vim
```diff
@@ -0,0 +1,109 @@
...
+endfunc
+
+func! s:test(expected, verbose) abort
+  if has('nvim')
+    " don't test neovim; it seems to mostly work, but the errors it displays
+    " are slightly out of order due to the way it interleaves stderr and
+    " stdout lines from async jobs, and it doesn't implement support for
+    " 'errorformat' correctly; multi-line errors get handled differently than
+    " they do for Vim, and the messages produce by neovim include characters
+    " that get trimmed out of Vim's messages.
```
Carpetsmoker: Does this mean that `:GoTest` on NeoVim is broken, or just that it's hard to write tests for NeoVim?

autoload/go/test_test.vim
bhcleek: loadfixture only copies one file, but this test depends on more files than just the one being loaded.

autoload/go/test_test.vim
bhcleek: Neither, actually. Neovim's implementation of errorformat works, but results in slightly different results. Honestly, I'm ready to drop support for Neovim.

autoload/go/test_test.vim
Carpetsmoker: Oh, right. Should probably add support for multiple paths in that case. I'll look at that later (this is fine for now).

autoload/go/test_test.vim
Carpetsmoker: > Honestly, I'm ready to drop support for Neovim.

Aren't a lot of people using NeoVim?

bhcleek: I'm sure they are, but first-class support in vim-go for Neovim is explicitly *not* a goal for vim-go, and there are a lot of special cases in the code for it. Neovim isn't Vim, and it just causes clutter to pretend that it is. 

YMMV, and I'll continue to support it as best we can until we decide to do otherwise.

Carpetsmoker:
Seems to work well for the most part, except for:

> Stack trace produced by panics are now mostly elided in favor of a
> single line whose message is the panic message and whose filename and
> line number correctly identify the line that produced the panic.

This:

```
func p() {
    panic("asd")
}   

func TestFoo(t *testing.T) {
    p()
}
```

gives me:

    /usr/lib/go/src/testing/testing.go|711| panic: asd [recovered]

same with e.g.:

    l := []string{}
    fmt.Println(l[4])

which gives me:

    /usr/lib/go/src/testing/testing.go|711| panic: runtime error: index out of range [recovered]                 

I tried a few other things, but no matter what I do the filename is always set to `/usr/lib/go/src/testing/testing.go|711|`

bhcleek:
Good catch. I've added a panic in a top-level test in the test fixture, and will try to resolve this soon.

bhcleek:
@Carpetsmoker PTAL. I refactored to address the problems with handling panics that you discovered; there are now tests for panics that happen on the `go tests` main goroutine and panics that happen on another goroutine.

I also added tests for compiler errors.

Carpetsmoker:
This:

```
func TestFoo(t *testing.T) {
    panic("X")
}
```

gives me:

```
|| --- FAIL: TestFoo (0.00s)                                                                   
/usr/lib/go/src/testing/testing.go|711| panic: X [recovered]
/usr/lib/go/src/runtime/panic.go|491| 
a_test.go|8| 
/usr/lib/go/src/testing/testing.go|746| 
/usr/lib/go/src/testing/testing.go|789| 
```

Which is not super-great output, but acceptable. But this:

```
func TestFoo(t *testing.T) {
    t.Run("sub", func(t *testing.T) {
        panic("X")
    })  
}   
```

gives me the same, minus the test name:

```
/usr/lib/go/src/testing/testing.go|711| panic: X [recovered]                                   
/usr/lib/go/src/runtime/panic.go|491| 
a_test.go|9| 
/usr/lib/go/src/testing/testing.go|746| 
/usr/lib/go/src/testing/testing.go|789| 
```

*and* it moves my cursor to `/usr/lib/go/src/testing/testing.go:711`.

Carpetsmoker:
Actually, I have the same with regular tests too, if I disable `g:go_test_prepend_name`.

bhcleek:
Yeah, that's true. I'm not sure if there's a solution, though. Current master kind of has this same issue, but it doesn't recognize the lines as well.... I suppose I could filter out lines that are inside `GOROOT`. Does that sound like a reasonable solution to you?

Carpetsmoker:
It's not so much the extra lines that are the problem, but that the cursor jumps. I don't think that this ever happens with master like that?

Filtering GOROOT sounds okay, I think. Possible downside is that panics from within GOROOT may be obscured though. Maybe filtering just `$GOROOT/src/testing/testing.go` and `$GOROOT/src/runtime/panic.go` would be better, if that's possible?

bhcleek:
Perhaps, but I don't know that we can rely on those files across Go versions. Right now I'm planning to include the lines as messages, but without setting the file and line number matches, so they'll be available for users to read, but will be informational only.

autoload/go/test_test.vim
fatih: Neovim adoption is growing and many people using Neovim are using vim-go as well. If it works in Neovim, we shouldn't break it, especially for main commands such as `:GoTest`. I know it's hard to maintain both Vim and Neovim, part of me wants to get rid of Neovim as it's hard to keep up with two editors, but realities are different

autoload/go/test.vim
```diff
@@ -242,108 +240,141 @@ function! s:show_errors(args, exit_val, messages) abort
...
   endif
 endfunction
 
-function! s:parse_errors(lines) abort
-  let errors = []
-  let paniced = 0 " signals whether all remaining lines should be included in errors.
-  let test = ''
-
-  " NOTE(arslan): once we get JSON output everything will be easier :)
-  " https://github.com/golang/go/issues/2981
```
fatih: Can you add this back somewhere in the code. I'm still looking for the JSON support :)

fatih:
@bhcleek this looks good, was much needed. Thanks

autoload/go/test_test.vim
bhcleek: Yeah, I get it. I've tried hard to make sure this PR won't break NeoVim handling of `:GoTest` output; it significantly better, but it's a little different than Vim's.

bhcleek:
@fatih I put those comments back and added some more `TODO`s about JSON handling; I'll tackle the JSON handling as soon I have this PR wrapped up so now that it's in the Go 1.10 beta release.

bhcleek:
@Carpetsmoker thanks for the feedback. I also found a malformed pattern that, when fixed, caused some other issues. I've added a pattern to ignore stdlib lines, but this isn't quite ready yet. It's nearly done, but I need to figure out why the `%>` in the pattern that's matching the running go routine line of stack traces doesn't seem to be taking effect correctly.

I'll mention you both when this PR is ready for another review.

bhcleek:
@Carpetsmoker @fatih I think all of your concerns have been addressed, and we have much a much cleaner set of errors in the quickfix window now as well as clear descriptions of what each errorformat pattern for `:GoTest` output parsing does.

I rebased and squashed all the commits, as it was getting a bit noisy and didn't tell a great development story. Since the last review, I made sure to ignore files in GOROOT in panic stack traces, simplified the multi-line errorformat patterns (it turns out that in nearly every case I could elide the `%>` modifier), restored the `NOTE`s about JSON, and finally changed the name of the `go_test_prepend_name` option.

There's good test coverage of `autoload/go/test.vim` now, very few changes to how the test failure error messages are shown, etc.

PTAL.

bhcleek:
@fatih I added the skip for Neovim back. After correcting the test runner to run all the tests for Neovim, I saw again that the tests were failing for Neovim. (I'd originally seen them fail when running them manually, and had consequently disabled them, only to re-enable them later at your request when it seemed they passed after some refactors I'd done. At that point I had run them via the test runnner instead of manually).

But I did test look at the test output in Neovim and confirmed that the test output is consistent with current master. The only problem is that the location of a panic doesn't get shown in the quickfix window in Neovim, but since that's pre-existing, I'd like to move forward with this as is. After this is merged, I'll see if I can homogenize how Neovim and Vim deal with panics from `go test`.

bhcleek:
@Carpetsmoker I agree on the aesthetics of the test name, but it was a conscious tradeoff. I'm going to work on handling the JSON test output of `go test` in go 1.10 next, and will correct the placement of the test name as a precursor to that work; it turns out that the correction of the test name placement dovetails nicely with how we'll need to handle the JSON output.

Thanks for sticking with this; I know it was a long process.

PR #1585, opened 2017-11-26
Use :sbuffer in :GoAlternate

Carpetsmoker:
Use the `:sbuffer` command when it's already loaded. This is especially
useful if switchbuf is set to useopen or usetab.
Otherwise setting g:go_alternate_mode to `split` or `tabe` will just
keep on creating new windows or tabs.

This also fixes a bug where paths with e.g. a space didn't work
correctly.

Also add a test for go#alternate#Switch() and a few help for tests. If
`g:go_running_tests` is set, `go#util#Echo*` will record messages in
`g:go_messages` so we can easily access it from tests.

### DIFF
--- a/autoload/go/alternate.vim
+++ b/autoload/go/alternate.vim
@@ -1,32 +1,36 @@
-" By default use edit (current buffer view) to switch
-if !exists("g:go_alternate_mode")
-  let g:go_alternate_mode = "edit"
-endif
-
-" Test alternates between the implementation of code and the test code.
-function! go#alternate#Switch(bang, cmd) abort
-  let file = expand('%')
-  if empty(file)
-    call go#util#EchoError("no buffer name")
+" Alternate between the implementation of code and the test code.
+"
+" if a:bang is true then we will switch to the new file even if it doesn't
+" exist.
+" a:cmd is the edit command, such as :edit, :tabedit, :vsplit, etc.
+function! go#alternate#Switch(bang, cmd)
+  let l:file = expand('%')
+  if empty(l:file)
+    call go#util#EchoError('no buffer name')
     return
-  elseif file =~# '^\f\+_test\.go$'
-    let l:root = split(file, '_test.go$')[0]
-    let l:alt_file = l:root . ".go"
-  elseif file =~# '^\f\+\.go$'
-    let l:root = split(file, ".go$")[0]
+  elseif l:file[-8:] is# '_test.go'
+    let l:root = split(l:file, '_test.go$')[0]
+    let l:alt_file = l:root . '.go'
+  elseif l:file[-3:] is# '.go'
+    let l:root = split(l:file, '.go$')[0]
     let l:alt_file = l:root . '_test.go'
   else
-    call go#util#EchoError("not a go file")
+    call go#util#EchoError(printf('%s is not a go file', l:file))
     return
   endif
-  if !filereadable(alt_file) && !bufexists(alt_file) && !a:bang
-    call go#util#EchoError("couldn't find ".alt_file)
+
+  if !filereadable(l:alt_file) && !bufexists(l:alt_file) && !a:bang
+    call go#util#EchoError("couldn't find " . l:alt_file)
     return
-  elseif empty(a:cmd)
-    execute ":" . g:go_alternate_mode . " " . alt_file
+  endif
+
+  if bufloaded(l:alt_file)
+    let l:cmd = 'sbuffer'
   else
-    execute ":" . a:cmd . " " . alt_file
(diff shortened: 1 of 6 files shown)


bhcleek:
LGTM

fatih:
Somehow this broke my alternate mappings:

```
  autocmd Filetype go command! -bang A call go#alternate#Switch(<bang>0, 'edit')
  autocmd Filetype go command! -bang AV call go#alternate#Switch(<bang>0, 'vsplit')
  autocmd Filetype go command! -bang AS call go#alternate#Switch(<bang>0, 'split')
  autocmd Filetype go command! -bang AT call go#alternate#Switch(<bang>0, 'tabe')
```

I was using `:A` to switch. Now instead of replacing the buffer it opens another one in split mode. This shouldn't happen imho. You can reproduce this by calling it multiple times.

bhcleek:
@fatih what is your `'switchbuf'` (e.g. `set switchbuf?`) set to?

fatih:
It's empty. So this breaks now when people don't set it by default?

bhcleek:
No, it shouldn't.

Carpetsmoker:
I don't think it's related to the `switchbuf` setting, as it works for me as intended. `:A` behaves as if I had typed `:edit foo_test.go` for me with both my regular vimrc, my vimrc with `switchbuf` commented out, and with a minimal vimrc. So I think there is another setting or something?

Carpetsmoker:
But I can reproduce with your vimrc here: https://raw.githubusercontent.com/fatih/dotfiles/master/vimrc

I'll get it fixed.

fatih:
Yeah, I'm also not sure what breaks it. The vimrc above is %90 right as I don't change the base settings very often (I only add new plugins or settings sometimes). I'll also try to check it out, post more if I can find something.

Carpetsmoker:
It's `set hidden`; need to use `:buffer` in that case, and not `:sbuffer`.

Here's a quick patch. I'll make a PR out of it with a test when it's not 3 am :-)

```
diff --git i/autoload/go/alternate.vim w/autoload/go/alternate.vim
index 71e0da9..3e4d098 100644
--- i/autoload/go/alternate.vim
+++ w/autoload/go/alternate.vim
@@ -24,10 +24,13 @@ function! go#alternate#Switch(bang, cmd)
     return
   endif
 
+  let l:cmd = empty(a:cmd) ? get(g:, 'go_alternate_mode', 'edit') : a:cmd
   if bufloaded(l:alt_file)
-    let l:cmd = 'sbuffer'
-  else
-    let l:cmd = empty(a:cmd) ? get(g:, 'go_alternate_mode', 'edit') : a:cmd
+    if l:cmd is# 'edit'
+      let l:cmd = 'buffer'
+    else
+      let l:cmd = 'sbuffer'
+    endif
   endif
 
   exe printf(':%s %s', l:cmd, fnameescape(l:alt_file))
```

bhcleek:
@Carpetsmoker I was able to duplicate this, too, but it's not clear to my why `set hidden` is a problem here.

I think we should revert this PR and split out the things it addressed into separate PRs; some of what it introduced (e.g. fixing bugs) is useful, but whether or not vim-go should use `sbuffer` and how it should do so is questionable in light of `set hidden`.

bhcleek:
@Carpetsmoker I was able to duplicate this, too.

I don't think the patch you suggested is going to be sufficient, though. Taking @fatih's example: assuming a file named `foo.go` is loaded in the current window, `go#alternate#Switch(<bang>0, 'vsplit')` should execute `:vsplit foo_test.go`, but would instead execute `:buffer foo_test.go`.

I think we should revert this PR and split out the things it addressed into separate PRs; some of what it introduced (e.g. fixing bugs) is useful, but whether or not vim-go should use `sbuffer` and how it should do so is questionable in light of `set hidden`.

Carpetsmoker:
> I don't think the patch you suggested is going to be sufficient, though. Taking @fatih's example: assuming a file named foo.go is loaded in the current window, go#alternate#Switch(<bang>0, 'vsplit') should execute :vsplit foo_test.go, but would instead execute :buffer foo_test.go.

Running `:buffer` or `:sbuffer` is the point, since repeatedly running `:vsplit` would keep on splitting to a new window, which seems fairly useless to me. With `:(s)buffer` it will switch to it, which seems much better UX to me.

> I think we should revert this PR and split out the things it addressed into separate PRs; some of what it introduced (e.g. fixing bugs) is useful, but whether or not vim-go should use sbuffer and how it should do so is questionable in light of set hidden.

I was away for the weekend, but yes, you could have reverted it. Don't be afraid to hurt my feelings :-)
Either way, I'll look this evening after work and either fix it or revert it.

PR #1589, opened 2017-11-28
Support minisnip

Carpetsmoker:
Needs https://github.com/joereynolds/vim-minisnip/pull/5

I've also changed the default for `g:go_snippet_engine` to `automatic`,
which is a new value that will use the first snippet engine that's
loaded. I think that's more user-friendly than expecting the user to set
it.

I also changed the detection method to use the variables, instead of
scanning `runtimepath`. This is more efficient.

Need to actually add the snippets for minisnip though.


fatih:
What makes minisnip superior to ultisnips or neosnippets? Back then I've added them to add support, but I'm not sure we should support other plugins, especially new plugins those with a low userbase. I'm not against it, just trying to understand the value it'll bring.

Carpetsmoker:
> What makes minisnip superior to ultisnips or neosnippets

It's a very simple plugin which Just Works™ without too much Magic. I just personally prefer it over a large plugin written in Python etc.

> I'm not sure we should support other plugins, especially new plugins those with a low userbase

That okay; I don't really care. I can reduce the PR to just some of the small improvements I made.

It just seemed to me that adding this doesn't really affect much of anything. If you don't use it then you won't know it's there, and it adds pretty much zero maintenance overhead.

fatih:
> It just seemed to me that adding this doesn't really affect much of anything. If you don't use it then you won't know it's there, and it adds pretty much zero maintenance overhead.

True that, I was just worried that this can easily can spread to other plugins. I'm curious if we can somehow do the detection manual. I'm ok merging this btw, I'll chime more if I have something to show.

PR #1592, opened 2017-11-30
fix vim-vint installation

bhcleek:
Move RUN instruction to install vim-vint so that it will be in $PATH.
By moving the instruction to before the USER instruction, vim-vint will
be installed to /usr/local/bin instead of /vim-go/.local/bin;
$HOME/.local/bin is not in $PATH...

Also upgrade to Go 1.9.2.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,11 @@
-FROM golang:1.9.1
+FROM golang:1.9.2
 
 RUN apt-get update -y && \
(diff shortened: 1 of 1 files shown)


fatih:
lgtm

> $HOME/.local/bin is not in $PATH...

maybe we should fix it by adding it to PATH? I'm merging as it works now.

PR #1595, opened 2017-12-04
Revert "Use :sbuffer in :GoAlternate"

Carpetsmoker:
Reverts fatih/vim-go#1585

See discussion – don't really have time to fix/test *right now* (which is basically code for "I don't feel like it"), so revert and fix when I do.


fatih:
Thanks @Carpetsmoker, hope to see why it fails.

PR #1598, opened 2017-12-05
do not re-install vim on every vim-go change

bhcleek:


fatih:
This failed for me with the error:
```
...
vim-go: could not find 'fillstruct'. Run :GoInstallBinaries to fix it
FAIL fillstruct_test.vim        0.052731s / 1 tests

Some tests FAILED
Makefile:10: recipe for target 'test' failed
```

bhcleek:
Ah, of course. Thanks @fatih.

I moved the installation of all the go tools out of `scripts/install-vim` and into `scripts/test`. This has the disadvantage that they'll be installed on every `make docker`, but it's still much faster than it was when all the Vims were being installed on each `make docker`.

PTAL

fatih:
> This has the disadvantage that they'll be installed on every make docker

Can't we make it that it doesn't install it everytime? it just should continue using the tool. I think we should abandon building from the localhost and use a public vim-go-test image where it downloads if it's not available and uses it. Later whenever we want something new in it, we built and publish it. This way I'll only pull the image. I'll try your changes and report back.

scripts/test
```diff
@@ -51,6 +51,20 @@ if [ ! -f "$vimdir/bin/vim" ]; then
   exit 1
 fi
 
+
+# Make sure all Go tools and other dependencies are installed.
+echo "Installing Go binaries"
+"$vimgodir/scripts/run-vim" $vim +':silent :GoUpdateBinaries' +':qa'
```
Carpetsmoker: This should be protected by an `if` so it doesn't get run on every test invocation.

bhcleek: What kind of check do you propose?

bhcleek:
@fatih The disadvantage of such a thing, though, is that you don't ever really know if the latest vim-go is going to work with the latest version of some tool; this has the advantage that when tooling changes in a way that breaks vim-go, we'll know if we use `make docker`.

You're free to not use `make docker`. I frequently use `docker run --rm -i -t -v $PWD:/vim-go vim-go-test` and `docker run --rm -i -t -v $PWD:/vim-go -e VIMS=vim-8.0 vim-go-test` to avoid rebuilding the image. I've always thought it was wrong for `make docker` to build the image and run the tests, but it's easy enough to work around....

Perhaps just using the `docker run` command to test in a container instead of `make docker` would resolve the problem you were seeing to begin with?

fatih:
I would expect a docker build gives me the same cached result every time I do. If it doesn't it means it's not created in a way that allows us to reproduce the issue. It means after couple of months, I can't just git checkout it any more, because it will not build the image I would expect. This is definitely the wrong way we tackle this. At a given git SHA, a docker build always should give us the correct image and always the same one.

bhcleek:
It's the same image; but the tests run against the tools tools that are current as of the time the tests run. That seems correct to me. What you're describing isn't really possible, because a `git checkout bc12345 && docker build -t vim-go-test . && docker run --rm -i -t vim-go-test` today will *never* be guaranteed to do the same today as it does tomorrow, because `:GoUpdateBinaries` will give different results through time.

scripts/test
Carpetsmoker: Maybe check the contents of GOPATH or something?

Right now running tests is very fast, which is a nice feature to have IMHO. This would make it significantly slower.

(I also think that install-vim is a better place btw, but practicality beats
purity, so if this makes solving the Docker problem a lot easier then this is
okay).

bhcleek:
I can install the binaries efficiently during the docker build, too, but I won't be able to get to it for a few days.

bhcleek:
PTAL.

I think this state is a reasonable tradeoff. The installation of vims is cached in the image and will only be invalidated if `scripts` is change, installation of go tools happens in the image instead of during running of tests. These changes allow `make docker` to run in about a 1 minute on my system when the vim-go sources are changed but the installation of the vims has been cached. For iterating on changes more quickly, the resulting image can be used by running `docker run --rm -v $PWD:/vim-go vim-go-test` (about 30 seconds on my system). When combined with the changes in https://github.com/fatih/vim-go/pull/1583/files, that 30 seconds can be reduced even further by iterating on only a single version of vim...

bhcleek:
The Travis build is failing, but I'm not sure why. @Carpetsmoker any ideas?

Carpetsmoker:
The problem is that `install-go-tools` is never called if you use `./install-vim vim-8.0 && ./test vim-8.0`.

bhcleek:
Good call @Carpetsmoker ! I've rebased and fixed the Travis CI config file.

.travis.yml
```diff
@@ -9,6 +9,8 @@ matrix:
     - env: SCRIPT=lint VIM_VERSION=vim-8.0
 install:
   - ./scripts/install-vim $VIM_VERSION
+  - ./scripts/install-go-tools vim-8.0
```
Carpetsmoker: This needs to be `$VIM_VERSION`, otherwise it tries to use a non-existing Vim version.

.travis.yml
```diff
@@ -9,6 +9,8 @@ matrix:
...
   - pip install --user vim-vint covimerage codecov
+  - pip install --user vim-vint
```
Carpetsmoker: I think removing `covimerage codecov` is accidental?

Makefile
```diff
@@ -7,6 +7,7 @@ install:
 	@for vim in $(VIMS); do \
 		./scripts/install-vim $$vim; \
 	done
+	@./scripts/install-go-tools vim-8.0
```
Carpetsmoker: Same as the .travis.yml: if you try to install Vim 7.4 and you don't have vim 8.0, it will try to use a `vim` binary which doesn't exist.

Carpetsmoker:
If I run `./install-vim vim-8.0` then I would expect it to install Vim + set up all dependencies. That's what it currently does, anyway.

With this PR this is changed, since I now need to call *two* scripts. can't say I'm hugely in favour of that to be honest; it was jut one script originally, then it was two, and now three :-/

What problem does this solve? That `:GoInstallBinaries` installs from `master` and changes the files? I think we should fix that from just the Dockerfile, for example by not building a Docker image locally for everyone but pulling a image from Dockerhub or something.

bhcleek:
By not building locally and pulling an image instead? That image still has to be built some place, then you're back to @fatih's original concern. This was an attempt to address @fatih's concern about running `make docker` repeatedly and it installing Vim everytime just because something in `autoload/go` changed, for instance. Personally, I still think this is a reasonable compromise, but 🤷‍♂️ .

Just consider this workflow:
edit
`make docker`
edit
`make docker`

of course, one can:
edit
`make docker`
edit
`docker run --rm -v $PWD:/vim-go-test vim-go-test`

Which is what I usually do, is a bit more of a burden on contributors. But again: 🤷‍♂️ . All I really want is a way to run the tests in a docker environment so test workflow doesn't pollute my local environment and there's no concern about it accidentally overwriting my local vim configuration...

Makefile
```diff
@@ -8,6 +8,9 @@ install:
 		./scripts/install-vim $$vim; \
 	done
 
+	# only need to install the go tools once...
```
Carpetsmoker: Actually this isn't true, as it installs the Go tools to a temporary GOPATH in the Vim's install directory, so it needs to be installed for every Vim. This is why it was in the `install-vim` script.

scripts/install-go-tools
```diff
@@ -0,0 +1,13 @@
...
+#
+
+set -euC
+vimgodir=$(cd -P "$(dirname "$0")/.." > /dev/null && pwd)
+cd "$vimgodir"
+
+vim=$1
+# Make sure all Go tools and other dependencies are installed.
+echo "Installing Go binaries"
+"$vimgodir/scripts/run-vim" $vim +':silent :GoUpdateBinaries' +':qa'
```
Carpetsmoker: Oh ... you removed the `export GOPATH=$installdir` :-/

Now it will overwrite people's GOPATHs.

Carpetsmoker:
I think we should make the stuff in `./scripts` user-friendly and easy to use. If folks want to run that in Docker then fine, but if they don't then that should work just as well.

IMHO relying wrappers in the form of `make` or `docker` is rarely a good thing.

I'm not sure what we're trying to fix here, either. It seems to me that the problem is that `./scripts/docker-test` runs `docker build` on every invocation. If it only gets built when there is no image yet, then it's also fixed, no?

bhcleek:
As a developer, I'd expect `make docker` to rebuild the image if it needs to be rebuilt, not just build it if it doesn't exist. i.e. If any vim-go files are different from what's in the image, then it needs to be rebuilt, otherwise `docker run --rm vim-go-test` only tests what was last built into the image. Consider:

```
...edit
make docker
...edit
make docker
```

In that scenario, the second `make docker` would not run the tests against the vim-go changes that were introduced in the second edit.

Carpetsmoker:
Wouldn't adding a mount make more sense for that?

bhcleek:
That's basically what I suggested in https://github.com/fatih/vim-go/pull/1598#issuecomment-349460466, and is frequently how *I* work, but it's not sufficient if `plugin/go.vim` changes; in that case, the go tools could be out of date. So now we get back to the tradeoffs that this PR makes: https://github.com/fatih/vim-go/pull/1598#issuecomment-349789418

What we're trying to do here is to make sure that `make docker` is as fast as possible while also caching the vim installations so they aren't done on every `docker build` if the scripts that install vim haven't changed, while also building the latest `vim-go` sources into the image. Of course, we could take a middle ground (and maybe this is what you're suggesting?) by not building the vim-go sources (other than `scripts`) into the docker image, and instead modify `scripts/docker-test` to run the image with a mounted volume, and then also install the Go tools and run the tests, but that doesn't seem to quite be what you want either, because the installation of the Go tools is still separated. And yet it's not actually possible to reliably install the go tooling as part of installing Vim in the image *and* make keep the caching of the vim installation and the installing of the Go tools at separate layers.

PR #1601, opened 2017-12-06
Append Go file type when creating temporary file

### LINKED ISSUE #208: goreturns configuration
I am struggling setting up vim-go with goreturns [https://github.com/sqs/goreturns](https://github.com/sqs/goreturns). What I tried is this:

```
    let g:go_fmt_command = 'goreturns'
    let g:go_fmt_options = '-w'
```

but it seems it always ignores the `-w` flag (and just prints the output to stdout). Any tips on that?

Thanx!

mocheryl:
A lot of scripts filter files based on their name to make sure they
operate only on the appropriate content. By appending a Go file type to
a temporary file when running a formatter, we add support for a broader
range of Go formating tools.

Fixes #208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,9 @@ IMPROVEMENTS:
 * `g:go_snippet_engine` now defaults to `automatic` to use the first installed
   snippet engine it can find.
   [[GH-1589]](https://github.com/fatih/vim-go/pull/1589).
-
+* Make sure temporary files created for `:GoFmt` end with `.go` suffix as this
+  is required by some Go formatting tools
+  [[GH-1601]](https://github.com/fatih/vim-go/pull/1601).
 
 BUG FIXES:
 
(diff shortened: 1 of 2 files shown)


Carpetsmoker:
I think this should be fine, but would be good to test on Windows, too. Just to be sure.

I don't think that any of the regular Go maintainers use Windows – I don't have a Windows machine anyway – but I'll look at running the tests on AppVeyor too, so we can test it in the CI.

mocheryl:
> I don't have a Windows machine anyway

Neither do I (tried on Linux using neovim), but if there's no other way, I'll see if I can procure one from somewhere.

If it doesn't work, then another way would be to pipe the buffer content to standard input of the formatter, but that would require be a bit more code change.

Carpetsmoker:
Well, I acquired a Windows VM from modern.ie. Setting up AppVeyor is going to be a lot more work than I thought.

I'll test tomorrow. It was mentioned in another PR a while ago that `tempname()` behaves quite different on Windows than it does on Linux/macOS, which is why it would be good to test this PR on Windows.

Carpetsmoker:
On Windows it returns a path like `C:\Users\IEUser\AppData\Local\Temp\VIC99BD.tmp`; this patch seems to work well. Thanks! :+1:

PR #1604, opened 2017-12-07
Add support for denite.nvim about :GoDecls / :GoDeclsDir

delphinus:
[denite.nvim][] is the successor of [unite.vim][].  This PR enables `:GoDecls` & `:GoDeclsDir` features by denite as same as unite in #1391.

[denite.nvim]: https://github.com/Shougo/denite.nvim
[unite.vim]: https://github.com/Shougo/unite.vim

```vim
" available below as same as :Unite decls
:Denite decls
:Denite decls:foo.go
:Denite decls:foo/bar
```

denite is written for NeoVim / Vim 8 and it uses more progressive features than unite.  The expansions of denite must be written in Python3.

| `:Denite decls` |
|---|
|  |

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -149,7 +149,8 @@ The following plugins are supported for use with vim-go:
 * Interactive |:GoDecls| and |:GoDeclsDir|:
   https://github.com/ctrlpvim/ctrlp.vim or
   https://github.com/junegunn/fzf.vim or
-  https://github.com/Shougo/unite.vim
+  https://github.com/Shougo/unite.vim or
+  https://github.com/Shougo/denite.nvim
 
 ==============================================================================
 COMMANDS                                                         *go-commands*
@@ -687,7 +688,7 @@ CTRL-t
     Requires `ctrlp.vim` or `fzf`; it will autodetect the plugin if installed,
     but you can use |'g:go_decls_mode'| to force using one or the other.
     By default `type` and `func` declarations are shown. This can be changed
-    via |'g:go_decls_includes'|. Also see |unite-decls|.
+    via |'g:go_decls_includes'|. Also see |unite-decls|, |denite-decls|.
 
                                                                  *:GoDeclsDir*
 :GoDeclsDir [dir]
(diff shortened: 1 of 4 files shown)


fatih:
Thanks @delphinus,

I'm not sure this is something I would like to merge. Who is going to maintain the python script afterwards? We already have fzf, ctrlp and unite.

Carpetsmoker:
My 2 cents is that this adding this won't affect any existing code or features. If you don't use denite.vim then you won't even know vim-go has a plugin thingy for it. If it ever breaks in the future we can just remove it if no one steps up to fix it.

fatih:
@Carpetsmoker what do you plan if a bug report comes for denite?

Carpetsmoker:
Tell the submitter to fix it or learn to live with the bug :-) If it's a critical bug we can simply remove the denite.vim plugin if no one steps up.

fatih:
Sounds fair :)

doc/vim-go.txt
```diff
@@ -693,20 +694,25 @@ CTRL-t
     [dir] is given it parses the given directory.
 
                                                                  *unite-decls*
+                                                                *denite-decls*
 :Unite decls[:file or dir]
+:Denite decls[:file or dir]
```
Carpetsmoker: `:Denite decls dir` doesn't seem to work? I get `[denite] NameError: Source "dir" is not found.`.

doc/vim-go.txt
```diff
@@ -1564,10 +1570,10 @@ By default the template file specified by |'g:go_template_file'| is used.
...
                                                        *'g:go_decls_includes'*
 
-Only useful if `ctrlp.vim`, `unite.vim` or `fzf` are installed. This sets
-which declarations to show for |:GoDecls| (`ctrp.vim`) and |unite-decls|
-(`unite.vim`).  It is a Comma delimited list Possible options are:
-{func,type}.  The default is: >
+Only useful if `ctrlp.vim`, `unite.vim`, `denite.nvim` or `fzf` are installed.
+This sets which declarations to show for |:GoDecls| (`ctrp.vim`),
+|unite-decls| (`unite.vim`) and |denite-decls| (`denite.nvim`).  It is a Comma
+delimited list Possible options are: {func,type}.  The default is: >
```
Carpetsmoker: Very minor, but missing full stop before "Possible" here.

Carpetsmoker:
After testing I have `rplugin/python3/denite/source/__pycache__/decls.cpython-36.pyc` as an untracked file in git. Would be a good idea to add that to gitignore to prevent it from being accidentally checked in.

Can just add `*.pyc`; that will cover this and all future cases.

doc/vim-go.txt
delphinus: denite needs options `:` separated.  And the `file or dir` is some illustration, so users should supply the real path.

```vim
:Denite decls:~/.go/src/github.com/golang/appengine
" relative path also available. same meaning if you are in ~/.go/src/github.com
:Denite decls:golang/appengine
```

The description is a bit obscure maybe, but I followed descriptions already written about unite.

doc/vim-go.txt
delphinus: 👍

delphinus:
Added to `.gitignore`. check it. 👍

fatih:
> The expansions of denite must be written in Python3.

Can we include this information in the doc? I assume Vim also needs to be compiled with +python right?

delphinus:
Added requirements for denite. 👍

doc/vim-go.txt
Carpetsmoker: Oh, right. I normally don't use this plugin so excuse my ignorance. I've tweaked the documentation a bit, which makes it a bit clearer I think.

Carpetsmoker:
Thanks! :+1:

PR #1607, opened 2017-12-10
Implement -line for fillstruct

Carpetsmoker:
This way the cursor doesn't need to be exactly on the struct.

If multiple structs are found it will still prefer the one under the
cursor. Otherwise it will expand both.

Needs https://github.com/davidrjenni/reftools/pull/8 (tests will fail before that's merged)


bhcleek:
LGTM

PR #1608, opened 2017-12-10
 Use :sbuffer in :GoAlternate

Carpetsmoker:
Second attempt at #1585; I'll add some more tests later. Just opening this so I don't forget.


fatih:
@Carpetsmoker Martin, what's the state of this? Would be great if we can finish and ship it, if not we can close if you don't have the time for this.

PR #1613, opened 2017-12-14
Make sure all tests run in Neovim

bhcleek:
Neovim closes stdin, so make sure to give it a stdin that is not
inherited from the parent process, because the parent process is reading
from its stdin...

Add .local/ to the list of ignored patterns for Docker so that Neovim
does not encounter errors trying to read the ShaDa file when trying to
build the Docker image after having run the Docker container with the
cwd mounted into the container at /vim-go.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
+.local/
 .git/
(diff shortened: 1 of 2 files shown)


Carpetsmoker:
Ah, nice find :+1:

PR #1616, opened 2017-12-16
[snippets] make "error multiple return" distinct

explodes:
errn was defined twice. Once for 
return nil
and another for 
return ${1:nil}, err

I propose naming the second snippet errnn, as it would commonly be used to return nil, err.

### DIFF
--- a/gosnippets/UltiSnips/go.snippets
+++ b/gosnippets/UltiSnips/go.snippets
@@ -167,7 +167,7 @@ ${0}
 endsnippet
 
 # error multiple return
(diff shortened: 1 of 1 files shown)


explodes:
Its probably worth mentioning that "errn" expansion defaults to the first definition, "return nil", so this change will be backwards compatible.

Carpetsmoker:
I could be wrong and I didn't test it, but it looks like one snippet is `errn`, while the other is `errn,` with a comma?

explodes:
Neat! I was mistaken.

PR #1618, opened 2017-12-17
Fix go#template#create() for directories that don't exist yet

Carpetsmoker:
Previously `vim ~/go/src/newdir/main.go` would result in:

	"~/go/src/newdir/main.go" [New DIRECTORY]
	Error detected while processing function <SNR>56_template_autocreate[3]..go#template#create:
	line    6:
	E344: Can't find directory "/home/martin/go/src/newdir" in cdpath
	E472: Command failed

With this change this gets treated like a directory which isn't a Go
package yet.


bhcleek:
I found another scenario where this fails. I created two new tests: https://github.com/fatih/vim-go/compare/refs/pull/1559/head...bhcleek:pr-1559/more-tests

The `Test_import_noemptyline` one fails.

Carpetsmoker:
Wrong PR I think @bhcleek? :-)

PR #1621, opened 2017-12-19
Active buffer relative to workdir for gometalint

tombee:
When `go#lint#Gometa` is triggered by autosave, we attempt to only show
lint warnings and errors that are applicable for the active buffer. This
is done by passing the path to the file represented by the active buffer
to gometalinter's `--include` option.

The existing behaviour expands the absolute path, whereas gometalinter
outputs paths that are relative to the current working directory.

This change modifies the variable used for matching so that we also use
the current working directory for matching against lint errors/warnings.

Signed-off-by: Tom Barlow <tomwbarlow@gmail.com>

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -88,7 +88,7 @@ function! go#lint#Gometa(autosave, ...) abort
 
   if a:autosave
     " include only messages for the active buffer
-    let cmd += ["--include='^" . expand('%:p') . ".*$'"]
+    let cmd += ["--include='" . @% . ".*$'"]
   endif
 
 
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -62,6 +62,11 @@ function! go#lint#Gometa(autosave, ...) abort
     let cmd += split(g:go_metalinter_command, " ")
   endif
 
+  " Include only messages for the active buffer for autosave.
+  if a:autosave
+    let cmd += [printf('--include=%s.*$', @%)]
+  endif
+
   " gometalinter has a default deadline of 5 seconds.
   "
   " For async mode (s:lint_job), we want to override the default deadline only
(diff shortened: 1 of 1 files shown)


tombee:
I'd been fighting with this for a couple of days, wondering why `let g:go_metalinter_autosave = 1` wasn't working for me. After debugging (and learning to debug vimscript), I discovered the command that was being executed was returning no results for me on the CLI.

This looks to be caused by the modification in `gometalinter` to make using the relative path the default: https://github.com/alecthomas/gometalinter/pull/394.

I considered modifying the `--format` parameter to use `{{.Path.Abs}}` instead to maintain backward compatibility. I felt this might be a little too opinionated, and cause breakages if users wish to override that option.

I've decided to drop the `^` for matching the beginning of the path, hopefully that means that we maintain backward compatibility with older versions of `gometalinter` and continue to work with newer versions that use relative paths by default.

I'd appreciate your thoughts, I'm a total newbie at vimscript and not at all familiar with the codebase.

Carpetsmoker:
Thanks!

The code path you modified won't get run for must users; a little bit above it is:

```
" Call gometalinter asynchronously.
if go#util#has_job() && has('lambda')
    " [..]
    call s:lint_job({'cmd': cmd})
    return
endif
```

It's only a fallback for people with older versions of Vim.

I pushed an update which should make it work for both code paths, which seems to work well in my test. I'll see if I can add some automated tests, too.

tombee:
@Carpetsmoker Oh I see, thanks for that! 

I'm running neovim (NVIM v0.2.2), and this is the codepath I was hitting when debugging it. I guess I might be missing something in my environment to get the async runner working.

bhcleek:
Built #1640 on top of this. Closing in favor of #1640; note that it contains all the work done here.

Thanks @tombee !

tombee:
Thanks @bhcleek 🎉

PR #1626, opened 2017-12-28
Add note for deoplete on Vim 8

delphinus:
Now [deoplete.nvim][] supports Vim 8 in addition to NeoVim.

[deoplete.nvim]: https://github.com/Shougo/deoplete.nvim


fatih:
Thanks @delphinus

PR #1629, opened 2017-12-28
add additional godoc url configuration

stamblerre:
The user's configured "go_doc_url" variable should still work correctly in the case when the "gogetdoc" command isn't installed.

### DIFF
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -29,18 +29,7 @@ function! go#doc#OpenBrowser(...) abort
     let name = out["name"]
     let decl = out["decl"]
 
-    let godoc_url = get(g:, 'go_doc_url', 'https://godoc.org')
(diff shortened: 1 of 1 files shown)


fatih:
Hi @stamblerre 

Seems like this is still missing things? I dont' see the actual fix you describe in the commit message. I assume you want to move out the `gogetdoc` check from the `go#doc#OpenBrowser()` function however this PR doesn't include it. Let me know if I miss something.

stamblerre:
Sorry, I probably wasn't very clear in the message. I just wanted to change the place on line 64 where "https://godoc.org/" was still hardcoded.

fatih:
This looks good. The only small thing would is the function signature. In Vim script we use `snake_case`, but I'll do that after merging. Thanks @stamblerre

PR #1632, opened 2017-12-30
Set filetype on Bufread

happyyi008:
by setting the filetype on buf read we can do stuff like
autocmd FileType go ...

I use this for setting comment leaders in my vimrc

### DIFF
--- a/ftdetect/gofiletype.vim
+++ b/ftdetect/gofiletype.vim
@@ -20,7 +20,7 @@ endfunction
 
 augroup vim-go-filetype
   autocmd!
-  au BufNewFile *.go setfiletype go | setlocal fileencoding=utf-8 fileformat=unix
+  au BufNewFile,BufRead *.go setfiletype go | setlocal fileencoding=utf-8 fileformat=unix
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
Thanks!

This is done already from:

    au BufRead *.go call s:gofiletype_pre("go")

And that function ends with:

    let &l:filetype = a:type

I can confirm it works. I've been using `autocmd FileType go` in [my vimrc](https://github.com/Carpetsmoker/dotfiles/blob/master/vim/p-go.vim#L23) for quite some time.

Does this not work for you? Or do you have a problem? I think that may be caused by some other setting or issue?

happyyi008:
hmm... I just checked it again and it seems to be working now without the change.
I'm pretty new to vimscript so I may have been doing something wrong.

https://github.com/happyyi008/dotfiles/blob/master/.vimrc#L55
This is where I'm using it. And when I tried to use to macro `,cc` it would say b:comment_leader not defined.

Carpetsmoker:
That should work fine; I don't know why it didn't for you.

But if it works now then great! :-) Feel free to open an issue if you have problems.

PR #1633, opened 2017-12-31
fix quickfix list when test duration exceeds timeout

bhcleek:
Do not treat a panic caused by a test timeout as an error that occurred
within a test's callstack. When go test panics due to a test timeout,
the running goroutine's stacktrace only contains entries for the
standard library, and there is one goroutine whose stacktrace contains
entries for the standard library and the test binary. None of the other
goroutines should be considered to contain an error, either, because the
panic was not due to any code executing within those goroutines' stacks.
Therefore, treat panics caused by a test timeout as an error without a
file location.

### DIFF
--- a/autoload/go/test_test.vim
+++ b/autoload/go/test_test.vim
@@ -47,6 +47,16 @@ func! Test_GoTestCompilerError() abort
   call s:test('compilerror/compilerror_test.go', expected)
 endfunc
 
+func! Test_GoTestTimeout() abort
+  let expected = [
+        \ {'lnum': 0, 'bufnr': 0, 'col': 0, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': '', 'pattern': '', 'text': 'panic: test timed out after 500ms'}
+      \ ]
+
+  let g:go_test_timeout="500ms"
+  call s:test('timeout/timeout_test.go', expected)
+  unlet g:go_test_timeout
+endfunc
+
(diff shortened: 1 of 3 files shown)


autoload/go/test_test.vim
```diff
@@ -47,6 +47,16 @@ func! Test_GoTestCompilerError() abort
   call s:test('compilerror/compilerror_test.go', expected)
 endfunc
 
+func! Test_GoTestTimeout() abort
+  let expected = [
+        \ {'lnum': 0, 'bufnr': 0, 'col': 0, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': '', 'pattern': '', 'text': 'panic: test timed out after 2s'}
+      \ ]
+
+  let g:go_test_timeout="2s"
```
Carpetsmoker: Is it possible to use a shorter timeout here, like `100ms`? Or at the very least `1s`? 2 seconds isn't *that* long, but if you're modifying something then it's just too long for a "smooth" modify/test workflow, and these small timeouts tend to add up once the testing coverage increases.

If it's not easily possible then don't bother. It's a very small thing. Looks good otherwise :+1:.

bhcleek: It's possible. I just chose two seconds because it's reasonably short, and I was pretty sure everything that needs to start would actually be started in that time. I'll try dialing it back and will merge with the reduced timeout or this if it's not possible to reduce it reliably...

PR #1634, opened 2018-01-02
support syntax highlighting for bracketed expression in fmt string

e-zhang:
Per the go docs for the fmt package, there is a bracket expression that is supported in format strings. 

For example,
`
fmt.Sprintf("%[2]d %[1]d\n", 11, 22)
fmt.Sprintf("%[3]*.[2]*[1]f", 12.0, 2, 6)
fmt.Sprintf("%d %d %#[1]x %#x", 16, 17)
`

Should all be valid syntax highlighted format strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ BUG FIXES:
 * The user's configured `g:go_doc_url` variable wasn't working correctly in the
   case when the "gogetdoc" command isn't installed.
   [[GH-1629]](https://github.com/fatih/vim-go/pull/1629)
+* Highlight format specifiers with an index (e.g. `%[2]d`).
+  [[GH-1634]](https://github.com/fatih/vim-go/pull/1634)
 
 ## 1.16 - (December 29, 2017)
 
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -202,7 +202,7 @@ else
 endif
 
 if g:go_highlight_format_strings != 0
-  syn match       goFormatSpecifier   /\([^%]\(%%\)*\)\@<=%[-#0 +]*\%(\%(\%(\[\d\]\)\=\*\)\|\d\+\)\=\%(\.\%(\%(\%(\[\d\]\)\=\*\)\|\d\+\)\)\=\%(\[\d\]\)\=[vTtbcdoqxXUeEfgGsp]/ contained containedin=goString,goRawString
+  syn match       goFormatSpecifier   /\([^%]\(%%\)*\)\@<=%[-#0 +]*\%(\%(\%(\[\d\]\)\=\*\)\|\d\+\)\=\%(\.\%(\%(\%(\[\d\]\)\=\*\)\|\d\+\)\)\=\%(\[\d\]\)\=[vTtbcdoqxXUeEfFgGsp]/ contained containedin=goString,goRawString
   hi def link     goFormatSpecifier   goSpecialString
 endif
 
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -202,7 +202,7 @@ else
 endif
 
 if g:go_highlight_format_strings != 0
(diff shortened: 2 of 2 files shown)


Carpetsmoker:
Thanks! Seems to work well judging from looking at `fmt/fmt_test.go` in Vim. I added a commit to support `%F` too, as that didn't work either.

The only pattern that don't work yet are:

    {"%.f", 0i, "(0+0i)"},
    {"% .f", 0i, "( 0+0i)"},
    {"%+.f", 0i, "(+0+0i)"},
    {"% +.f", 0i, "(+0+0i)"},

I tried to fix it, but [leaning toothpick syndrome](https://en.wikipedia.org/wiki/Leaning_toothpick_syndrome) hurts by brain at this hour. I think we can make this more readable by some strategic insertions of `\v` and such.

I also wonder if there's a good way to split this across multiple lines, preferably with comments like in Python or Ruby:

```
 pattern = """
    ^                   # beginning of string
    M{0,4}              # thousands - 0 to 4 M's
    (CM|CD|D?C{0,3})    # hundreds - 900 (CM), 400 (CD), 0-300 (0 to 3 C's),
                        #            or 500-800 (D, followed by 0 to 3 C's)
    (XC|XL|L?X{0,3})    # tens - 90 (XC), 40 (XL), 0-30 (0 to 3 X's),
                        #        or 50-80 (L, followed by 0 to 3 X's)
    (IX|IV|V?I{0,3})    # ones - 9 (IX), 4 (IV), 0-3 (0 to 3 I's),
                        #        or 5-8 (V, followed by 0 to 3 I's)
    $                   # end of string
    """
```

That would make it much easier to maintain!

e-zhang:
I will take a look at those and learn to run the units as well. Thanks for the feedback, I'll try to update this PR with those changes as soon as a I can.

Carpetsmoker:
Here's the best I could come up with:

```
diff --git i/syntax/go.vim w/syntax/go.vim
index 89b0e9a..a7bc2e1 100644
--- i/syntax/go.vim
+++ w/syntax/go.vim
@@ -202,7 +202,19 @@ else
 endif
 
 if g:go_highlight_format_strings != 0
-  syn match       goFormatSpecifier   /\([^%]\(%%\)*\)\@<=%[-#0 +]*\%(\%(\%(\[\d\]\)\=\*\)\|\d\+\)\=\%(\.\%(\%(\%(\[\d\]\)\=\*\)\|\d\+\)\)\=\%(\[\d\]\)\=[vTtbcdoqxXUeEfFgGsp]/ contained containedin=goString,goRawString
+  " 1. Match a literal % not preceded by a %.
+  " 2. Match -, #, 0, space, or +
+  " 3. ??
+  " 4. ??
+  " 5. ??
+  " 6. Match formatting verb.
+  syn match       goFormatSpecifier   /\
+                  \([^%]\(%%\)*\)\
+                  \@<=%[-#0 +]*\
+                  \%(\%(\%(\[\d\]\)\=\*\)\|\d\+\)\=\
+                  \%(\.\%(\%(\%(\[\d\]\)\=\*\)\|\d\+\)\)\=\
+                  \%(\[\d\]\)\
+                  \=[vTtbcdoqxXUeEfFgGsp]/ contained containedin=goString,goRawString
   hi def link     goFormatSpecifier   goSpecialString
 endif
 ```

Vim is *very* finicky about where you can place the `\` though, and it doesn't always work :-(

Maybe a better solution is to use string concatenation (untested):

```
" Match a literal % not preceded by a %.
let s:p = '\([^%]\(%%\)*\)\'

" Match -, #, 0, space, or +
let s:p .= '\@<=%[-#0 +]*\'

[...]

exe 'syn match goFormatSpecifier /' . s:p . '/ contained containedin=goString,goRawString'
```

e-zhang:
Took your suggestion for doing multi-line continuation and added comments. I *think* everything in `fmt/fmt_test.go` is highlighted correctly now. 

I'm not familiar enough with the \v option in vim, so I punted on doing that for now.

Carpetsmoker:
Thanks! Works well, and much more readable/maintainable like this, even though it's still a long/ugly regex :-)

fatih:
Nice addition 👍

PR #1635, opened 2018-01-03
Remove screenshot

mgumz:
This commit reduces the file size of the assets/screenshot.png file
from 742kb to 80kb by reducing the amount of used colors to 32
(`pngquant 32 assets/screenshot.png`).


fatih:
Hi @mgumz  We don't use it anymore. I think it's safe to delete it. You can change the PR to delete the screenshot if you wish.

Carpetsmoker:
Thanks! Gotta save those bytes :-)

mgumz:
@Carpetsmoker hey, when i checked all the "bundles.zip" my little vopher (https://github.com/mgumz/vopher) fetched ... vim-go was the biggest by a wide margin. you have to fix what you have to fix :)

PR #1637, opened 2018-01-04
handle testify output

### LINKED ISSUE #1636: `let g:go_jump_to_error = 0` does not prevent new buffer from opening when using 3rd party testing packages
Related to #367.

### What did you do?
I ran `:GoTest` from a file containing a test that intentionally fails.

I expected the QuickFix buffer to open, showing me information about the error(s).

### What happened instead?
The QuickFix buffer opened, along with a new file `mypackage/assertions.go`, which contained only the line `package main`. My cursor jumped into this new file.

The screenshot below shows my directory structure, a file containing some code, the file testing that code, and the `mypackage/assertions.go` buffer that opened. Note: the auto-opening buffer replaced my test file buffer, but for the screenshot I reopened the test file next to it.

image

### Configuration:

VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Dec 18 2017 10:09:08)
macOS version
Included patches: 1-1400
Compiled by Homebrew

* Go version (`go version`):
go version go1.9.2 darwin/amd64

* Go environment (`go env`):
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/sasha/go"
GORACE=""
GOROOT="/usr/local/Cellar/go/1.9.2/libexec"
GOTOOLDIR="/usr/local/Cellar/go/1.9.2/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/kb/t4v3k6fx1hq6xlsd9mnyt2rc0000gn/T/go-build898171635=/tmp/go-build -gno-record-gcc-switches -fno-common"
CXX="clang++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"

* vim-go version:
v1.16

If you'd like my full .vimrc, please let me know. Otherwise, these are the vim-go related options I have set.
```vimrc

Plug 'fatih/vim-go'

call plug#end()

" Vim-Go
let g:go_fmt_command = "goimports"
let g:go_jump_to_error = 0
```

Thanks in advance for your advice!

bhcleek:
I'm not sure if we want to do this or not, but the fix was relatively simple... testify is its own beast, though, and I'm reluctant to add official support for it.

Fixes #1636

### DIFF
--- a/autoload/go/test-fixtures/test/src/testify/testify_test.go
+++ b/autoload/go/test-fixtures/test/src/testify/testify_test.go
@@ -1,0 +1,27 @@
+package testify
+
+import (
+	"testing"
+)
+
+func Nop() {}
+
+func TestTestify(t *testing.T) {
+	// write consistent with github.com/stretchr/testify output that indicates
+	// the file location is line 7, the line that has the Nop function.
+	t.Errorf("%s", "\r             \r\tError Trace:\ttestify_test.go:7\n\r\tError:      \ttestify is not go test")
+}
+
+/*
+func TestMain(t *testing.T) {
+	t.Run("trying to hide", func(t *testing.T) {
+		t.Errorf("%s", "\rwhere is the indentation and filename?")
+	})
+	t.Run("fail", func(t *testing.T) {
+		t.Error("fail")
+	})
+	t.Run("is false", func(t *testing.T) {
+		assert.False(t, true)
+	})
+}
+*/
--- a/autoload/go/test.vim
+++ b/autoload/go/test.vim
@@ -295,6 +295,13 @@ function! s:errorformat() abort
     let format .= ",%-G" . indent . "--- FAIL: %.%#"
   endif
 
+  " Match github.com/stretchr/testify output. We won't make a great effort
+  " (e.g. we won't try to match messages that contain embedded newlines), but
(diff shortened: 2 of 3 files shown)


fatih:
This came up before and I've said 👎  Reason is that maintaining third party compatibility is not worth it. We should aim to support for `Go` only. This should go into a plugin called `vim-go-testify` or something alike.  We use third party tools, but those provide a feature that we control, here someone decides to use on their own and we're trying to fix it.  There was also another PR that would add support for a third party templating library, but I've also rejected that as well due the same reason. 

I know you want to be helpful, but this is not what vim-go should do going forward.

ernesto-jimenez:
I believe this should address the issue: https://github.com/stretchr/testify/pull/554

PR #1641, opened 2018-01-09
respect a changed g:go_test_show_name

bhcleek:
Respect the current value of g:go_test_show_name on :GoTest runs after
the first one by invalidating the cached errorformat value when the
g:go_test_show_name value is not the same as it was when the errorformat
value was last cached.

Only show the test name for failed tests instead of including the
'--- FAIL: ' prefix and the duration suffix when g:go_test_show_name=1.

Resolve a TODO item in the test errorformat so that the regular
expression used for test errorformat values is more precise.

Remove a bit of testing code that is no longer needed; it was probably
rendered obsolete with some recent changes to how the errorformat value
is created.

### DIFF
--- a/autoload/go/test-fixtures/test/src/showname/showname_test.go
+++ b/autoload/go/test-fixtures/test/src/showname/showname_test.go
@@ -1,0 +1,11 @@
+package main
+
+import "testing"
+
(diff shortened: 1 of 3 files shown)


fatih:
lgtm

PR #1643, opened 2018-01-10
FIX #1642: use Execute for Gorename

### LINKED ISSUE #1642: Renaming a variable in a file that is in a subdirectory won't rename the whole repo

GoRename a variable in a file that is in a subdirectory of a repository directory.

Should rename all the typenames that are used in other subdirectories in the same repo.

### What happened instead?

Only files in package/subpackageA are renamed. 

`Vim 8.0`

* Go version (`go version`):

`go version go1.8.3 darwin/amd64`

* Go environment (`go env`):

I don't think this is related to env.

* vim-go version:

master 3ceecc2bfa682a5f1f1db226b16bcd6edda8027b

c9s:
#1642

### DIFF
--- a/autoload/go/rename.vim
+++ b/autoload/go/rename.vim
@@ -65,7 +65,7 @@ function! go#rename#Rename(bang, ...) abort
   endif
 
   let command = join(cmd, " ")
-  let out = go#tool#ExecuteInDir(command)
+  let out = go#tool#Execute(command)
 
   let splitted = split(out, '\n')
   call s:parse_errors(go#util#ShellError(), a:bang, splitted)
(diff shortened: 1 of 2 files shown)


bhcleek:
Thanks you for contributing, @c9s. I'm not clear what this is fixing, though; under what conditions is this necessary?

Carpetsmoker:
Thanks for your PR! This change seems okay in a quick test as such, but I'm not entirely sure what it's fixing? `:GoRename` on a new (unsaved) file still doesn't work, as it operates on the file path and not the buffer.

I'll close this for now, but please feel free to add additional details on what this fixes and I'll be happy to reopen it.

bhcleek:
Just in case this gets re-opened, one of the things that needs to be tested is that calling `:GoRename` works correctly when the buffer is not in $PWD.

PR #1645, opened 2018-01-11
Fix ftdetect

ichizok:
When sourcing `ftdetect/*.vim` (in `$VIMRUNTIME/filetype.vim`), `augroup filetypedetect` is already set.

Thus using `augroup` in ftdetect script, autocmds in gofiletype.vim and in other ftdetect scripts sourced after gofiletype.vim do not belong to `augroup filetypedetect`.
This makes some behaviors about filetype (e.g. `:filetype off`) broken.

### DIFF
--- a/ftdetect/gofiletype.vim
+++ b/ftdetect/gofiletype.vim
@@ -1,3 +1,5 @@
+" vint: -ProhibitAutocmdWithNoGroup
+
 " We take care to preserve the user's fileencodings and fileformats,
 " because those settings are global (not buffer local), yet we want
 " to override them for loading Go files, which are defined to be UTF-8.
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @ichizok 

This is from the help doc `:help ftdetect`:

```
Note that there is no "augroup" command, this has already been done
when sourcing your file.  You could also use the pattern "*" and then
check the contents of the file to recognize it.
```

I've looked at the Git history and the change was introduced by @Carpetsmoker. I assume you must be right (according to the doc) but let me ping him as well.

Carpetsmoker:
Yeah, this change seems okay :+1: I didn't know that ftdetect already wrapped stuff in an augroup.

PR #1648, opened 2018-01-16
Fix handling and change name of g:go_metalinter_excludes

bhcleek:
Use gometalinter's `--disable` option instead of its `--exclude` to
disable linters specified in `g:go_metalinter_excludes`; `--exclude` is
to exclude files, `--disable` is for disabling linters.

Rename `g:go_metalinter_excludes` to `g:go_metalinter_disabled` to more
accurately reflect its purpose in its name and so that it's symmetric
with `g:go_metalinter_enabled`.

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -10,8 +10,8 @@ if !exists("g:go_metalinter_enabled")
   let g:go_metalinter_enabled = ['vet', 'golint', 'errcheck']
 endif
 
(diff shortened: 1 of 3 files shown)


fatih:
lgtm

bhcleek:
TravisCI is currently having problems. Once they're back on their feet and this PR builds successfully, I'll merge it.

fatih:
restarted the build

PR #1652, opened 2018-01-19
separate gometalinter error list use cases

### LINKED ISSUE #1644: I want to use quickfix when execute GoMetalinter but use locationlist when autosave

Use GoMetalinter and `g:go_metalinter_autosave`

We can separate list depending on autosave or not

### What happened instead?

We cannot it now

NVIM v0.2.2

* Go version (`go version`):

1.9.2

* Go environment (`go env`):

* vim-go version:

`3ceecc2bfa682a5f1f1db226b16bcd6edda8027b` (latest of master branch)

---

I use quickfix when I explicitly run, and want to use locationlist in lint which runs automatically.

How about to allow following?
```
let g:go_list_type_commands = {"GoMetaLinterAutoSave": "locationlist"}
```

bhcleek:
gometalinter is used in two distinct cases. When run using
:GoMetaLinter, all of its output is used. When run automatically on save
(e.g. g:go_metalinter_autosave=1), only its output that applies to the
current buffer is used. In that case, it makes sense to show its
messages in the location list without forcing the user to also use the
location list when the user executes :GoMetaLinter.

Therefore, add a new key to g:go_list_type_commands,
GoMetaLinterAutoSave, with a default value of locationlist and modify
go#lint#Gometa to distinguish between the two cases.

Fixes #1644

### DIFF
--- a/autoload/go/lint_test.vim
+++ b/autoload/go/lint_test.vim
@@ -74,8 +74,10 @@ func! Test_GometaAutoSave() abort
         \ {'lnum': 5, 'bufnr': 2, 'col': 1, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'w', 'pattern': '', 'text': 'exported function MissingDoc should have comment or be unexported (golint)'}
       \ ]
 
-  " clear the quickfix lists
-  call setqflist([], 'r')
+  let winnr = winnr()
+
+  " clear the location lists
+  call setloclist(l:winnr, [], 'r')
 
   " call go#lint#ToggleMetaLinterAutoSave from lint.vim so that the file will
   " be autoloaded and the default for g:go_metalinter_autosave_enabled will be
@@ -89,11 +91,11 @@ func! Test_GometaAutoSave() abort
 
   call go#lint#Gometa(1)
 
-  let actual = getqflist()
+  let actual = getloclist(l:winnr)
   let start = reltime()
   while len(actual) == 0 && reltimefloat(reltime(start)) < 10
     sleep 100m
-    let actual = getqflist()
+    let actual = getloclist(l:winnr)
   endwhile
 
   call gotest#assert_quickfix(actual, expected)
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -797,6 +797,9 @@ CTRL-t
 
     Toggles |'g:go_metalinter_autosave'|.
 
+    By default, `gometalinter` messages will be shown in the |location-list|
+    window. The list to use can be set using |'g:go_list_type_commands'|.
+
                                                  *:GoTemplateAutoCreateToggle*
 :GoTemplateAutoCreateToggle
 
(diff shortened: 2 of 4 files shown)


doc/vim-go.txt
```diff
@@ -1437,9 +1437,9 @@ Specifies the type of list to use for command outputs (such as errors from
 builds, results from static analysis commands, etc...). When an expected key
 is not present in the dictionary, |'g:go_list_type'| will be used instead.
 Supported keys are "GoBuild", "GoErrCheck", "GoFmt", "GoInstall", "GoLint",
-"GoMetaLinter", "GoModifyTags" (used for both :GoAddTags and :GoRemoveTags),
-"GoRename", "GoRun", and "GoTest".  Supported values for each command are
-"quickfix" and "locationlist".
+"GoMetaLinter", "GoMetaLinterAutoSave", "GoModifyTags" (used for both
+:GoAddTags and :GoRemoveTags), "GoRename", "GoRun", and "GoTest".  Supported
+values for each command are "quickfix" and "locationlist".
```
fatih: Can you also update the doc for `g:go_metalinter_autosave` setting and `:GoMetaLinterAutoSaveToggle` command to add a notice that `locationlist` is used instead of `quickfix`. These two:

```
                                                  *'g:go_metalinter_autosave'*

Use this option to auto |:GoMetaLinter| on save. Only linter messages for
the active buffer will be shown. By default it's disabled >

  let g:go_metalinter_autosave = 0
<
```

and

```
                                                 *:GoMetaLinterAutoSaveToggle*
:GoMetaLinterAutoSaveToggle

    Toggles |'g:go_metalinter_autosave'|.
```

It's worth adding this because for example I use `quickfix` for everything and have custom mappings to select next/previous items from the list. So people should be aware of the side effects.

autoload/go/lint_test.vim
```diff
@@ -89,11 +91,11 @@ func! Test_GometaAutoSave() abort
 
   call go#lint#Gometa(1)
 
-  let actual = getqflist()
+  let actual = getloclist(l:winnr)
   let start = reltime()
   while len(actual) == 0 && reltimefloat(reltime(start)) < 10
     sleep 100m
-    let actual = getqflist()
+    let actual = getloclist(l:winnr)
```
fatih: These commands should be not removed, instead it should be based on `go#list#Type("GoMetaLinterAutoSave")` and act accordingly, whether it's  a location list or quickfix list.

autoload/go/lint_test.vim
```diff
@@ -74,8 +74,10 @@ func! Test_GometaAutoSave() abort
         \ {'lnum': 5, 'bufnr': 2, 'col': 1, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'w', 'pattern': '', 'text': 'exported function MissingDoc should have comment or be unexported (golint)'}
       \ ]
 
-  " clear the quickfix lists
-  call setqflist([], 'r')
+  let winnr = winnr()
+
+  " clear the location lists
+  call setloclist(l:winnr, [], 'r')
```
fatih: Same here. We should have a switch and call `setqflist()` or `setloclist()` based `go#list#Type("GoMetaLinterAutoSave")`

bhcleek: This test is only about the autosave case, and it only tests with the default list type `GoMetaLinterAutoSave`. Why would it check the list type before clearing the list?

autoload/go/lint_test.vim
bhcleek: Same comment as above.

doc/vim-go.txt
bhcleek: done

PR #1653, opened 2018-01-20
handle autowriteall the same as autowrite

bhcleek:
Vim docs say that autowriteall implies autowrite, so make sure that
go#cmd#autowrite() has the same behavior when only one of either
autowrite or autowriteall is set.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -1,5 +1,5 @@
 function! go#cmd#autowrite() abort
-  if &autowrite == 1
+  if &autowrite == 1 || &autowriteall == 1
     silent! wall
   endif
 endfunction


fatih:
@bhcleek I'm not sure about this change. Maybe I don't want to enable `autowriteall` because I don't want it to save on `:exit` ? (this is what I do btw). So in that regard, I think `go#cmd#autowrite()` is correct that it only checks`autowrite`. Is there something I'm missing here?

bhcleek:
@fatih Thanks for taking the time to review this tiny PR.

Like you, my vim is configured with `autowrite` and `noautowriteall`, but after reading the docs for `autowrite`, and verifying some of vim's behaviors with it's configured with `noautowrite` and `autowriteall`, I realized that `go#cmd#autowrite` was incomplete.

* `go#cmd#autowrite` is called to make sure any unsaved changes in a buffer are saved before running a command like `:GoVet` if the user has set `autowrite`.
*  When a user sets only `autowriteall` and not `autowrite` (e.g. `set autowriteall`, `set noautowrite`), vim will behave as if `autowrite` is set. From `:help autowriteall`:
> Setting this option also implies that Vim behaves like 'autowrite' has been set.
* You can see for yourself that when `autowriteall` is set and `autowrite` is not, that moving to another buffer saves the changes before showing the next buffer.

All this PR does is make sure that vim-go saves the buffer using the same rules with respect to `autowrite` and `autowriteall` that vim itself applies.

PR #1656, opened 2018-01-22
Fix async race

### LINKED ISSUE #1440: Check for race conditions similar to the one fixed in #1439
In #1439 it was mentioned that "the other async routines likely need to be fixed in the same manner, but I didn't look into it."

bhcleek:
Eliminate race conditions that were possible in vim8 job handling.

Most jobs were only handling the `exit_cb` handler for job options. To make sure that all messages are handled, `close_cb` has to be hooked up and coordinated with `exit_cb`, because there is no guarantee about which order they will be called in.

Refactor `autoload/go/job.vim` to call the caller-provided callbacks only after the job has exited _and_ channel has been closed so that the actual exit status and all messages can be provided. I also took this opportunity to clean up and simplify `go#job#Spawn` so that there's only one callback called when the job completes instead of two and so that the return value only exposes the callbacks that are needed by the caller.

Refactor `autoload/go/guru.vim` to account for the fact that it's possible that `exit_cb` may be called _after_ `close_cb`.

Refactor all other async jobs to coordinate calling functions that should be called only after the job's exit status is known and all the channel is closed.

Because this touched so much code, I tried to make each of the commits focused on a single area of improvement. You may find it much easier to review each commit rather than reviewing them all at once.

Closes #1440

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -269,7 +269,7 @@ function s:cmd_job(args) abort
   " autowrite is not enabled for jobs
   call go#cmd#autowrite()
 
-  function! s:error_info_cb(job, exit_status, data) closure abort
+  function! s:complete(job, exit_status, data) closure abort
     let status = {
           \ 'desc': 'last status',
           \ 'type': a:args.cmd[1],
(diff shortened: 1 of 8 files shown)


bhcleek:
I've marked this as WIP, because I've discovered a problem with guru. I'll remove the label once it's fixed.

bhcleek:
The problem I was encountering wasn't related to these changes.

PR #1669, opened 2018-02-05
[WIP] def: implement lazy evaluation of definition call

fatih:
This speeds up guru/godef jumping immensely by computing the destination
before the user decided to jump. This makes it feel like any jump is
instantaneous as the evaluating is done in background.

Because we don't want to block the UI during the computing it's only
supported for async mode.

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -1,7 +1,10 @@
 let s:go_stack = []
 let s:go_stack_level = 0
 
-function! go#def#Jump(mode) abort
+let s:last_res = {}
+let s:last_job = test_null_job()
+
+function! go#def#Compute(mode, jump) abort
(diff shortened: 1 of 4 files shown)


fatih:
Tests are passing for Vim 8.0 branch, but not Vim 7.4 and NeoVim (as expected).

fatih:
After using this for following days this quote came to my mind:

```
There are only two hard things in Computer Science: cache invalidation and naming things.

-- Phil Karlton
```

Indeed the Cache invalidation stuff is hard to get right. There are many edge cases, everytime I fix one, something else appear. I'm closing this issue. I think there are better solutions (like server based guru that maintains the cache itself), instead of this. Let's keep it simple. Thanks for reviewing nevertheless @bhcleek

PR #1673, opened 2018-02-08
Ensure that goPackageComment region ends after comment.

thsnr:
If a package comment contained a line starting with "package", then it terminated the package comment. Ensure that "package" is preceded with a comment block terminator to fix the highlighting.

Before and after:
before after


thsnr:
The build failure seems to be caused by a network connection dropping and unrelated to the pull request.

Carpetsmoker:
Thanks, seems to work well in my testing. Can't find any cases where this causes problems.

PR #1293, opened 2017-05-18
do not jump to the quickfix window when metalint

jinleileiking:
#1290 

copy codes from `ale`

Issue #1073, opened 2016-10-04
[vim-8.0] quickfix or locationlist steal focus after save


achilleas-k:
### Actual behavior

With the current vim-8.0 branch, after saving a file, linting begins in the background (async). If there are errors, the quickfix or location-list buffer appears and the cursor moves into the new split. This is especially bad with the new async feature, since the user is free to continue typing or move around the file while linting is running in the background (which can take a few seconds) and the focus stealing interrups flow.
### Expected behavior

Quickfix or location list should appear and not steal focus (or it should be optional).
### Steps to reproduce:

Reproducible with minimal vimrc file. Only vim-go installed. Just save a go file with an error that would be caught by an active linter.
### Configuration

vim version: 8.0
vim-go version: vim-8.0 head (`29196e5cb81793983c6838fbf15871c9298c35ff`)

Configuration:

``` vim
let g:go_metalinter_enabled = ['go', 'vet', 'golint', 'errcheck']
let g:go_metalinter_autosave = 1
let g:go_metalinter_autosave_enabled = ['go', 'vet', 'golint', 'errcheck']
```

fatih:
Hi @achilleas-k 

I agree, but I couldn't find a solution for that. Even though I say don't jump, Vim still jumps :) If you find something let me know

achilleas-k:
Not at all familiar with vimscript but I'll have a poke.

achilleas-k:
Well, I managed to hack up a workaround, at least for the quickfix window (not sure about location list).
From [this SO solution](http://superuser.com/a/775751/286961), the following does prevent the qf window from stealing focus:
```vim
augroup quickfix
  autocmd!
  autocmd Syntax qf wincmd p
augroup END
```
The problem then becomes that the cursor jumps to the beginning of the file when qf opens.

So after that, I wrapped `copen` in a winview save and restore. The entire callback in the `lint_job` function now looks like this:
```vim
  function! s:callback(chan, msg) closure
    let old_errorformat = &errorformat
    let &errorformat = l:errformat
    caddexpr a:msg
    let &errorformat = old_errorformat

    " TODO(arslan): cursor still jumps to first error even If I don't want
    " it. Seems like there is a regression somewhere, but not sure where.
    augroup quickfix
      autocmd!
      autocmd Syntax qf wincmd p
    augroup END
    let winview=winsaveview()
    copen
    call winrestview(winview)
  endfunction
```
and it works as expected.

[...]

PR #1679, opened 2018-02-21
Fix for locationlist or quickfix list stealing focus on save.

mikijov:
Related to issue #1073. The issue must have regressed since the fix in #1293. The PR is simply re-adding original fix, albeit to different location since the codebase has changed.

I have fixed the tests as well. However, I am not very familiar with how vim testing works, so I would appreciate someone verifying it.

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -122,6 +122,8 @@ function! go#lint#Gometa(autosave, ...) abort
 
     if !a:autosave
       call go#list#JumpToFirst(l:listtype)
+    else
+      wincmd p " the cursor jumped to qf: go back to the active window
     endif
   endif
 endfunction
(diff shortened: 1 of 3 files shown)


bhcleek:

This will affect all codepaths that populate and show the error lists, not just on save. Can you target the fix to apply only to linting that happens when saving the file?

mikijov:
Well, here you go. Thanks for pushing me to learn a bit of vimscript. Now I feel I accomplished something (as compared to 8 hours of meetings). :-) I hope it's what you wanted.

bhcleek:
Thanks @mikijov. On first glance, this looks right.

bhcleek:
Thank you @mikijov . With some slight modifications, your changes have been incorporated into master as of https://github.com/fatih/vim-go/commit/eaef6f79323a3df110928020dfc877d26332c80a

PR #1684, opened 2018-02-26
Flags to disabled gocode omnifunc.

gianarb:
There are now other autocomplete available for go. I am currently using
YCM and I would like to disable gocode.

That's why I added a parameter: `g:go_omnifunc_enabled`.

Signed-off-by: Gianluca Arbezzano <gianarb92@gmail.com>

### DIFF
--- a/ftplugin/go.vim
+++ b/ftplugin/go.vim
@@ -21,7 +21,9 @@ setlocal noexpandtab
 compiler go
 
 " Set gocode completion
-setlocal omnifunc=go#complete#Complete
+if get(g:, "go_omnifunc_enabled", 1)
+  setlocal omnifunc=go#complete#Complete
+endif
 
 if get(g:, "go_doc_keywordprg_enabled", 1)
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
You can reset it with `autocmd FileType go setlocal omnifunc=`. I'm not sure if we need/want another option for this, as the help page is already quite long. Also not sure why you need to disable it, is it somehow "in the way"?

gianarb:
Hello @Carpetsmoker , I didn't know that you are not accepting new options anymore.

```
> Also not sure why you need to disable it
```
I used YCM for other languages and it looks a bit faster compared with `gocode` so, I am happy to disable it from my setup.

Carpetsmoker:
> I used YCM for other languages and it looks a bit faster compared with gocode so, I am happy to disable it from my setup.

Okay, but why does it matter if the `omnifunc` setting is set or not?

gianarb:
If it's not set YCM will take action.

PR #1686, opened 2018-02-27
Make quickfix window always occupy the full width

cakturk:
When I run a vim-go command which causes the quickfix window to open,
that window is placed at the bottom of the rightmost column of windows
if there are vertical splits, which, I think is annoying.

To make it occupy the full width, we can use the `:botright` command
modifier for `:copen`.

### DIFF
--- a/autoload/go/list.vim
+++ b/autoload/go/list.vim
@@ -42,7 +42,7 @@ function! go#list#Window(listtype, ...) abort
   if a:listtype == "locationlist"
     exe 'lopen ' . height
   else
-    exe 'copen ' . height
+    exe 'botright copen ' . height
   endif
 endfunction
 


fatih:
Hi @cakturk 

You can put the quickfix window to the bottom by changing it yourself. Put the following to your vimrc:

```
" put quickfix window always to the bottom
augroup quickfix
    autocmd!
    autocmd FileType qf wincmd J
    autocmd FileType qf setlocal wrap
augroup END
```

Thanks for the fix though.

cakturk:
Thanks for your suggestion. I initially thought that this would be something everyone could enjoy.

Rican7:
> You can put the quickfix window to the bottom by changing it yourself. Put the following to your vimrc:
> 
> ```
> " put quickfix window always to the bottom
> augroup quickfix
>     autocmd!
>     autocmd FileType qf wincmd J
>     autocmd FileType qf setlocal wrap
> augroup END
> ```

Commenting here for future reference, as this came up in search.

While adding `autocmd FileType qf wincmd J` to your vimrc will work, its not ideal, as "location lists" _also_ share the `qf` FileType.

If you'd like to set all quickfix windows to open full-width at the bottom, without effecting window-local location-lists, add this to your vimrc instead:

```vim
" Position the (global) quickfix window at the very bottom of the window
" (useful for making sure that it appears underneath splits)
"
" NOTE: Using a check here to make sure that window-specific location-lists
" aren't effected, as they use the same `FileType` as quickfix-lists.
autocmd FileType qf if (getwininfo(win_getid())[0].loclist != 1) | wincmd J | endif
```

(See https://github.com/Rican7/dotfiles/commit/5e00fa97267dcdef5b76f5289aefea14edcc2fc5 for example)

PR #1696, opened 2018-02-28
fix missing <buffer> for debug mappings

Jagua:


Carpetsmoker:
Thanks for your PR?

Could you explain what issue this fixes? Right now I can do:

```
vim a.go
:GoDebugStart
<F9>

:e b.go
<F9>
<F5>
```

To place a breakpoint in both `a.go` and `b.go`, which I think is useful. But with this PR I won't be able to do that any more, since the mappings are just for the `a.go` buffer.

Does this fix a specific error or issue? If so, then we should probably fix that, instead of making the mappings buffer-local.

Jagua:
I have fixed that issue since my first PR was incomplete.

I think that a plugin should not override GLOBAL mappings.
That is, once executing `:GoDebugStart`, valuable keymaps such as `<F6>` will remain broken when editing for other languages.

This PR fixes this issue.

Carpetsmoker:
Thanks. The issue with this is that the mappings are, well, only available in the buffer. For example with `:tabedit foo.go` they won't be available. Personally I think it's useful to have that work. Same with some other cases.

I think a better fix is to not map stuff if it's already mapped:

```
if !hasmapto('<Plug>(go-debug-continue)')
  nmap <F5>   <Plug>(go-debug-continue)
endif
if !hasmapto('<Plug>('(go-debug-print)')
  nmap <F6>   <Plug>(go-debug-print)
endif

" ..etc..
```

Or only map the F-keys if it's not mapped to something else already:

```
if mapcheck('<F5>') is ''
  nmap <F5>   <Plug>(go-debug-continue)
endif
```

We could perhaps also add an option:

```
if get(g:, 'go_debug_map', 1) is 0
  nmap <F5>   <Plug>(go-debug-continue)
endif
```

You can then add `let g:go_debug_map` to disable the default mappings.

I don't mind making a patch, but can't really decide which option is best here.

fatih:
> I don't mind making a patch, but can't really decide which option is best here.

Making them optional with `<Plug>` is the correct way. That's the way to remap mappings and we already have a dedicated section of mappings and started using them since the project has started. But I think we should hold it up to the next release. It's not urgent imho.

Jagua:
> For example with `:tabedit foo.go` they won't be available.

Hmm, my patch (second commit) works fine in my environment.
(Because, mappings are defined by `autocmd` in each buffer which filetype is "go".)

However, if it does not work properly, I believe it is a good idea to prepare a variable such like `g:go_debug_map` . Thanks.

Carpetsmoker:
Hm, maybe I tested it wrong, but when I tried it, it didn't seem to work. I will look again later.

Carpetsmoker:
Okay, does seem to work. Dunno what I did wrong previously.

Should still make it configurable in the future. I'll make an issue for that, but this improves things so I'll merge it as well. Thanks :+1:

PR #1701, opened 2018-03-02
cmd: go 1.10 maintains a build cache now

fatih:
We don't need this hack anymore to speed up building.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -17,7 +17,7 @@ function! go#cmd#Build(bang, ...) abort
   let args =
         \ ["build"] +
         \ map(copy(a:000), "expand(v:val)") +
-        \ ["-i", ".", "errors"]
+        \ [".", "errors"]
 
   " Vim async.
   if go#util#has_job()


bhcleek:
Do we want to remove this already? For how long should we support optimizations for past releases of go?

fatih:
I usually trying to keep up with latest. It helps the community to upgrade to the latest version and also makes the life easier for us (like how Apple encourages people to upgrade to the latest iOS version). 

Of course not everyone can upgrade their Go version. We can check the version, like how we do it in `plugin/go.vim` but that will add additional latency and we need to add caching to mitigate it. I say let's merge this as this is added recently already.

Carpetsmoker:
But it's not like `go build -i` does any harm, right?

Either way, I think it's fine to both merge it and keep as-is, but I do think it would be a good idea to add an entry in the `BACKWARDS INCOMPATIBILITIES` section of the ChangeLog.

bhcleek:
Lgtm

fatih:
> But it's not like go build -i does any harm, right?

yeah definitely, but `:GoBuild` in its plan form implies that it only runs `go build`. But we change the semantics. Not that we didn't do it for other commands or whatsoever. Just with `1.10` this is not required anymore. Keeping things like this bloats the codebase with time because people tend to forget things in the long term. Merging this and adding an entry under `BACKWARDS INCOMPATIBILITIES`

PR #1702, opened 2018-03-04
Add build tags 'g:go_build_tags' for gogetdoc

larryzju:
Specify build tags for `gogetdoc` command, with the `g:go_build_tags` setting.

### DIFF
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -148,6 +148,11 @@ function! s:gogetdoc(json) abort
   let pos = shellescape(fname.':#'.offset)
 
   let cmd += ["-pos", pos]
+
+  if exists('g:go_build_tags')
+    let cmd += ["-tags", get(g:, 'go_build_tags')]
+  endif
+
   if a:json
     let cmd += ["-json"]
   endif


autoload/go/doc.vim
```diff
@@ -148,6 +148,11 @@ function! s:gogetdoc(json) abort
   let pos = shellescape(fname.':#'.offset)
 
   let cmd += ["-pos", pos]
+
+  if exists('g:go_build_tags')
+    let cmd += ["-tags", get(g:, 'go_build_tags')]
```
Carpetsmoker: Did you test that this will work with multiple build tags? I'm, not 100% sure, but I think you need to do:

    ['-tags', go#util#Shellescape(get(g:, 'go_build_tags'))]

So that `g:go_build_tags='foo bar'` will be:

    gogetdoc -tags 'foo bar' someident

Instead of:

    gogetdoc -tags foo bar someIdent

larryzju: Yes, you are right, sorry for that changes. I had read the doc about build constraints again.
And I saw your commit #1705. That's much better. Thanks for your advice.

Issue #1412, opened 2017-08-16
Running GoTest with g:go_term_enabled and splitright puts editor into insert mode


mightyguava:
### Behavior

When running `:GoTest` or other terminal-enabled commands, I expect a split pane to open and enter "insert" mode, so that I can press `q` to exit the terminal.

Instead, I observe that the split pane opens, but the editor pane that my code is in enters insert mode instead, and I have to exit insert mode, go to the terminal pane, and then press `:q` to close it.

### Steps to reproduce:

Run `:GoTest` in a Go file.

### Configuration

vimrc
```vim
call plug#begin('~/.vim/plugged')
Plug 'fatih/vim-go'
call plug#end()

set splitright
let g:go_term_enabled = 1
```

* vim version: NVIM v0.2.0
* vim-go version: `525fe8af97433c4a661de9548be5ed1c15451a0c`
* go version: `1.8.3 darwin/amd64`

mocheryl:
I'm having the same issue. This appeared due to changes in commit 4fda5960187dc038a35231b25da80a1008b25116 and consequentially 79152447b34bba8c8bb39d95fd279bae1f951326. By reverting them, I get back the expected behaviour.

Also by reverting I could still not replicate the problem the first commit solved. The current working directory always stays the same. Using neovim v0.2.2

fatih:
I had something similiar back then, but it would put me inside insert mode. The following (didn't tested) should get you out of insert mode:

```
autocmd BufWinEnter,WinEnter term://* stopinsert
```

I think you could easily do anything else instead of calling stopinsert, such as calling `feedkeys()` or calling `:q`, etc.. Can you please try it and let us know ?

mocheryl:
Maybe I wasn't clear enough in my previous comment, because there is a bit of a misunderstanding here.

The issue is not in not exiting insert mode after running a Go command (in this case `:GoTest`). That has been resolved some time after the initial comment of this issue (second commit mentioned in my previous comment). It's the whole user experience when dealing with the terminal. When opened, it used to automatically switch control to it and enter terminal (insert) mode so that you can issue commands to it (in this case hitting `q` to exit, close and restore control back to the original window) just as it was originally intended.

[...]

PR #1703, opened 2018-03-04
Some improvements for :GoRun

hiberabyss:
1. Now will close previous `:GoRun` window when run more than one time;
2. Fix bug when there is error after running `:GoRun`.

Bug Reproduce:
1. Create directory `/tmp/gotest`
2. Create file `/tmp/gotest/test.go` with following content:

```
package main

import "fmt"

func main() {
	abc
	greating := "hello, world!"
	fmt.Printf(greating)
}
```
3.  Run `cd /tmp/gotest && nvim test.go`;
4. Run `:GoRun`;

Following result will occur:
image

The root cause is that ` s:on_exit` close wrong window, since window is switched in `go#term#newmode`.

Thanks!

### DIFF
--- a/autoload/go/term.vim
+++ b/autoload/go/term.vim
@@ -18,6 +18,11 @@ function! go#term#newmode(bang, cmd, mode) abort
     let mode = g:go_term_mode
   endif
 
+  if exists("b:goterm_buf_id") && bufloaded(b:goterm_buf_id)
+    execute("bd! " . b:goterm_buf_id)
+    unlet b:goterm_buf_id
+  endif
+
   " execute go build in the files directory
   let l:winnr = winnr()
   let cd = exists('*haslocaldir') && haslocaldir() ? 'lcd ' : 'cd '
@@ -70,7 +75,9 @@ function! go#term#newmode(bang, cmd, mode) abort
   stopinsert
 
   if l:winnr !=# winnr()
+    let l:bufid = bufnr('%')
     exe l:winnr . "wincmd w"
+    let b:goterm_buf_id = l:bufid
   endif
 
   return id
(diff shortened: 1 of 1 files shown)


hiberabyss:
Hi,

Is there any suggestion for this PR?
Thanks!

hiberabyss:
Hi @bhcleek ,

1. Need to `cd` into the directory which contain the go file with error. This condition was missed in my previous bug reproduce description, I have updated it now.
2. The bug could also be reproduced with the latest `vim-go` master branch (d2b0a234).
3. My `g:go_term_mode` setting is `botright split`.
4. My neovim version is `v0.2.3-835-g739fb93a9`. Different neovim version should have no compact.

Thanks!

bhcleek:
With `g:go_term_mode` set to one of the options that starts with `botright` or `rightbelow`, I get a new location list every time `:GoRun` is executed. When `g:go_term_mode` is set to one of the options that starts with `leftabove` or `topleft`, I don't see any problems. I've documented what I discovered in #1720.

Interestingly, I can only get the terminal window that shown in your screenshot if I comment out https://github.com/fatih/vim-go/blob/master/autoload/go/term.vim#L118.

While I agree there is _a_ bug, I'm unable to duplicate the bug you're describing.

bhcleek:
@hiberabyss Can you give #1721 a run and see if it resolves the issue you're seeing? I think the source of the bug you're seeing is a race condition that #1721 should resolve.

hiberabyss:
Hi @bhcleek ,

#1721 has resolved the issue, thanks for your great work!

Currently, when `:GoRun` N times, there will be N new windows. I expect there will be only one running window even when run `:GoRun`  N times. Could this also be improved?

image

Thanks!

bhcleek:
I'm going to look at improving that experience when I do #1412 after #1721 is merged

PR #1705, opened 2018-03-05
Add build tags to many commands

### LINKED ISSUE #1464: Working in project that uses build tags
### Behavior

When working in files that have build tags, not much works. There is the command `:GoBuildTags` but it isn't pervasive. It would be nice if when you set that completion, go-to's, etc worked. This also applies to test files as well. If I set the build tag and run `:GoTestFunc` it would be nice for those to be applied rather than having to write `:GoTestFunc -tags <tag>` every time.

### Steps to reproduce:

1) Create two files, one guarded by a build tag `foo` and a main.go
2) Define a struct in the build tag version
3) Set `:GoBuilldTag foo`
4) Try using the struct in the main.go, autocomplete, goto, etc won't work.

### Configuration

* `vimrc`: Don't think it will be necessary to repro
* vim version: nvim v0.2.0
* vim-go version: Latest
* go version: 1.9.0

Carpetsmoker:
Fixes #1464

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -15,10 +15,8 @@ function! go#config#BuildTags() abort
 endfunction
 
 function! go#config#SetBuildTags(value) abort
-  if a:value == ""
-    if exists('g:go_build_tags')
-      unlet g:go_build_tags
-    endif
+  if a:value is ''
+    silent! unlet g:go_build_tags
     return
   endif
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ FEATURES:
 
 IMPROVEMENTS:
 
+* Add build tags (with `g:go_build_tags`) to all commands that support it.
+  [[GH-1705]](https://github.com/fatih/vim-go/pull/1705)
+
 BUG FIXES:
 
 * Update the correct window's location list after a long running async job
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -14,8 +14,8 @@ function! go#cmd#Build(bang, ...) abort
   " placeholder with the current folder (indicated with '.'). We also pass -i
   " that tries to install the dependencies, this has the side effect that it
   " caches the build results, so every other build is faster.
-  let args =
-        \ ["build"] +
+  let l:args =
+        \ ['build', '-tags', go#config#BuildTags()] +
         \ map(copy(a:000), "expand(v:val)") +
         \ [".", "errors"]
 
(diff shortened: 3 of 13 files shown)


fatih:
We should keep `:GoBuildTags`. :GoBuildTags is dynamic, and sometimes it's handy to have it. `go_build_tags` on the other hand is permanent, for people who always work on files that have it stick in their files.

Carpetsmoker:
> :GoBuildTags is dynamic, and sometimes it's handy to have it. go_build_tags on the other hand is permanent, for people who always work on files that have it stick in their files.

What do you mean with "dynamic" and "permanent" in this context? All `:GoBuildTags` does is set `go_build_tags`, so it behaves the same?

fatih:
Yeah I know that it sets it, but it's get resets when you restart vim. Hence the word dynamic. Of course it's not the best solution, because you override the prior value which was set via `go_build_tags`. The better solution would be to not override `go_build_tags` , but not sure if that is worth the complexity.  But the command is really handy, sometimes you see a new file that has a build tag and you want to change it quickly. I think we should have it, but maybe we should change how it behaves. Dunno.

bhcleek:
FWIW, I agree with @fatih; I use `:GoBuildTags` frequently. As a general guideline, anytime we expect users to be able to change a config value during a session, I'm in favor of providing a convenience function to set it.

Carpetsmoker:
Okay, fair enough. I added `:GoBuildTags` back. clarified the documentation, and rebased on mater after the config changes. Should be good to merge now.

Also fixes :GoBuildTags "" to clear the tags. Looks like this didn't work anymore (probably regression from Billy's config PR? Didn't investigate too deeply).

autoload/go/config.vim
```diff
@@ -15,10 +15,8 @@ function! go#config#BuildTags() abort
 endfunction
 
 function! go#config#SetBuildTags(value) abort
-  if a:value == ""
-    if exists('g:go_build_tags')
-      unlet g:go_build_tags
-    endif
+  if a:value is '' || a:value is '""' || a:value is "''"
```
bhcleek: The conversion from `''` and `""` to the empty string should happen in `go#cmd#BuildTags`. I'll submit a separate PR to do that so it's resolved separately...

Carpetsmoker: Alright, I merged your PR and rebased on master.

bhcleek:
assuming builds pass, lgtm

jdef:
My project doesn't use tags and I'm running golang 1.10. When I updated vim-go to `master` the `:GoCoverage` command complained that `tags` may only be specified once. When I rolled back `vim-go` to the commit just prior to his one, `:GoCoverage` worked again.

PR #1707, opened 2018-03-05
Use signs to keep track of breakpoints

### LINKED ISSUE #1688: Breakpoints are kept on the same line when lines are added removed
This causes issues after running `:GoDebugRestart`, as it will now be on an unexpected line.

Carpetsmoker:
Fixes #1688

Signs already get updated correct when adding/removing lines, so use
information from `:sign place` instead of keeping track of our own list
of breakpoints. This also makes sure that what the user sees always
matches with that Vim/Delve thinks should happen.

TODO: needs some more testing. I saw an error or two I didn't have
before.

### DIFF
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -16,7 +16,6 @@ if !exists('s:state')
   let s:state = {
       \ 'rpcid': 1,
       \ 'running': 0,
-      \ 'breakpoint': {},
       \ 'currentThread': {},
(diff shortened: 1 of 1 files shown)


fatih:
Another ping :)

Carpetsmoker:
I'm not going to continue working on this; I worked on the debugger initially because I thought it would be useful, but found that it's, well, not.

PR #1711, opened 2018-03-07
Adds vim-plug post-update hook to documentation

tutabeier:
I use `vim-plug` to manage my Vim plugins and I think we can include the post-update hook to the docs.


Carpetsmoker:
Will this get run on *install* or on every *update*? If it's the latter, then you will probably want to run `:GoUpdateBinaries` instead.

PR #1712, opened 2018-03-08
Consolidate configs

bhcleek:
Consolidate (nearly) all the config values into `autoload/go/config.vim`. The value of doing this is:
* Defaults don't have to be repeated in multiple places(e.g. avoid proliferation of `get(g:, "go_build_tags", "")`)
* Makes it easier to change default values (related to ☝️ ).
* There's a consistent pattern of using getters and setters instead of trying to apply a heuristic to determine when a configuration value should be given a getter and setter and then trying to monitor to make sure that value is always accessed via the getter and setter.

All getter and setter functions are in a new file, `autoload/go/config.vim`. All of the getters are named by stripping off the `g:go_` prefix, and then applying the conventional rules for converting snake case to PascalCase. This simple transformation, if we continue to follow the pattern it establishes, will make it easy on contributors to always know what the getter for a particular configuration value is named. Setters are named similarly, but are prefixed with `Set`.

I considered putting the getter and setter functions closer to where they're often used, but trying to do so would leave some unanswerable questions about what the getters for some values should be named.

This also sets us up to start preferring buffer, window, and tab local values over global values so that users can have fine grained control of the workspaces within a single session of Vim.

I did each of these in individual commits to make reviewing easier. I can squash it after reviews are complete if that's desirable.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -21,7 +21,7 @@ function! go#cmd#Build(bang, ...) abort
 
   " Vim async.
   if go#util#has_job()
-    if get(g:, 'go_echo_command_info', 1)
+    if go#config#EchoCommandInfo()
       call go#util#EchoProgress("building dispatched ...")
     endif
 
@@ -33,7 +33,7 @@ function! go#cmd#Build(bang, ...) abort
 
   " Nvim async.
   elseif has('nvim')
-    if get(g:, 'go_echo_command_info', 1)
+    if go#config#EchoCommandInfo()
       call go#util#EchoProgress("building dispatched ...")
     endif
 
@@ -171,7 +171,7 @@ function! go#cmd#Install(bang, ...) abort
     " expand all wildcards(i.e: '%' to the current file name)
     let goargs = map(copy(a:000), "expand(v:val)")
 
-    if get(g:, 'go_echo_command_info', 1)
+    if go#config#EchoCommandInfo()
       call go#util#EchoProgress("installing dispatched ...")
     endif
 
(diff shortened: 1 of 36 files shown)


fatih:
Let's merge this after a new release. This is a big change and I don't want it a part of the upcoming release.

bhcleek:
SGTM

bhcleek:
rebased to resolve a conflict

Carpetsmoker:
> Let's merge this after a new release. This is a big change and I don't want it a part of the upcoming release.

That were my thoughts exactly as well. I thought I had commented on this actually, but guess that happened only in my mind and not in reality >_<

autoload/go/cmd.vim
```diff
@@ -170,7 +171,7 @@ function! go#cmd#Install(bang, ...) abort
     " expand all wildcards(i.e: '%' to the current file name)
     let goargs = map(copy(a:000), "expand(v:val)")
 
-    if get(g:, 'go_echo_command_info', 1)
+    if getgo#config#EchoCommandInfo()
```
Carpetsmoker: `getgo` seems wrong here, I think?

autoload/go/fmt.vim
```diff
@@ -254,13 +238,13 @@ function! s:show_errors(errors) abort
...
-  if get(g:, "go_fmt_autosave", 1)
-    let g:go_fmt_autosave = 0
+  if go#config#FmtAutosave()
+    call go#config#FmtAutosave(0)
     call go#util#EchoProgress("auto fmt disabled")
     return
   end
 
-  let g:go_fmt_autosave = 1
+	call go#config#FmtAutosave(1)
```
Carpetsmoker: Minor nitpick, but for some reason it looks like a tab snuck in here?

PR #1713, opened 2018-03-08
cache import path lookups results

bhcleek:
Cache the results of import path lookups so that go list doesn't have to
be run every time linting is run or go#guru#Sameids() is run, because go
list can take a noticeably long time to return results and has to be
executed synchronously for all of its current use cases.

This further reduces the likelihood that characters will be rendered at
the cursor position when in normal mode.


fatih:
lgtm

PR #1716, opened 2018-03-12
Fix go build error parse

bhcleek:
Errors from `:GoBuild` were including the column number in the quickfix error's `text` message instead of it parsing it into in the quick fix error's `col` field, because `errorformat` wasn't being using used parse the errors when `go build` was run asynchronously.

This PR uses `go#list#Parse` to parse the errors so that the result in vim8 and Neovim is the same as it is in vim7.4.

While I was at it, I added a test for `:GoVet` and modified it to also use `go#list#Parse`.

### DIFF
--- a/autoload/go/cmd_test.vim
+++ b/autoload/go/cmd_test.vim
@@ -6,7 +6,7 @@ func! Test_GoBuildErrors()
 
     " set the compiler type so that the errorformat option will be set
     " correctly.
-    compiler! go
+    compiler go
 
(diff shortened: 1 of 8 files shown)


bhcleek:
`:GoBuild` isn't dropping the first line comment unless I run `compiler! go` first. Feel free to review ahead of a fix for that, but this won't be done until I fix that.

bhcleek:
☝️ fixed.

fatih:
lgtm

PR #1721, opened 2018-03-19
Fix various race conditions when using terminal windows

### LINKED ISSUE #1720: Using :GoRun in terminal mode N times opens N location list windows
```
#!/bin/sh

DIR=$(mktemp -d -t vimgo-term-mode)
printf "using %s\n" "$DIR"

cat <<EOF > "$DIR/main.go"
package main

import "fmt"

func main() {
	abc
	greating := "hello, world!"
	fmt.Printf(greating)
}
EOF

cd "$DIR"
nvim -c 'let g:go_term_mode="botright split"' main.go
```
And then run `:GoRun` twice.

A single location list window to display the error reading `undefined: abc`.

### What happened instead?
Two location windows display the error reading `undefied: abc`.

NeoVim v0.2.2

* Go version (`go version`):
1.9.4

* Go environment (`go env`):

* vim-go version:
master as of `d2b0a234ffb5441a3488c78fe8e5f551ddbdd454`

bhcleek:
* Explicitly close the terminal window instead of assuming the terminal
  window is the current window and closing that.
* Eliminate the outstanding job list in favor of using a more object
  oriented like approach.
* Make sure the correct location list gets used instead of assuming that
  the current window is the window whose location list should be used.

Fixes #1720

### DIFF
--- a/autoload/go/job.vim
+++ b/autoload/go/job.vim
@@ -64,7 +64,7 @@ function go#job#Spawn(args)
   function! s:callback(chan, msg) dict
     call add(self.messages, a:msg)
   endfunction
-  " explicitly bind callback so that to state so that within it, self will
+  " explicitly bind callback to state so that within it, self will
   " always refer to state. See :help Partial for more information.
   let cbs.callback = function('s:callback', [], state)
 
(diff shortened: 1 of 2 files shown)


fatih:
tested it with nvim. lgtm

PR #1722, opened 2018-03-19
close the correct location list

bhcleek:
Make sure the correct location list is cleaned when terminal mode has no
output.


fatih:
Seems like this is a very common thing:
screen shot 2018-03-20 at 9 31 13 am

I'll merge this and open another PR that provides a single public function, such as `go#list#Clear()`

fatih:
Opened PR here: https://github.com/fatih/vim-go/pull/1724

PR #1724, opened 2018-03-20
list: make Clean() more useful

fatih:
After cleaning a list, there is no need to show the list anymore. That's
why we call `Window()` to invoke the close function. Remove all
`Window()` calls and make Clean() more useful to avoid adding the second
line. There is no single `Clean()` call independently, so it makes sense
to change the meaning

### DIFF
--- a/autoload/go/list.vim
+++ b/autoload/go/list.vim
@@ -107,12 +107,21 @@ function! go#list#JumpToFirst(listtype) abort
   endif
 endfunction
 
-" Clean cleans the location list
+" Clean cleans and closes the location list 
 function! go#list#Clean(listtype) abort
   if a:listtype == "locationlist"
     lex []
   else
     cex []
+  endif
+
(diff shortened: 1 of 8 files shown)


bhcleek:
LGTM

autoload/go/list.vim
```diff
@@ -107,13 +107,22 @@ function! go#list#JumpToFirst(listtype) abort
...
   endif
+
+  let autoclose_window = get(g:, 'go_list_autoclose', 1)
+  if autoclose_window
+    if a:listtype == "locationlist"
+      lclose
+    else
+      cclose
+    endif
+  endif
```
bhcleek: Instead of adding this block, calling `go#list#Window(a:listtype)` would reduce code duplication...

fatih: Agree, but `Window()` is doing something else as well and it's get bloated. I'll add a `Close()` function and move out the logic there.

PR #1725, opened 2018-03-20
Fix term enabled behavior

bhcleek:
* Make sure that behavior is consistent when `g:go_term_enabled=1` regardless of the value of the `splitright` option.

* Remove the startinsert and stopinsert pair; they cancel each other out
and are unnecessary.

* remove superfluous stderr callback.

### DIFF
--- a/autoload/go/term.vim
+++ b/autoload/go/term.vim
@@ -18,7 +18,7 @@ function! go#term#newmode(bang, cmd, mode) abort
   let state = {
         \ 'cmd': a:cmd,
         \ 'bang' : a:bang,
-        \ 'winnr': winnr(),
+        \ 'winid': win_getid(winnr()),
         \ 'stdout': []
       \ }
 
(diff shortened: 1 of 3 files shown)


autoload/go/term.vim
```diff
@@ -67,11 +66,7 @@ function! go#term#newmode(bang, cmd, mode) abort
   " we also need to resize the pty, so there you go...
   call jobresize(state.id, width, height)
 
-  stopinsert
-
-  if state.winnr !=# winnr()
-    exe state.winnr . "wincmd w"
-  endif
+  call win_gotoid(state.winid)
```
fatih: is this a native nvim function?

bhcleek: It is. We actually probably need to be using it many other places, but all the other cases where it should be used but isn't requires an unlikely edge case before a user would notice a problem. Because of the way the terminal works, though, we must use the window id instead of the window number to make sure the behavior is the same regardless of the value of `splitright`.

fatih: Thanks for the clarification.

fatih:
lgtm

Issue #1690, opened 2018-02-27
Can't debug some test files (?)


Carpetsmoker:
It works for some test files, but not this particular one?

```
[~tw/kommentaar](master)% vi parse/parse_test.go
```

`:GoDebugStart` gives me:

```
vim-go: Starting GoDebug...
vim-go: 2018/02/27 19:26:12 debugger.go:98: launching process with args: [/tmp/vNu86sR/2]
vim-go: could not launch process: fork/exec /tmp/vNu86sR/2: no such file or directory
vim-go: could not remove /tmp/vNu86sR/2: remove /tmp/vNu86sR/2: no such file or directory
vim-go: 2018/02/27 19:26:12 debugger.go:98: launching process with args: [/tmp/vNu86sR/2]
vim-go: could not launch process: fork/exec /tmp/vNu86sR/2: no such file or directory
vim-go: could not remove /tmp/vNu86sR/2: remove /tmp/vNu86sR/2: no such file or directory
```

Seems to work on `parse/parse.go`, as well as on simpler test files (although this is not a particularly complicated file?)

This project isn't public, but just for my own future reference it can be reproduced with commit 5309c47

Carpetsmoker:
I discovered the problem is how I opened the files. This fails

```
[~tw/kommentaar](master)% vi parse/parse_test.go
```

but this works:

```
[~tw/kommentaar](master)% cd parse
[~tw/kommentaar/parse](master)% vi parse_test.go
```

fatih:
I see. I use the following in my vimrc:

```
" Enter automatically into the files directory
autocmd BufEnter * silent! lcd %:p:h
```

as I prefer to be in the directory of the file. Seems like this could be solved in a hacky way by cd'ing into the folder before we call delve.

Carpetsmoker:
> Seems like this could be solved in a hacky way by cd'ing into the folder before we call delve.

Maybe for test files. It's intentional that we don't do this for regular Go files. Consider a package `foo` that has `package main`, and you load `subdir/sub.go`. You will usually want to debug the `foo` package, and not the `foo/subdir` package (Delve can't even run that as that's a non-main package).

[...]

PR #1728, opened 2018-03-21
Add :GoDebugTest

Carpetsmoker:
This fixes the two issues reported in #1690.

The first issue is fixed by cd-ing in to directory before debugging
test files. This is also what `go test` does ad seems to work well
in my testing.

---

The second issue is fixed by adding `:GoDebugTest`:

Imagine I'm editing `a_test.go`, which is failing. To investigate I want
to add a breakpoint in a function in `a.go`. The current procedure for
that is something like:

- `:e a.go`
- `:GoDebugBreakpoint`
- `:e a_test.go`
- `:GoDebugStart`
- `<F5>`
- Buffer will switch back to `a.go`

The `:e_test.go` is needed because the only way to run `dlv test`
instead of `dlv debug` is to have the current buffer end in `_test.go`.

I think this is rather awkward, so add `:GoDebugTest` instead of trying
to guess what the users wants so you now can do:

- `:e a.go`
- `:GoDebugBreakpoint`
- `:GoDebugTest`
- `<F5>`

### DIFF
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -21,6 +21,7 @@ if !exists('s:state')
       \ 'localVars': {},
       \ 'functionArgs': {},
       \ 'message': [],
+      \ 'is_test': 0,
       \}
 
   if go#util#HasDebug('debugger-state')
@@ -521,6 +522,8 @@ function! go#debug#Start(is_test, ...) abort
     lcd %:p:h
   endif
 
+  let s:state.is_test = a:is_test
+
   let dlv = go#path#CheckBinPath("dlv")
   if empty(dlv)
     return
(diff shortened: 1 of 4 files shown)


Carpetsmoker:
I also added a commit to fix #1729. I originally fixed this by adding a breakpoint to the nearest function start, but then changed my mind and decided that just always adding a breakpoint on the current line makes much more sense and is a lot easier. I kept both commits so the code isn't lost in case we change behaviour after all.

fatih:
> The :e_test.go is needed because the only way to run dlv test
instead of dlv debug is to have the current buffer end in _test.go.

Can't we just check the suffix and run `dlv debug` or `dvl test` automatically? If I've opened a test file, and type `:GoDebugStart` it would mean that I want to debug the test file. So `dlv test` should be called. Same for a non `_test.go` file, where we would call `dlv debug`. Can we do this automatically? Otherwise I imagine I would add something like:

```
function! s:debug_go()
  let l:file = expand('%')
  if l:file =~# '^\f\+_test\.go$'
    call go#debug#Start(1, <f-args>)
  elseif l:file =~# '^\f\+\.go$'
    call go#debug#Start(0, <f-args>)
  endif
endfunction

autocmd Filetype go command! -bang GoDebug :<C-u>call <SID>debug_go()<CR>
```

I know we have distinct `:GoBuild` and `:GoTest` commands, but those are two different actions being done. For debugging I think we should just have a single command. Wdyt?

Carpetsmoker:
Yes, that is the solution we have now, but it doesn't work as the buffer you have open isn't always indicative of what you want to do. Certainly in my own usage I rarely want to place breakpoints in test files, and always in "regular" Go files, and found it works very poorly.

fatih:
So to recap, If I set a breakpoint to a regulat file and I want to debug the test file, that's not possible right now? Trying to understanding the issue here. 

> but it doesn't work as the buffer you have open isn't always indicative of what you want to do. 

Why not? We know that the user has set a breakpoint somewhere but wants to test the current buffer. Or isn't that the case always? Do you mean if we set a breakpoint in a regular file and I'm inside a test file, it means I want to debug the regular file?

Carpetsmoker:
> So to recap, If I set a breakpoint to a regulat file and I want to debug the test file, that's not possible right now?

It's possible, but you need to switch to a `_test.go` buffer first, which is annoying. See the commit/PR description.
It's like combining `:GoInstall` and `:GoTest` in one command: just having `a.go` open doesn't really say anything about what you're doing: you *could* be fixing a test and want to run `go test` (and not `install`).

fatih:
> It's possible, but you need to switch to a _test.go buffer first, which is annoying

Ok so what's the problem then we we call `dvl test` under the hood if we detect that the buffer file name ends with `_test.go`? I think I still miss something here

Carpetsmoker:
You need to do an extra `:e a_test.go` before being able to start tests. I don't know how to better describe it :-(

Carpetsmoker:
> I already have opened the file, I'm inside a_test.go file

No, that's the point: you're often not. How often do you want to place breakpoints inside a `_test.go` file? Probably not very often, you usually want to place a breakpoint in the regular `a.go` file that the `_test.go` file is testing. So you're *actually* in the `a.go` file most of the time (that is my experience anyway).

It's like `:GoTest`, you will often run this while you're fixing a bug in the implementation or changing the implementation, so you're in the `a.go` file and not `_test.go` file.

Carpetsmoker:
In principle I agree that the current behaviour (that checks the buffer's file name) is better, which is why I wrote it like this originally. It's just that in my testing it works very poorly/cumbersome in this case. If it's not clear what I mean then maybe I can record a video later to show the workflow and problems with it.

fatih:
> In principle I agree that the current behaviour (that checks the buffer's file name) is better, which is why I wrote it like this originally. It's just that in my testing it works very poorly/cumbersome in this case.

I understand what you mean now. I agree that we should have a dedicated debug command for tests. It's not possible to obtain the preference of the user by assuming they want to debug the test or the file.

PR #1730, opened 2018-03-21
Add codecov.io configuration

bhcleek:
* do not post comments to pull requests
* ignore coverage of test files.
* allow checks to pass with up to 1% reduction in code coverage.


.codecov.yml
```diff
@@ -0,0 +1,12 @@
+---
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 1
+        base: auto
+comment: false
```
Carpetsmoker: You should also be able to make the comment smaller with something like:

```
comment:
  layout: "diff"
```

Either `diff` or `files` is probably the most useful part.

See: https://docs.codecov.io/docs/pull-request-comments

.codecov.yml
```diff
@@ -0,0 +1,12 @@
...
+ignore:
+  - "autoload/go/*_test.vim"
+  - "autoload/gotest.vim"
```
Carpetsmoker: Probably want to ignore everything that doesn't match `*.vim` (or `*.vim~*_test.vim`)? Looks like you can use regexps: https://docs.codecov.io/docs/ignoring-paths

Carpetsmoker: Actually, I think we can ignore everything outside `autoload/go`? We have one file outside of that right now (`test/gopath_test.vim`), but that can/should probably be moved to `autoload/go/path_test.vim`.

PR #1733, opened 2018-03-21
redraw before echo

### LINKED ISSUE #1732: Unexpected `Press ENTER or type command to continue` message with `:GoCoverage`
### What did you do?
I executed `:GoCoverage`

Get coverage info (colours) in my open buffers

### What happened instead?
I first need to press ENTER or a command to see the coverage:
image

### Configuration:
```
let g:go_def_mapping_enabled = 0
let g:go_fmt_command = "goimports"
let g:go_fmt_fail_silently = 1
let g:go_list_type = "quickfix"
let g:go_highlight_build_constraints = 1
let g:go_highlight_generate_tags = 1
let g:go_highlight_function_calls = 1
let g:go_highlight_operators = 1
let g:go_snippet_engine = "neosnippet"
let g:go_statusline_duration = 10000
```

```
NVIM v0.2.2
Build type: Release
```

bhcleek:
Redraw before echoing a progress message.

Fixes #1732

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -313,6 +313,7 @@ function! s:echo(msg, hi)
   " Tabs display as ^I or <09>, so manually expand them.
   let l:msg = map(l:msg, 'substitute(v:val, "\t", "        ", "")')
 
+  redraw
   exe 'echohl ' . a:hi
   for line in l:msg
     echom "vim-go: " . line


fatih:
lgtm

svanharmelen:
Very nice @bhcleek! Thanks!

Carpetsmoker:
Even more important, this will cause you to lose messages:

    fun! XXX()
        call go#util#EchoError("a")
        call go#util#EchoError("b")
    endfun

On `call XXX()` I see just:

    vim-go: b

No way to tell that `vim-go: a` was also there. If I revert the `:redraw` I get:

    vim-go: a
    vim-go: b

The prompt is there because there are multiple messages to display. I think this is a feature, and I don't think we should work around it. I don't know what the best solution is here, but I don't think this is it.

bhcleek:
I disagree. Showing requiring the user to press Enter just because there are multiple progress messages is a bad user experience that we should try to avoid. I'll grant you that if there are multiple errors that perhaps they should see them all, but I'm not convinced, especially since it's possible to still see them all using `:messages`. In any case, I'll submit a PR shortly to scope the redraw to the `EchoProgress` only and we can hash out the larger question later.

fatih:
I think it might due the message. For example in `gorename` we have this:

```
  " strip out newline on the end that gorename puts. If we don't remove, it
  " will trigger the 'Hit ENTER to continue' prompt
  call go#list#Clean(l:listtype)
  call go#util#EchoSuccess(a:out[0])
```

if redraw makes thing worse I'm in favor of removing it. I think the better solution is to check why there are two testing messaging showing up. Imho that's wrong and shouldn't happen. I don't think showing the `Hit Enter...` prompt is user friendly, but I believe there might be better fixes for this.

bhcleek:
The problem in this case was that `go#coverage#Buffer` outputs a message, `testing...` and then calls `go#test#Test`, which does the same. While it's certainly reasonable to remove the progress message in `go#coverage#Buffer`, that only solves this specific problem. Redrawing before showing a progress message solves the general problem, which is what I was going for 🤷‍♂️

fatih:
> Redrawing before showing a progress message solves the general problem

Yeah this makes sense in that context, but if this causes problem on other's people machines I think it's wiser to remove the message from the coverage file.

@Carpetsmoker can you try out the new PR that Billie has sent? if it doesn't cause any problems I suggest merging it and go forward, if not let's remove the message in `coverage#Buffer` as that fixes the problem the other way.

Carpetsmoker:
> Requiring the user to press Enter just because there are multiple progress messages is a bad user experience

I don't disagree with that, but I think hiding messages is even worse. If a message isn't worth being shown then perhaps not show it?

PR #1734, opened 2018-03-21
use window id instead of number in all async jobs

bhcleek:
Use window ids instead of window numbers to select windows in all async
job callbacks so that the correct window is selected even when the user
modifies the window layout between when the job is started and when the
callback is executed.

### DIFF
--- a/autoload/go/job.vim
+++ b/autoload/go/job.vim
@@ -33,7 +33,7 @@
 function go#job#Spawn(args)
   let cbs = {}
   let state = {
-        \ 'winnr': winnr(),
+        \ 'winid': win_getid(winnr()),
         \ 'dir': getcwd(),
         \ 'jobdir': fnameescape(expand("%:p:h")),
         \ 'messages': [],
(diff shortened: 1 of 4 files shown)


autoload/go/job.vim
```diff
@@ -104,15 +104,20 @@ function go#job#Spawn(args)
   let cbs.close_cb = function('s:close_cb', [], state)
 
   function state.show_errors(job, exit_status, data)
+    let l:winid = win_getid(winnr())
+    call win_gotoid(self.winid)
+
     let l:listtype = go#list#Type(self.for)
     if a:exit_status == 0
       call go#list#Clean(l:listtype)
+      call win_gotoid(l:winid)
```
fatih: This is slightly confusing. We jump two times.  First to the windows where the job was started, and then to the current window. Why do we have a logic like this? If we jump to the window the job was started, I wouldn't want to go back, but I think here we jump back because there are no errors right ? Maybe we should have an `if .. else ...` clause instead of jumping twice.

bhcleek: Assume the user has 2 windows open. While in window 1, the user runs an async command whose list type is `locationlist`. Before the async command completes, the user moves to window 2. Then the async command completes successfully. We need to clean the list of window 1 (hence the first jump). And we want to avoid stealing window focus, so we make sure to make window 2 active again before returning from the callback.

PR #1736, opened 2018-03-23
redraw before showing progress message only

bhcleek:
Do not redraw before showing all messages; only progress messages will
cause a redraw.

Convert the progress message in `go#coverage#Buffer` to a call to
`go#util#EchoProgress`.


Carpetsmoker:
This is still problematic in cases where you have an error followed by a progress message; but it's a lot better, and also solves the flickering issues I mention in the other PR.

PR #1738, opened 2018-03-23
Add :GoIssue to open bug reports on GitHub for vim-go

bhcleek:
Add a new command, `:GoIssue`, to make it easier for users to open bug reports with most system information filled in.

I'm not completely happy with the name of the command and am open to other suggestions. I thought about using `:GoBug`, but that may lead people to believe that it's `go bug`. `:VimGoBug` also doesn't seem right, because none of the other commands start with `VimGo` :thinking:.

### DIFF
--- a/autoload/go/issue.vim
+++ b/autoload/go/issue.vim
@@ -1,0 +1,34 @@
+let s:templatepath = go#util#Join(expand('<sfile>:p:h:h:h'), '.github', 'ISSUE_TEMPLATE.md')
+
+function! go#issue#New() abort
+  let body = substitute(s:issuebody(), '[^A-Za-z0-9_.~-]', '\="%".printf("%02X",char2nr(submatch(0)))', 'g')
+  let url = "https://github.com/fatih/vim-go/issues/new?body=" . l:body
+  call go#tool#OpenBrowser(l:url)
+endfunction
+
+function! s:issuebody() abort
+  let lines = readfile(s:templatepath)
+
+  let rtrimpat = '[[:space:]]\+$'
+  let body = []
+  for l in lines
+    let body = add(body, l)
+
+    if l =~ '^\* Vim version'
+      redir => out
+        silent version
+      redir END
+      let body = extend(body, split(out, "\n")[0:2])
+    elseif l =~ '^\* Go version'
+      let [out, err] = go#util#Exec(['go', 'version'])
+      let body = add(body, substitute(l:out, rtrimpat, '', ''))
+    elseif l =~ '^\* Go environment'
+      let [out, err] = go#util#Exec(['go', 'env'])
+      let body = add(body, substitute(l:out, rtrimpat, '', ''))
+    endif
+  endfor
+
+  return join(body, "\n")
+endfunction
+
+" vim: sw=2 ts=2 et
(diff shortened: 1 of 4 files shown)


fatih:
This looks interesting :) Instead of copy/paste we could point them to call this command. I definitely thing it should be `:GoReportIssue`, `:GoReportBug` or even `:GoReportGithubIssue` I think the latter is much more clear, it's not short so people can't complete accidentally as well. Because it's something that is never called or just once, it's ok to have a long command name.

bhcleek:
PTAL

I also have some code locally that will include the values of all the `g:go_*` variables. We often need those when diagnosing an issue, but I was hesitant about including them. Do you think the report should include them?

autoload/go/issue.vim
```diff
@@ -0,0 +1,34 @@
+let s:templatepath = go#util#Join(fnamemodify('<sfile>', ':p:h'), '.github', 'ISSUE_TEMPLATE.md')
```
Carpetsmoker: This doesn't seem to work:

```
Error detected while processing function go#issue#New[1]..<SNR>104_issuebody:
line    1:
E484: Can't open file /home/martin/work/src/github.com/teamwork/desk/.github/ISSUE_TEMPLATE.md
```

The (only) file I have open is `/home/martin/work/src/github.com/teamwork/desk/main.go`.

autoload/go/issue.vim
```diff
@@ -0,0 +1,34 @@
...
+
+  let rtrimpat = '[[:space:]]\+$'
+  let body = []
+  for l in lines
+    let body = add(body, l)
+
+    if l =~ '^\* Vim version'
+      redir => out
+        silent version
+      redir END
```
Carpetsmoker: It is probably better to use `let l:out = execute('silent version')` instead of `redir`. `redir` looks nice, but there some issues when you run multiple `redir`s at the same time (e.g. nesting, from autocmds) which can result in very strange and confusing errors.  
Since we don't control the entire setup (other plugins, people's vimrcs, etc.) it's usually best to avoid `redir` unless there is no other way.

autoload/go/issue.vim
bhcleek: How did you source the file?

autoload/go/issue.vim
bhcleek: Under what conditions would there be another `redir` being executed?

autoload/go/issue.vim
bhcleek: hmm, I can duplicate what you're seeing. I'll investigate later.

autoload/go/issue.vim
bhcleek: Given that vim 7.4 doesn't have `execute()`, that doesn't seem like a viable option until we drop support for vim 7.4.

bhcleek:
As I mentioned in https://github.com/fatih/vim-go/pull/1738#issuecomment-375685493, I'm a little reluctant to add support for dumping all of the vim-go config values out. Let's get this done first, and then talk about how to extend it.

autoload/go/issue.vim
Carpetsmoker: Oh, that was added in 7.4.2008, which is just a bit after the 7.4.1689 we support – thought it was earlier.

> Under what conditions would there be another redir being executed?

Mostly in cases with autocmds or another function calling this. You can never really know, that's the thing.

bhcleek:
Thanks for the double-check @Carpetsmoker . I'm not quite sure how I managed to mess that up before pushing. Fixed now, though.

PTAL

PR #1741, opened 2018-03-27
Add q to close doc window

damnever:
Or, make this configurable?


Carpetsmoker:
Personally I don't think that adding an ever-growing list of key mappings is a good idea. Arguably, mapping `<Esc>` and `<CR>` is *already* not a good idea.

The `filetype` is set to `godoc`, so you can use an `autocmd` to map your own mappings:

    autocmd Filetype godoc noremap <buffer> <silent> q :<C-U>close<CR>

Issue #1742, opened 2018-03-27
:GoDebugStart doesn't show a useful message if the build fails


fatih:
If the file doesn't compile, `:GoDebugStart` doesn't start and exits with the message `exit status 2`. When called `:messages` we can see the error message there which contains the actual error message from the build output. We should show a better error message instead of showing `exit status 2`.

Carpetsmoker:
Even worse is that this is shown as EchoProgress, instead of EchoError. I at least fixed that for now (see PR). We can populate the quickfix list in a future PR I think.

PR #1743, opened 2018-03-27
Show debug errors as errors and not progress

Carpetsmoker:
Show stderr as error, not progress.

It *still* only shows the last time, but that's the nature of outputting
stuff in a job callback: you don't get the "hit ENTER" prompt (which in
this case is *wanted*). I don't know if there is any way to fix that
(it's a problem we have in several commands)?

We can probably populate the quickfix like we do with `:GoInstall` and
such, but that's a PR for another day.

See #1742


fatih:
lgtm

PR #1746, opened 2018-03-28
Allow configurable `go_templates_dir`

bramswenson:
Defaults to the normal path within the plugin, but allows any other path to be set as the template directory.


Carpetsmoker:
Thanks for your PR!

I think it makes more sense to allow setting `g:go_template_file` and `g:go_template_test_file` to an absolute path, rather than adding a new setting. For example:

    let g:go_template_file = '~/.vim/templates'     " Need to call expand()
    let g:go_template_file = '/data/templates/go'

What do you think?

bhcleek:
Closing due to inactivity.

PR #1753, opened 2018-04-04
Write buffer before running GoDebugStart

### LINKED ISSUE #1744: Starting the debugger on a buffer with unsaved changes
If you use `:GoDebugStart` then it will always operate on the files on disk, and not in the buffer. So if you have unsaved changes then this can be rather confusing.

Copying to a tmp dir isn't really an option, since the debugger operates on packages rather than files. We don't want to copy entire packages.

I'm not sure what the best course of action is here? Maybe just do nothing and leave it as is? Or silently write buffers? Or show a warning? Or something else?

Carpetsmoker:
Fixes #1744. Second part is that it should probably warn/error
if `autowrite` is disabled and there are unsaved changes, but that
is a more general issue that also applies to other commands.

### DIFF
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -490,6 +490,8 @@ endfunction
 " Start the debug mode. The first argument is the package name to compile and
 " debug, anything else will be passed to the running program.
 function! go#debug#Start(is_test, ...) abort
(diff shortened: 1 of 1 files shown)


bhcleek:
Lgtm

PR #1754, opened 2018-04-04
Warn when running go#cmd#autowrite() without autowrite enabled

Carpetsmoker:
The `go#cmd#autowrite()` will write buffers to disk for commands that
operate on files (rather than buffers). This is useful, but if you
*don't* have this option enabled you never get an error message.

This changed that to show a non-fatal warning. This is the same
behaviour as the `:make` command.

Right now this doesn't work very well because the various status
messages from the jobs push the error away, making it hard to see (I
don't get the expected "hit enter" prompt), so we need a better solution
for that first.


Carpetsmoker:
I added a 1 second sleep to make the error message a bit more visible. It's a bit meh, but not sure how to do it better without completely revamping the way we echo messages.

bhcleek:
1 second for _each_ file that's modified could be excessive. Maybe do that _after_  iterating if any buffers are modified?

Carpetsmoker:
Ah yes, that is a bit silly.

I changed it to show the message once only, which is also what `:make` does. Showing the same message more than once doesn't really add any value.

bhcleek:
Does the message disappear due to a redraw without the sleep? Assuming yes, lgtm.

Carpetsmoker:
> Does the message disappear due to a redraw without the sleep

Not due to redaw as such, but because `echo` from an async callback will just override the contents, instead of invoking the "hit ENTER" prompt like when you run `echo` twice.

bhcleek:
Ship it.

autoload/go/cmd.vim
```diff
@@ -1,6 +1,17 @@
...
   if &autowrite == 1 || &autowriteall == 1
     silent! wall
+  else
+    for l:nr in range(0, bufnr('$'))
+      if buflisted(l:nr) && getbufvar(l:nr, '&modified')
+        " Sleep one second to make sure people see the message. Otherwise it is
+        " often immediacy overwritten by the async messages (which also don't
+        " invoke the "hit ENTER" prompt).
+        call go#util#EchoWarning('[No write since last change]')
+        sleep 1
```
fatih: Will sleep block any other running async calls? Seems like a dangerous thing to add in that case.

PR #1756, opened 2018-04-05
Use Go 1.10 in tests

Carpetsmoker:
Also set `let g:go_echo_command_info = 0` and add some `silent`s to make
normal (non-failing) test output less noisy.


bhcleek:
lgtm

PR #1761, opened 2018-04-05
add tabstop on err

ldelossa:
Suggestion for a more robust errn, snippet. This arose by the need to format the error message at times. For example

if err != nil {
  return nil, fmt.Errorf("some context: %s", err)
}

The snippet will allow you to tabstop to the err keyword for replacement.


Carpetsmoker:
Seems like a good idea. Thanks!

PR #1765, opened 2018-04-06
snippets per Go spec

ldelossa:
Branch begins to create compossible snippets using the Go spec as reference. Idea is to have snippets be able to be nested when creating structure like Switches. For sample

switch\<tab\> creates

```
switch {
  case "somecase":
    someaction()
}
```

When inside this switch case doing the following creates another case statement

```
switch {
  case "somecase":
    someaction()
  case\<tab\>
}
```

There's still some snippets that need to be defined but the majority of the Go spec is accounted for.

One nasty issue is that SuperTab begins to evaluate gocode completion when inside brackets. So unfortunately this was not possible for me to do:

```
al\<tab\>

[]Bool{t\<tab\>, t\<tab\>}
```

We would expect t<tab> to expand to "true" however supertab kicks in with omni complete. Maybe this can be fixed by changing my trigger key for Utlisnips - not sure.

### DIFF
--- a/gosnippets/UltiSnips/go.snippets
+++ b/gosnippets/UltiSnips/go.snippets
@@ -264,7 +264,7 @@ case ${1:x}:
 	${2:action}
 endsnippet
 
-snippet default "default entry"
+snippet cde "case default entry" m
 default:
 	${1:action}
 endsnippet
@@ -306,8 +306,8 @@ endsnippet
 ## Select statements
 snippet select "select statement"
 select {
-case ${1:x} ${2::=} ${3:<-c}:
-	${4:action}
+case ${1:x} ${2::=} ${3:<-}${4:c}:
+	${5:action}
 }
 ${0}
 endsnippet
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
Changing the entire file in one go makes it very hard to see what got added/changed/removed.

ldelossa:
@Carpetsmoker yes I see, however it is a little difficult to keep the structure of the initial file while trying to make a "markdown-ish" layout that can be searched. Do you suggest I drop the layout idea?

Carpetsmoker:
Just split out the changes in separate commits or PRs: if you first make one to change the order of things (but not add/remove/change any functionality), then the change in other commits/PRs is actually reviewable.

Right now the diff is all over the place: I see [opening_par removed here](https://github.com/fatih/vim-go/pull/1765/files#diff-35d0e56aade2bfcbce075c538827af8dL449), as well as the two snippets that rely on it. What happened to that? Is that rewritten to something else? Accidentally removed? Something else? Very difficult to see.

And that just so happened to be the first example I search. I picked another example earlier, and that was removed as well.

ldelossa:
Sounds good, I'm going to close this one and open another more comprehendible one.

PR #1774, opened 2018-04-07
Add ./scripts/bench-syntax

Carpetsmoker:
A script to benchmark the `syntax/go.vim` file. It's currently rather
slow if you enable all the `g:go_highlight` options even on modern
machines. This will allow us to easily do performance measurements.


fatih:
sgtm

PR #1775, opened 2018-04-07
Go spec snippets

ldelossa:
PR begins to create compossible snippets using the Go spec as reference. Idea is to have snippets be able to be nested when creating structure like Switches. For sample

```
switch<tab> creates

switch {
  case "somecase":
    someaction()
}
```
When inside this switch case doing the following creates another case statement

```
switch {
  case "somecase":
    someaction()
  case<tab>
}
```

There's still some snippets that need to be defined but the majority of the Go spec is accounted for.

One nasty issue is that SuperTab begins to evaluate gocode completion when inside brackets. So unfortunately this was not possible for me to do:

```
al\<tab\>

[]Bool{t\<tab\>, t\<tab\>}
```

We would expect t to expand to "true" however supertab kicks in with omni complete. Maybe this can be fixed by changing my trigger key for Utlisnips - not sure.

### DIFF
--- a/gosnippets/UltiSnips/go.snippets
+++ b/gosnippets/UltiSnips/go.snippets
@@ -189,47 +189,220 @@ snippet ta "type assertion"
 ${1:x}, ok := ${2:y}.(${3:type})
 endsnippet
 
-# anonymous function
-snippet anon "fn := func() { ... }"
-${1:fn} := func() {
-	${2:${VISUAL}}
+# Statements
+## panic
+snippet pn "panic()"
+panic("${0:msg}")
+endsnippet
+
+## Send Statements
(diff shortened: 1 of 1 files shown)


fatih:
Thanks @ldelossa for all this work. However I have to said this is unfortunately too long to be reviewed and merged. I'm all improving the current situation, but I prefer small, useful PR's so we can track them if something goes wrong. PR's like this creates a burden to us, the reviewers. I suggest you open small PR's with good description so we know what's going on and also can review them whenever we have time.

fatih:
Closing this as it's not managable to review (see my previous comment). Feel free to open small, readable and reviewable PR's if anyone want to get these into vim-go.

PR #1779, opened 2018-04-08
Add "ac" ("a comment") text object

### LINKED ISSUE #1508: Better reformatting for comments
### Behavior

I use `vipgq` to reformat a "paragraph" of text - typically a multiline comment. This works fine if there is at least 1 blank line above and below the paragraph. If there is go code immediately adjacent, the resulting formatted text is messed up.

### Steps to reproduce:

1. Write something like this:
```go
func main() {
        // Line 1.
        // Line 2.
        // Line 3.
        x := 0
}
```
2. Put the cursor on one of the comment lines
3. Type `vipgq` to select the paragraph and reformat it

Result:
```go
func main() {
        // Line 1.  Line 2.  Line 3.
        x := 0 }

```

### Configuration

Add here your current configuration and additional information that might be
useful, such as:

* `vimrc` you used to reproduce
```
set nocompatible
Plug 'fatih/vim-go', {'tag': '*'}
call plug#end()
```
* vim version: 8.0.1100_1
* vim-go version: v1.13
* go version: 1.9

Carpetsmoker:
To select a comment block. Fixes #1508

Requires fatih/motion#4

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1088,6 +1088,9 @@ if            "inside a function", select contents of a function,
                                                *go-v_ac* *go-ac*
 ac            "a comment", select contents of the current comment block.
 
+                                               *go-v_ic* *go-ic*
+ic            "inner comment", select contents of the current comment block,
+              excluding the start and end comment markers.
 
 vim-go also defines the following text motion objects:
 
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -1085,6 +1085,9 @@ if            "inside a function", select contents of a function,
               excluding the function definition and the closing bracket. This
               text-object also supports literal functions
 
+                                               *go-v_ac* *go-ac*
+ac            "a comment", select contents of the current comment block.
+
 
 vim-go also defines the following text motion objects:
 
--- a/autoload/go/textobj.vim
+++ b/autoload/go/textobj.vim
@@ -5,7 +5,7 @@
 " < >
 " t for tag
 
-function! go#textobj#Comment() abort
+function! go#textobj#Comment(mode) abort
   let l:fname = expand('%:p')
 
   try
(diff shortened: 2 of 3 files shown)


doc/vim-go.txt
```diff
@@ -1085,6 +1085,9 @@ if            "inside a function", select contents of a function,
               excluding the function definition and the closing bracket. This
               text-object also supports literal functions
 
+                                               *go-v_ac* *go-ac*
+ac            "a comment", select contents of the current comment block.
+
```
fatih: Is there way we could support `ic` as well? This could be handy if we want to remove the comment with leaving it intact. For example `// this is a comment` could be deleted via `dic`, so it would end up with `//`

fatih:
Just tried it out and seems like `ic` is not working then it's intended (groups of const, vars, imports, etc..). For example:

```
// Options is used to setup Kubernetes
type Options struct {
	// Client is the underlying client manager uses for core resources
	Client kubernetes.Interface
}
```

`ic` works for the first comment above, not for the second one. It deletes the whole line when you type `dic`

fatih:
The following fixes it:

```
$ vim-go git:(Carpetsmoker-ac) ✗ git diff
diff --git a/autoload/go/textobj.vim b/autoload/go/textobj.vim
index 2f60046..164b769 100644
--- a/autoload/go/textobj.vim
+++ b/autoload/go/textobj.vim
@@ -44,7 +44,9 @@ function! go#textobj#Comment(mode) abort
   " Adjust cursor to exclude start comment markers. Try to be a little bit
   " clever when using multi-line '/*' markers.
   if a:mode is# 'i'
-    let l:line = getline('.')
+    " trim whitespace so matching below works correctly
+    let l:line = substitute(getline('.'), '^\s*\(.\{-}\)\s*$', '\1', '')
+
     " //text
     if l:line[:2] is# '// '
       call cursor(l:info.startLine, l:info.startCol+3)
```

I'll merge this PR and apply the fix above.

PR #1787, opened 2018-04-08
Remove old goimports -srcdir workaround and fix :GoFmtAutoSaveToggle

Carpetsmoker:
This check was added two years ago in #735. I think it's safe to remove
now? It doesn't really do any harm, but does simplify things a bit.

Also happened to notice that `:GoFmtAutoSaveToggle` was broken:

  Error detected while processing function go#fmt#ToggleFmtAutoSave:
  line    7:
  E118: Too many arguments for function: go#config#FmtAutosave

So fix that as well.

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -157,27 +157,8 @@ function! s:fmt_cmd(bin_name, source, target)
     let opts = has_key(opts, a:bin_name) ? opts[a:bin_name] : ""
   endif
   call extend(cmd, split(opts, " "))
-
-  if a:bin_name == "goimports"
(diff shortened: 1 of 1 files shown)


bhcleek:
That flag isn't safe to remove; it's essential to how formatting with goimports works.

The `:GoFmtAutoSaveToggle` looks good to me, though.

Carpetsmoker:
Looks like the `call extend(cmd, ["-srcdir", a:target])` line got lost somehow >_< I intended to only remove the check for `-srcdir` support.

bhcleek:
lgtm

PR #1789, opened 2018-04-08
Allow re-running the last matched function with :GoTestFunc last

### LINKED ISSUE #1780: Test - Re-run the last test
Hi!

I use :GoTestFunc a lot, but it would be wonderful if there is a command to re-run the last test executed like :GoTestLastFunc.
It would be a great feature for TDD methodology, specially when I am doing refactor.

Thanks!

Carpetsmoker:
I'm not sure if `last` is the best keyword, but I think this is a nice
feature to add.

Also reword the documentation for `:GoTest` a bit.

~As a related change, this drops `-o` from `:GoTestCompile`. The output
file was in the Vim instances tmpdir, which is in the form of
`/tmp/$random/$n.vim-go.test`. There is no easy way to figure out the
value of either `$random` nor `$n`, so actually getting to the binary is
hard (you can use `g:go_debug = ['shell-command']` that was added just a
few days ago).~

~So, I don't think anyone is using that output file, and it's just taking
up disk space in `/tmp` (since Go binaries tend to be large it can add
up).~

Fixes #1780

### DIFF
--- a/autoload/go/test.vim
+++ b/autoload/go/test.vim
@@ -6,7 +6,9 @@ function! go#test#Test(bang, compile, ...) abort
 
   " don't run the test, only compile it. Useful to capture and fix errors.
   if a:compile
-    call extend(args, ['-c'])
+    " TODO: we should clean up this temp file.
+    let l:testfile = tempname() . '.vim-go.test'
+    call extend(args, ['-c', '-o', l:testfile])
   endif
 
   if a:0
--- a/autoload/go/test.vim
+++ b/autoload/go/test.vim
@@ -6,8 +6,7 @@ function! go#test#Test(bang, compile, ...) abort
 
   " don't run the test, only compile it. Useful to capture and fix errors.
   if a:compile
-    let testfile = tempname() . ".vim-go.test"
-    call extend(args, ["-c", "-o", testfile])
+    call extend(args, ['-c'])
   endif
 
   if a:0
@@ -22,8 +21,7 @@ function! go#test#Test(bang, compile, ...) abort
     call extend(args, goargs, 1)
   else
     " only add this if no custom flags are passed
-    let timeout = go#config#TestTimeout()
-    call add(args, printf("-timeout=%s", timeout))
+    call add(args, printf("-timeout=%s", go#config#TestTimeout()))
   endif
 
   if go#config#EchoCommandInfo()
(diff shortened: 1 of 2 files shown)


bhcleek:
The `-o` is important to keep; it makes sure that the user's source directories don't get cluttered with test binaries. The purpose of `:GoTestCompile` is not to actually get a binary to run; it's just to make sure the tests will actually build. Operating systems have all kind of ways to manage temp directories. Additionally, vim's temporary directory gets cleaned out when Vim exits.

As for using `last` as an argument to `:GoTestFunc`; using it would prevent a user from running `:GoTestFunc last` to run the tests for a package named `last`. Probably not the end of the world, but there's no need to introduce that edge case. `#` is another option that would at least be congruent with Vim buffers. I'm not keen on it, but it's something to consider.

Personally, I think #1780 should be closed as a won't fix. It's easier to just type `:GoTestF` and hit the up arrow key once than it is to type `:GoTestFunc last`. :shrug:

Carpetsmoker:
> The -o is important to keep; it makes sure that the user's source directories don't get cluttered with test binaries.

Oh, right. Looks like vim-go doesn't issue any error/warnings on packages without any test files, so I didn't notice that.

I'll change it later to clean up the tmp file, instead.

>  Operating systems have all kind of ways to manage temp directories

Usually only in between reboots though. Randomly removing files from `/tmp` can cause a lot of problems, as can having it fill up:

    [~]% dd if=/dev/zero of=/tmp/xxx
    dd: writing to '/tmp/xxx': No space left on device

    [~]% echo hello > /tmp/new
    echo: write error: no space left on device

> Additionally, vim's temporary directory gets cleaned out when Vim exits.

Yeah, but this can be a while. Running `:GoTestFunc` 100 times will mean 700M of tmp files.

> Personally, I think #1780 should be closed as a won't fix. It's easier to just type :GoTestF and hit the up arrow key once than it is to type :GoTestFunc last. :shrug:

The use case is there you type `:GoTestFunc`, then go to `implementation.go` to fix something, and want to run the same test again without going back to the test file. Could use `GoTest -run TestMyStuff` as well, but this is a bit easier, I think.

bhcleek:
> Running :GoTestFunc 100 times will mean >700M of tmp files
Not necessarily. Docs say 
> The name is different for at least 26 consecutive calls.

Depending on the platform in that scenario (100 runs), I'd expect there would be at least 42M of temp files, but I'd need to check source code to see what actually happens on various platforms. 🤷‍♂️ 

Given that the arguments aren't _just_ flags, I'm not sure the `[expand]`s in the docs should be changed to `[flags]` . 🤔 , but otherwise LGTM.

fatih:
Don't forget to drop a line to the changelog under backwards incompatibilities due the expand/flags change.

> Given that the arguments aren't just flags, I'm not sure the [expand]s in the docs should be changed to [flags] 

We actually use also `[options]` in most of the commands in addition to `[expand]`. I think we can go over this in a separate PR that fixes these (if any).

fatih:
ping @Carpetsmoker

mchacon-ml:
Sorry to bother you guys, but if the PR was already done, Is there an special reason why you did not merge it?

PR #1790, opened 2018-04-09
Fix gohtmltmpl syntax to show everywhere

powerman:
This change enable highlighting of `{{…}}` everywhere within html, for example in tags:
```html
<input type=checkbox {{ if .Check }}checked{{ end }} value="{{ .Value }}">
```


Carpetsmoker:
Seems to work well. Thanks :+1:

PR #1794, opened 2018-04-11
Fix go syntax: no goReceiverType in goVar

powerman:
Identifier after a pointer inside `var()` block incorrectly highlighted as goReceiverType:
```vim
let g:go_highlight_types = 1
```
```go
var (
	one = make(map[string]*wrongHERE.something)
	two *wrongHERE.something
)
```

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -148,14 +148,14 @@ endif
 " var, const
 if go#config#FoldEnable('varconst')
   syn region    goVar               start='var ('   end='^\s*)$' transparent fold
-                        \ contains=ALLBUT,goParen,goBlock,goFunction,goTypeName,goReceiverType,goReceiverVar,goArgumentName,goArgumentType,goSimpleArguments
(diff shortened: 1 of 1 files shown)


fatih:
Hi @powerman 

Can you please send a before/after picture. Even though I've added your change, nothing has changed for me.

powerman:
Sorry, it's activated by `let g:go_highlight_functions = 1`, not `g:go_highlight_types`.
vim-go-1
And then I press Enter (notice prepared vim command line above) and got this:
vim-go-2

Carpetsmoker:
Note that `let g:go_highlight_functions = 1` is documented as:

> Highlight function and method declarations.

So I'm not sure why you expect that to highlight `url` in the above example?

All these highlight settings and their interactions are so messy :-(

powerman:
> So I'm not sure why you expect that to highlight url in the above example?

I'm expecting it to *not* highlight anything inside `var(…)`, but it does highlight it. PR fixed that.

PR #1799, opened 2018-04-16
Improve performance of package comment regex

### LINKED ISSUE #1747: Package comment folding can cause serious slowdowns
Edited https://github.com/hashicorp/vault/blob/master/vault/logical_system.go#L69

Normal keyboard motion responsiveness.

### What happened instead?
Keyboard up/dn navigation became choppy around line 75.  I played around and noticed that the presence of numerous `/*` in the strings was related to the slowdown.  Made an extreme case by repeating line 85 many times and the file is barely usable in this section: https://gist.github.com/kalafut/efe4c028672df01e2dd93e8c4f358d72#file-logical_system-go-L85

Eventually traced the slowdown to package comment folding. When I turn package comment folding off via `let g:go_fold_enable = ['block', 'import', 'varconst' ]` in .vimrc, performance is fine.  So something about that pattern of text is really killing the package fold regex/logic.

VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Mar 24 2018 08:30:22)
macOS version

* Go version (`go version`):
go1.10 darwin/amd64

* Go environment (`go env`):
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/kalafut/Library/Caches/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/kalafut/go"
GORACE=""
GOROOT="/usr/local/Cellar/go/1.10/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.10/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/y1/vdw7z6tn3wj1drd98jmwwsdc0000gn/T/go-build427136356=/tmp/go-build -gno-record-gcc-switches -fno-common"

* vim-go version:
1.17

Simply enabling vim-go (no other plugins, no customization) will show the slow performance. Adding `let g:go_fold_enable = ['block', 'import', 'varconst' ]` will improve the performance.

kalafut:
By anchoring to the start of the document (since that's the only place
a package comment will appear), the number of false matches is greatly
reduced in some files.

Fixes #1747

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -351,11 +351,11 @@ if go#config#HighlightBuildConstraints() || go#config#FoldEnable('package_commen
   " matched as comments to avoid looking like working build constraints.
   " The he, me, and re options let the "package" itself be highlighted by
   " the usual rules.
-  exe 'syn region  goPackageComment    start=/\v(\/\/.*\n)+\s*package/'
+  exe 'syn region  goPackageComment    start=/\v%^(\s*\n)*(\/\/.*\n)+\s*package/'
         \ . ' end=/\v\n\s*package/he=e-7,me=e-7,re=e-7'
         \ . ' contains=@goCommentGroup,@Spell'
         \ . (go#config#FoldEnable('package_comment') ? ' fold' : '')
-  exe 'syn region  goPackageComment    start=/\v\/\*.*\n(.*\n)*\s*\*\/\npackage/'
+  exe 'syn region  goPackageComment    start=/\v%^(\s*\n)*\/\*.*\n(.*\n)*\s*\*\/\npackage/'
         \ . ' end=/\v\*\/\n\s*package/he=e-7,me=e-7,re=e-7'
         \ . ' contains=@goCommentGroup,@Spell'
         \ . (go#config#FoldEnable('package_comment') ? ' fold' : '')
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -351,11 +351,11 @@ if go#config#HighlightBuildConstraints() || go#config#FoldEnable('package_commen
   " matched as comments to avoid looking like working build constraints.
   " The he, me, and re options let the "package" itself be highlighted by
   " the usual rules.
-  exe 'syn region  goPackageComment    start=/\v%^(\s*\n)*(\/\/.*\n)+\s*package/'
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
A package comment is not always at the start of the document; it can be preceded with one or more other comment blocks.

For example this file will no longer work correct:

```
// +build foo

/*
Copyright ...
*/

// Package main should be folded
// fold this.
package main

func main() {
}
```

I'm not sure if there is a way to make this significantly faster; maybe using `\%^.*` is faster, or maybe not.


<details>
<summary>Screenshots</summary>

Master:
2018-04-16-052451_623x379_scrot

This PR:
2018-04-16-052432_513x368_scrot
</details>

kalafut:
OK, I see what you mean. Let me see if I can adjust this.

kalafut:
I played around with a number of regex that mostly handled these cases, but they became uncomfortably complex. Taking a step back, the real performance problems stem from the /* ... */ matching, since they can be started with any `/*` and end up searching the whole rest of the file.  So a completely different approach to this PR is simply to replace:

`start=/\v\/\*.*\n(.*\n)*\s*\*\/\npackage/'` with
`start=/\v^\s*\/\*.*\n(.*\n)*\s*\*\/\npackage/`

This start of line anchor helps a lot because the it prevents quoted `/*` from starting new searches, and is (I think?) still valid for package comments. 

Note that if there are block (/* .. */) comments (like a license) and then a block commented package comment, the license+package get folded together, like you were just mentioning. But that behavior is present even in the released version, and line comment style is way more comment anyway.

If you like this approach I'll update that PR with it.

Carpetsmoker:
Seems like a good solution :+1: Thanks!

PR #1814, opened 2018-05-04
gocode: Change repo from nsf/gocode to mdempsky/gocode

rhysd:
Hi,

I noticed that recently nsf/gocode is no longer maintained (please see [notice at the top of README](https://github.com/nsf/gocode#an-autocompletion-daemon-for-the-go-programming-language)). And now [mdempsky/gocode](https://github.com/mdempsky/gocode) is an official active repository of gocode.

So I think vim-go should use the new repository for now. This PR only changes the repository URL in vim-go repository.

Please note that it's not straightforward to change the URL because the new repo only supports Go 1.10+. What do you think, @fatih?

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -34,7 +34,7 @@ let s:packages = {
       \ 'dlv':           ['github.com/derekparker/delve/cmd/dlv'],
       \ 'errcheck':      ['github.com/kisielk/errcheck'],
       \ 'fillstruct':    ['github.com/davidrjenni/reftools/cmd/fillstruct'],
-      \ 'gocode':        ['github.com/nsf/gocode', {'windows': '-ldflags -H=windowsgui'}],
+      \ 'gocode':        ['github.com/mdempsky/gocode', {'windows': '-ldflags -H=windowsgui'}],
       \ 'godef':         ['github.com/rogpeppe/godef'],
       \ 'gogetdoc':      ['github.com/zmb3/gogetdoc'],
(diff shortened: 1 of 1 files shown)


fatih:
@rhysd this is definitely something we should take care of. If the maintainer also advices to use something else, I'm all for it. The only concern is that he somehow implies it's "temporary". I don't think it shouldn't be a problem at all of course, just something to be reminded. 

Let me look into this, I'm curious why there is even a fork and what the reason for that is. I'll merge this afterwards. Thanks @rhysd

fatih:
Ok this is the GH issue that goes more over why the fork exists:

https://github.com/nsf/gocode/issues/305

Also the note from the original author is also a sign that makes it for us safe to switch to the new fork:

> And if you eventually want to step over and maintain gocode for the rest of your life, I totally don't mind. I don't have time for it anymore.

> I can also put a message on README.md so that users try that other version of gocode, but the problem is - I don't think that many users actually read README.md. You see gocode is not an end user product, there is a ton of editors that rely on gocode being as it is and it puts a lot of pressure on me when it comes to global changes like that. I can't just throw in a new version of gocode silently. Then if it fails what do I do? Revert everything back? I would rather maintain it as it is.

I think it's safe to switch. The worst is we can switch back. I think it's better for us to choose stable and maintained projects over stalled one for forwards compatibility and maintenance of vim-go.

fatih:
Let's try this. Thanks @rhysd

rhysd:
Thank you. Please feel free to revert this branch if it causes some troubles.

PR #1826, opened 2018-05-15
guru.vim: defer ImportPath check

yousong:
In the k8s.io/kubernetes repo, vendor/k8s.io/{api,apiservers} etc. are
symbolic links to corresponding dir staging/src/k8s.io/{api,apiservers}.

    vendor/k8s.io/apiserver -> ../../staging/src/k8s.io/apiserver

ImportPath detection for files under these staging directories will fail
because '%:p:h' does not match any items of 'go list'

For large projects like k8s, users tend to set the g:go_guru_scope
variable.  The change is mainly a workaround trying to improve
experience so that those not worked before will work now and it will not
affect existing usage.

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -16,12 +16,6 @@ function! s:guru_cmd(args) range abort
   let selected = a:args.selected
 
   let result = {}
-  let pkg = go#package#ImportPath()
-
-  " this is important, check it!
-  if pkg == -1 && needs_scope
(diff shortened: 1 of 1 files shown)


yousong:
Hi, the change should be safe to merge.  It just moves the integrity check to where it was actually needed.  The check failure on test coverage was also not newly caused by the diff of the change.

bhcleek:
Thank you.

Issue #1839, opened 2018-05-25
Doesn't respect GoDebugCurrent/GoDebugBreakpoint highlight group customizations


mkomitee:
### What did you do? (required. The issue will be **closed** when not provided.)

I tweaked my colorscheme to specify a different highlight group configuration for GoDebugCurrent.

### What did you expect to happen?

When I started debugging, I expected the current line to be highlighted using my customizations.

### What happened instead?

When I started vim, I checked, and my customizations were applied, but when I started my debug session, it was overridden with the defaults from autoload/go/debug.vim.

### Configuration (**MUST** fill this out):

* vim-go version: 7ac1e62fa30d84946e93a3820bb6cf6a431186ed

* `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):

After installing vim-go using the recommended method for vim8 packages, the following vimrc can be used to replicate the problem:

```
runtime defaults.vim

augroup custom
  autocmd!
  autocmd FileType go hi! GoDebugCurrent term=underline ctermbg=237 guibg=#3c3836
augroup END
```

After opening a simple main.go file, the GoDebugCurrent highlight group from the vimrc is applied, but after actually starting a debug session, the defaults have been installed.

* Vim version (first three lines from `:version`):

```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Apr  3 2018 10:22:54)
Included patches: 1-1655
```

I compiled it myself, so the third line shows the host on which it was built.

* Go version (`go version`):
```
go version go1.10.2 linux/amd64
```

* Go environment (`go env`):
```
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/mkomitee/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/mkomitee/go"
GORACE=""
GOROOT="/home/mkomitee/go-1.10.2"
GOTMPDIR=""
GOTOOLDIR="/home/mkomitee/go-1.10.2/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
[...]

PR #1840, opened 2018-05-26
Respect user debug highlight group customization

mkomitee:
As mentioned in #1839, If the user sets GoDebugCurrent or
GoDebugBreakpoint, we should respect their customization and not
override them.

The problem is that we're defining it in autoload, so it gets evaluated
after user customizations have been applied. I suspect the right thing
to do is to move the definition of the groups & signs to the syntax
file, but I'm going to assume things were done the way they were for a
good reason, and not propose that change (I'm no vim-expert).

It's not obvious how we can check for this, so I've chosen to execute
"hi GoDebugBreakpoint" and look for a string that appears to be present
when it was not already set.

### DIFF
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -894,8 +894,13 @@ sign define godebugbreakpoint text=> texthl=GoDebugBreakpoint
 sign define godebugcurline text== linehl=GoDebugCurrent texthl=GoDebugCurrent
 
 fun! s:hi()
-  hi GoDebugBreakpoint term=standout ctermbg=117 ctermfg=0 guibg=#BAD4F5  guifg=Black
-  hi GoDebugCurrent    term=reverse  ctermbg=12  ctermfg=7 guibg=DarkBlue guifg=White
+
+  if execute("hi GoDebugBreakpoint", "silent!") =~ "xxx cleared"
+    hi GoDebugBreakpoint term=standout ctermbg=117 ctermfg=0 guibg=#BAD4F5  guifg=Black
+  endif
+  if execute("hi GoDebugCurrent", "silent!") =~ "xxx cleared"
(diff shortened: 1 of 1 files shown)


bhcleek:
Thanks for the contribution @mkomitee.

I think the problem you're encountering and trying to solve here is due to these highlight groups and the `s:hi()` function not being in the `syntax` directory. Can you try moving the original code there instead of trying to workaround the highlight groups being in the wrong place?

Carpetsmoker:
I think we can just add this to `syntax/go.vim`. It does little harm in being there, and would make things simpler. The only downside is that the highlight group will be defined even if you never use the debug feature (which is why it's here, but that's not a big deal, I think).

autoload/go/debug.vim
```diff
@@ -894,8 +894,13 @@ sign define godebugbreakpoint text=> texthl=GoDebugBreakpoint
 sign define godebugcurline text== linehl=GoDebugCurrent texthl=GoDebugCurrent
 
 fun! s:hi()
-  hi GoDebugBreakpoint term=standout ctermbg=117 ctermfg=0 guibg=#BAD4F5  guifg=Black
-  hi GoDebugCurrent    term=reverse  ctermbg=12  ctermfg=7 guibg=DarkBlue guifg=White
+
+  if execute("hi GoDebugBreakpoint", "silent!") =~ "xxx cleared"
```
Carpetsmoker: `execute()` was added in Vim 7.4.2008, and we support Vim 7.4.1689 and newer, so that can't be used. I did start an issue just now to discus relaxing that requirement though: #1847

bhcleek: FWIW, if the original code is just moved to `syntax`, then there won't be any need to guard the highlighting, because of the order in which highlight groups are applied.

mkomitee:
This can be closed, as I submit #1850 which addresses this in the suggested manner. Since there's some conversation here, I'll leave it to the maintainers to actually close it.

PR #1845, opened 2018-05-30
Add iferr

mattn:
iferr


mattn:
I will update doc in later.

fatih:
This is cool! @mattn looking forward to the doc. If you can't make it let me know, I'll add it myself after merging.

fatih:
@mattn would it possible to create also a `:GoIfErr` command (with the doce)? Not all vim-go users are advanced users and they heavily use commands. They know how to map commands to mappings.

fatih:
Thanks @koron and @mattn. This is just great! :)

PR #1846, opened 2018-05-30
fix guru output handling in Neovim

### LINKED ISSUE #1836: GoReferrers sometimes show raw text (not a link) with nvim
type :GoReferrers

List with links to the code where current variable used.

### What happened instead?
Some rows are not links, they are just text. See third row.
screenshot from 2018-05-22 21-38-09

* vim-go version:
1.17

```
call plug#begin('~/.config/nvim/plugged')
Plug 'fatih/vim-go', { 'do': ':GoUpdateBinaries' }
call plug#end()
```

NVIM v0.2.2

* Go version (`go version`):
go1.10.1 linux/amd64

* Go environment (`go env`):
```
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/evgen/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/evgen/go"
GORACE=""
GOROOT="/usr/lib/go-1.10"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go-1.10/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build845585396=/tmp/go-build -gno-record-gcc-switches"
```

bhcleek:
Output from a Neovim job may come in batches and splits on newlines.
When the output ends with a newline, the last element in the data passed
to the stdout and stderr callbacks will be the empty string. When it's
not the empty string, then last element is an incomplete line, and the
next call will contain the rest of the line in the first element of its
data.

Join shell arguments correctly when Neovim calls guru.

Fixes #1836

### DIFF
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -127,7 +127,7 @@ function! s:job_start(cmd, start_options) abort
   " use a shell for input redirection if needed
   let cmd = a:cmd
   if has_key(a:start_options, 'in_io') && a:start_options.in_io ==# 'file' && !empty(a:start_options.in_name)
-    let cmd = ['/bin/sh', '-c', join(a:cmd, ' ') . ' <' . a:start_options.in_name]
+    let cmd = ['/bin/sh', '-c', go#util#Shelljoin(a:cmd) . ' <' . a:start_options.in_name]
   endif
 
   return jobstart(cmd, opts)
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -113,12 +113,43 @@ function! s:job_start(cmd, start_options) abort
   endif
 
   let opts = {'stdout_buffered': v:true, 'stderr_buffered': v:true}
+
+  let stdout_buf = ""
   function opts.on_stdout(job_id, data, event) closure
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
Isn't this a problem in all locations where we read from stdout/stderr in the async API?

bhcleek:
Your instincts are right; this is likely a problem in other places, too. It's not a problem everywhere, though, because some places use JSON instead of being line-oriented output. I'm planning to go through and unify the vim8 and Neovim async job handling so we can have a consistent approach. When I do that, I'll be sure to fix any code locations that also have this problem.

PR #1850, opened 2018-06-04
Respect user debug highlight group customization

mkomitee:
Before, this was defined as part of the debug autoload, which would only
be invoked when a user started debugging for the first time. This caused
it to override any customizations a user may have had in their own
configuration.

By migrating the highlight groups to the syntax file, users can use
standard methods to override these colors if they don't work with their
chosen colorscheme, e.g. with a line like:

```vim
autocmd custom FileType go hi! GoDebugCurrent term=underline ctermbg=237 guibg=#3c3836
```

in their `vimrc`.

This supercedes fatih/vim-go#1840 and fixes fatih/vim-go#1839.


bhcleek:

lgtm

Issue #1817, opened 2018-05-06
gocode autobuild not being set to true by default


tombell:
### What did you do? (required. The issue will be **closed** when not provided.)

Typed package name `geo.`, then  <kbd>C-x</kbd><kbd>C-o</kbd> for completion of functions in the package `geo.` a sub package in the project.

### What did you expect to happen?

Completion of exported functions/vars for the package.

### What happened instead?

No completions found.

### Configuration (**MUST** fill this out):

* vim-go version:

`master (765aff6cfdebd27657bd796b0736cb05d992a0e6)`

* `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):

* Vim version (first three lines from `:version`):

```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Apr 17 2018 10:25:52)
macOS version
Included patches: 1-1633
```

* Go version (`go version`):

```
go version go1.10.2 darwin/amd64
```

* Go environment (`go env`):

```
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/tombell/Library/Caches/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/tombell/Code/go"
GORACE=""
GOROOT="/usr/local/Cellar/go/1.10.2/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.10.2/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/cv/1wjktrsd7gq5ttrjpcfkdvjh0000gn/T/go-build639365319=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

[...]

PR #1851, opened 2018-06-05
gocode: switch to https://github.com/mdempsky/gocode

fatih:
nsf/gocode is not working well with Go 1.10. I just encountered it myself again. The author of original gocode also states and recommends the mdempsky/gocode fork over nsf/gocode: https://github.com/nsf/gocode#an-autocompletion-daemon-for-the-go-programming-language

We actually [moved once](https://github.com/fatih/vim-go/pull/1814), but because we didn't removed the obsolote settings, we weren't prepared for issues that were [reported](https://github.com/fatih/vim-go/issues/1817), and hence [reverted it back](https://github.com/fatih/vim-go/pull/1823)

This PR also removes all settings. The reason is that the new gocode doesn't support settings anymore. The author decided not the support it.

* For `autobuild`, this was removed because the author believe it's not
worth tackling it inside `gocode`: https://github.com/mdempsky/gocode/issues/13
* For `propose builtins`, this is actually still supported but disabled in
the source code, we need to add it back again in the future:
https://github.com/mdempsky/gocode/commit/787cfae70965d528523902b4c12e85692e17336b
* For `unimported packages`: this was just removed and I couldn't find
upcoming plans for it

Going forward this package seems to be more stable and is a better bet. Most important, as I stated nfs/gocode is not 1.10 compatible and the author has no intentions to fix it.

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -43,31 +43,7 @@ function! s:sync_gocode(cmd, args, input) abort
   return l:result
 endfunction
 
-let s:optionsEnabled = 0
-function! s:gocodeEnableOptions() abort
-  if s:optionsEnabled
-    return
-  endif
-
-  let l:bin_path = go#path#CheckBinPath("gocode")
-  if empty(l:bin_path)
-    return
-  endif
-
-  let s:optionsEnabled = 1
-
-  call go#util#Exec(['gocode', 'set', 'propose-builtins', s:toBool(go#config#GocodeProposeBuiltins())])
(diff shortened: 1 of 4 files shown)


bhcleek:
lgtm

bhcleek:
I tested this poorly before merging and subsequently found problems. The merge has been reverted, and a new PR is open with these same changes at #1853.

fatih:
Could you write what these problems are? Would be helpful to fix those. I'm using it since I've switched it and it was working well.

bhcleek:
I've started documenting on #1853. Let's move to there.

PR #1854, opened 2018-06-08
Pretty print maps in JSON and YAML

fit-fg:
This marshals and prints a `map` in indented JSON or YAML format. 
Implements UltiSnips only.


bhcleek:
Thank you for the contribution. While this snippet may be useful for you personally, I am doubtful that it's generically useful enough for all users of vim-go to justify its inclusion in vim-go's set of snippets.

PR #1858, opened 2018-06-11
do not indent second line of raw string

### LINKED ISSUE #1800: Unexpected auto-indent within raw string literal

Create a go file with the following contents:

```go
package main

import "fmt"

func main() {
	const name = `
`
	fmt.Printf("this is my name: %v\n", name)
}
```

In normal mode, have the cursor over the `const` keyword. Now press `o` to begin a new line below the cursor and enter insert mode.

The cursor should be place at the 0th column because we are within a raw string literal (cursor position indicated by `^` below; the line that includes the `^` does not exist in reality):

```go
package main

import "fmt"

func main() {
	const name = `

^
`
	fmt.Printf("this is my name: %v\n", name)
}
```

### What happened instead?

The new line is automatically indented and the cursor is placed after the indent:

```go
package main

import "fmt"

func main() {
	const name = `
	
	^
`
	fmt.Printf("this is my name: %v\n", name)
}

```

* vim-go version: 7491209072ed4aa746e6fe7894f976ecd251801e

```
Plug 'fatih/vim-go'
call plug#end()
```

```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Apr 16 2018 07:40:43)
Included patches: 1-1722
Compiled by myitcv@myitcv-virtual-machine
```

* Go version (`go version`):

```
go version go1.10.1 linux/amd64
```

* Go environment (`go env`):

```
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/myitcv/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/myitcv/gostuff"
GORACE=""
GOROOT="/home/myitcv/gos"
GOTMPDIR=""
GOTOOLDIR="/home/myitcv/gos/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build746364427=/tmp/go-build -gno-record-gcc-switches"
```

bhcleek:
Do not match the indentation of code that assigns a multi-line raw string within the raw string (i.e. the second line of a raw string should not match the indentation of the line of code that assigns the raw string to a variable).

Fixes #1800

### DIFF
--- a/autoload/go/indent_test.vim
+++ b/autoload/go/indent_test.vim
@@ -1,0 +1,22 @@
+func! Test_indent_raw_string() abort
+  try
+    let l:dir= gotest#write_file('indent/indent.go', [
+          \ 'package main',
+          \ '',
+          \ 'import "fmt"',
+          \ '',
+          \ 'func main() {',
+          \	"\t\x1fconst msg = `",
+          \ '`',
+          \ '\tfmt.Println(msg)',
+          \ '}'])
+
+    silent execute "normal o" . "not indented\<Esc>"
+    let l:indent = indent(line('.'))
+    call assert_equal(0, l:indent)
+  finally
+    call delete(l:dir, 'rf')
+  endtry
+endfunc
+
+" vim: sw=2 ts=2 et
--- a/indent/go.vim
+++ b/indent/go.vim
@@ -38,6 +38,10 @@ function! GoIndent(lnum)
 
   let ind = previ
 
+  if prevl =~ ' = `[^`]*$'
+    " previous line started a multi-line raw string
+    return 0
+  endif
   if prevl =~ '[({]\s*$'
     " previous line opened a block
     let ind += shiftwidth()
--- a/indent/go.vim
+++ b/indent/go.vim
@@ -24,17 +24,6 @@ if exists("*GoIndent")
   finish
 endif
 
-" use shiftwidth function only if it's available
-if exists('*shiftwidth')
-  func s:sw()
-    return shiftwidth()
-  endfunc
-else
-  func s:sw()
-    return &sw
-  endfunc
-endif
-
 function! GoIndent(lnum)
   let prevlnum = prevnonblank(a:lnum-1)
   if prevlnum == 0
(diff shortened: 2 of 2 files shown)


Carpetsmoker:
I don't know. To me, it seems both behaviours can make sense, depending on what your goal is.

Also, it looks like it doesn't work for me with this example?

```
package main

import "fmt"

func main() {
    fmt.Println(`
```

But I'm not sure if it's worth fixing it. It's debatable if it's actually better, and getting it correct 100% of the time is hard. Right now at least the behaviour is consistent and predictable.

bhcleek:
I can't think of a time that the indent used by code is the indent that should be applied inside a string. Even if there is such a case, the two indent levels are only coincidentally the same.

Carpetsmoker:
I use it in tests, for example:

	tests := []struct{
		input, want string
	}{
		{
			input: `
				some
				multi-line
				string
			`,
			want: `
				another
				multi-line
				string
			`,
		},
	}

Another case is something like HTML templates. I actually wrote this:

	var (
		block = template.Must(template.New("block").Parse(`
			<div>
			...
		`))

		page = template.Must(template.New("page").Parse(`
			<div>
			...
		`))
	)

In both cases, indenting makes more sense to me. There are probably other cases as well.

bhcleek:
Interestingly, though, even in those examples you didn't maintain the indents of the lines of code that started the raw strings; you indented them further. Keeping the indentation of a line of code for a _raw_ string seems wrong on its face to me, because strings are values, not instructions.

I checked the standard library. There are a total of 204 raw strings:
```
pcregrep -c   --include '.*.go' -r "= \`$" * | grep -v ':0' | sed -n -e 's/^.*:\(.*\)/\1/p' | paste -sd+ - | bc
```
Of those, only 25 raw strings 
```
pcregrep -c  -M --include '.*.go' -r "= \`\n\t" * | grep -v ':0' | sed -n -e 's/^.*:\(.*\)/\1/p' | paste -sd+ - | bc
```
have leading whitespace on the second line (the example above will report 27, but there are two false positives.

fatih:
This looks good. I also had the same problem from time to time. Raw strings shouldn't be indented anyway per se.

PR #1864, opened 2018-06-20
unify async job handling

bhcleek:
Unify how jobs are handled in `autoload/go/job.vim` and `autoload/go/jobcontrol.vim` so that there's a single abstraction for async jobs instead of separate abstractions for Vim and Neovim. Unifying the abstractions allows us to provide the same behavior around updating status, writing progress, and handling errors in Vim and Neovim. 

As part of this work, I was able to refactoring async `go test` jobs to use the abstraction in `autoload/go/job.vim` and also enable the tests in `autoload/go/test_test.vim` for Neovim.

Among other things, this work paves the way to:
1. Move even more jobs that currently explicitly use `job_start` or `jobstart` into the abstraction so that there are fewer one offs.
2. Move jobs that currently use `job_start` in Vim, but use a synchronous call in Neovim, to use the abstraction instead.
3. Extend the vim-go term support to include Vim as well as Neovinm.
4. Extend the vim-go term support to more async jobs than is currently supported.

Though this PR is relatively large, the commits that make up this change tell a coherent story using small and focused changes; reviewing this PR will be made easier by reviewing each commit individually.


autoload/go/job.vim
```diff
@@ -1,7 +1,18 @@
-" Spawn returns callbacks to be used with job_start. It is abstracted to be
+let s:cd = exists('*haslocaldir') && haslocaldir() ? 'lcd' : 'cd'
```
Carpetsmoker: As I understand it, this doesn't do what you want. The `haslocaldir()` checks if the current window has a local path set already with `:lcd`. So evaluating it once here isn't really that useful. You need to evaluate it when using.

We could add a wrapper function in `util.vim` though.

bhcleek: Good catch.

bhcleek:
PTAL

PR #1771, opened 2018-04-07
Fix shell escaping for :GoDoc, :GoDef

Carpetsmoker:
Fixes #1766; Fixes #1767

Using `-tags ''` is valid in pretty much all commands, but due to inconsistent ad-hoc application of shell escaping empty tags sometimes wouldn't get passed correctly.

The solution is to use `go#util#Exec()`, which handles this for us.

This will also fix issues with multiple build tags set (e.g. `let g:go_build_tags = 'aa bb'`).

PR #1866, opened 2018-06-20
Fix wrong binary path for go#util#Exec

Lekensteyn:
Commands such as :GoImports ended up calling "goimports" directly rather
than using the full path based on g:go_bin_path. Restore the previous
behavior of using the expanded binary path.

Fixes regression from #1771
(commit 32ae8640716530bd55062379177da51efb37dfd2,
"Fix infinite recursion when GOPATH is empty, part two").
___
Cc: @Carpetsmoker

I don't have `~/go/bin` in my `$PATH`, but I do have `let g:go_bin_path = expand('~/go/bin')` in my vimrc.
It would be nice to have a regression test for this, but `scripts/install-vim` always adds `$GOPATH/bin` to `PATH`, so I am not sure if it is feasible to do this?

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -167,13 +167,16 @@ function! go#util#Exec(cmd, ...) abort
 
   let l:bin = a:cmd[0]
 
+  " Lookup the full path, respecting settings such as 'go_bin_path'. On errors,
   " CheckBinPath will show a warning for us.
   let l:bin = go#path#CheckBinPath(l:bin)
   if empty(l:bin)
     return ['', 1]
   endif
 
-  return call('s:exec', [a:cmd] + a:000)
+  " Finally execute the command using the full, resolved path. Do not pass the
+  " unmodified command as the correct program might not exist in $PATH.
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
I did it like this because e.g. `guru args` looks nicer in logs and such than `/full/path/to/guru args`, but I guess there are some cases where that doesn't work :-(

> It would be nice to have a regression test for this, but scripts/install-vim always adds $GOPATH/bin to PATH, so I am not sure if it is feasible to do this?

You can add a comment: "Make sure to use the full path so it works in cases where people don't have GOBIN in GOPATH.", or something like that.

Lekensteyn:
> You can add a comment: "Make sure to use the full path so it works in cases where people don't have GOBIN in GOPATH.", or something like that.

Done, hopefully this will avoid future regressions. I think that `scripts/install-vim` and `scripts/run-vim` can also be modified to avoid setting `PATH` (as this is discovered at automatically), but I'll leave this at your discretion.

bhcleek:
Thank you for the contribution, @Lekensteyn 

lgtm

PR #1878, opened 2018-07-09
Fix GoDeclsDir for ctrlp on a directory other than '.'

sorribas:
If you run `GoDeclsDir ./some/package` with vim-go using CtrlP, it show the list of declarations in the package, but when you hit enter on one of them, it just opens a new buffer with the name of the file containing the selected symbol, but it doesn't correctly open the existing file. This fixes that issue.


bhcleek:
Thank you for the contribution @sorribas 

LGTM

PR #1883, opened 2018-07-16
Add GoDeclsDirRecur command.

sorribas:
This PR depends on fatih/motion#5.

I really like the `:GoDeclsDir` command on vim-go. I think it's really useful to find a function or type directly, rather than a file. However, I think the limitation that it only does one package at a time is not great if you want to quickly navigate a project divided into many packages.

This PR aims to address this by adding a recursive version of this command, `:GoDeclsDirRecur`. Before this can work, my PR on motion has to be merged, so that motion itself supports the recursive option, which it currently doesn't.


fatih:
Closing this as I think this is not the correct way to implement, I added more details here: https://github.com/fatih/motion/pull/5 Thanks for the work nevertheless.

PR #1885, opened 2018-07-17
travis: fix temporary vint issue

fatih:
Until https://github.com/Kuniwak/vint/issues/287 is fixed


fatih:
Let's revert this once upstream is fixed. It's annoying to see all those reds for no reason.

bhcleek:
lgtm

PR #1886, opened 2018-07-18
Fix guru execution to use g:go_bin_path

dhduvall:
PR #1866 was incomplete: `go#def#Jump()` calls `guru` asyncronously, and
not through `go#util#Exec()`, and so `:GoDef` fails if `guru` isn't
found in `$PATH`.


fatih:
Thanks @dhduvall for the fix

PR #1888, opened 2018-07-19
Add goreturns

hori-ryota:
I wanted to use goreturns instead of goimports, so I added it.

I think that this issue has already been resolved.
[goreturns configuration · Issue \#208 · fatih/vim\-go](https://github.com/fatih/vim-go/issues/208)

Although it is not a perfect implementation, I was able to confirm the operation for the time being. Can you positively think about it?

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -13,7 +13,7 @@
 "  it doesn't undo changes and break undo history.  If you are here reading
 "  this and have VimL experience, please look at the function for
 "  improvements, patches are welcome :)
-function! go#fmt#Format(withGoimport) abort
+function! go#fmt#Format(fmtBinFlag) abort
   if go#config#FmtExperimental()
     " Using winsaveview to save/restore cursor state has the problem of
     " closing folds on save:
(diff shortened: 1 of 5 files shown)


autoload/go/fmt.vim
```diff
@@ -13,7 +13,7 @@
 "  it doesn't undo changes and break undo history.  If you are here reading
 "  this and have VimL experience, please look at the function for
 "  improvements, patches are welcome :)
-function! go#fmt#Format(withGoimport) abort
+function! go#fmt#Format(fmtBinFlag) abort
```
hori-ryota: How about replace flag to bin_name direct?

fatih:
Hi @hori-ryota 

Do we really need a new command for this? I regret adding `:GoImports` already and believe adding another command that overwrites `:GoFmt` is not needed. Doesn't `goreturns` works if you just replace the binary? Or is there something else to do?

hori-ryota:
As for goreturns, it is enough to overwrite goimports because the usage of the -srcdir, -w option is the same. That idea was missing to me. Thank you very much.

PR #1892, opened 2018-07-19
Vundle installation method added

idvoretskyi:
Vundle (https://github.com/VundleVim/Vundle.vim) is one of the most popular methods to install packages in vim; installation method added.


bhcleek:
Thanks!

PR #1894, opened 2018-07-21
Use the async api for gorename

bhcleek:


autoload/go/rename.vim
```diff
@@ -41,63 +41,34 @@ function! go#rename#Rename(bang, ...) abort
...
+  " checktime to trigger a reload of all files. See
+  " http://www.mail-archive.com/vim@vim.org/msg05900.html for more info
+  " about the autoread bug
+  let current_autoread = &autoread
+  set autoread
+  silent! checktime
+  let &autoread = current_autoread
 
-  call job_start(a:args.cmd, start_options)
+  call call(a:next, [a:job, a:exitval])
```
fatih: Why do we have here two call's? I think I saw this once but I'm not sure to be honest what it means in this context.

bhcleek: The first `call` is  required to call the function named `call`. We have to use the function named `call` to execute the funcref named `a:next`.
 
See :help :call` and `help call()`.

fatih:
lgtm

PR #1895, opened 2018-07-22
restore nvim's support for running go test in a terminal

bhcleek:
Support for running tests in a terminal in Neovim was accidentally dropped in #1864.


fatih:
lgtm

PR #1901, opened 2018-07-26
run gometalinter asynchronously in Neovim

bhcleek:
* Run gometalinter asynchronously and vim-go's job api to to run it both Vim8 and Neovim.
* Correct a couple of documentation errors I came across during implementation.


fatih:
lgtm

PR #1907, opened 2018-08-06
Add operator symbols in regex to show correct field highlight termination

### LINKED ISSUE #1849: go_highlight_fields doesn't highlight when field is followed by operators

Enable struct field names highlightling by `let g:go_highlight_fields = 1`.

struct field names are highlighting

### What happened instead?

When field names are immediately followed by some operators, they are not hightlighting.
Examples:
```
a.b+c
a.b-c
a.b*c
a.b/c
a.b%c
a.b&c
a.b|c
a.b^c
a.b<c
a.b>c
a.b==c
a.b!=c
```

I guess the [regular expression here](https://github.com/fatih/vim-go/blob/master/syntax/go.vim#L293) may be not correct.

* vim-go version: master

```
...
let g:go_highlight_fields = 1
...
```

```
VIM - Vi IMproved 8.1 (2018 May 17, compiled May 25 2018 00:52:08)
Included patches: 1-22
Compiled by Arch Linux
```
* Go version (`go version`):

`go version go1.10.2 linux/amd64`

zkry:
Fixes #1849. This is indeed a problem even when gofmt is on (ex. inside array index brackets). This should cover all of the operators.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -72,7 +72,7 @@ hi def link     goComment           Comment
 hi def link     goTodo              Todo
 
 if go#config#HighlightGenerateTags()
-  syn match       goGenerateVariables contained /\(\$GOARCH\|\$GOOS\|\$GOFILE\|\$GOLINE\|\$GOPACKAGE\|\$DOLLAR\)\>/
+  syn match       goGenerateVariables contained /\%(\$GOARCH\|\$GOOS\|\$GOFILE\|\$GOLINE\|\$GOPACKAGE\|\$DOLLAR\)\>/
   syn region      goGenerate          start="^\s*//go:generate" end="$" contains=goGenerateVariables
   hi def link     goGenerate          PreProc
   hi def link     goGenerateVariables Special
@@ -113,7 +113,7 @@ if go#config#HighlightFormatStrings()
   " 5. Match [n] or nothing before a verb
   " 6. Match a formatting verb
   syn match       goFormatSpecifier   /\
-        \([^%]\(%%\)*\)\
+        \%([^%]\%(%%\)*\)\
         \@<=%[-#0 +]*\
         \%(\%(\%(\[\d\+\]\)\=\*\)\|\d\+\)\=\
         \%(\.\%(\%(\%(\[\d\+\]\)\=\*\)\|\d\+\)\=\)\=\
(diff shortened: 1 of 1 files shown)


fatih:
Hi @zkry 

For this kind of PR's a before/after screenshot on what this is fixed is appreciated.

zkry:
Oh, ok. Here is a screenshot showing a simple example of before and after the change.

ekran resmi 2018-08-06 23 46 47

Carpetsmoker:
Also, could you post the results from the benchmarks with `./scripts/bench-syntax`? Then we have an idea how this will affect performance (it's already quite slow). Thanks!

bhcleek:
@Carpetsmoker I benchmarked these changes against master; they end up slowing things down about 10%, but if this PR goes one step further and uses `\%(` where possible, then this PR and the use of `\%(` results in a zero net change in performance.

zkry:
@bhcleek Thanks. I updated the PR with your suggestion. I was having trouble figuring out how to run the benchmark and get the relevant info. I'll write here when I can confirm that there is now slowdown.

bhcleek:
@zkry you'll likely need to replace more `\(` with `\%(` than just that one place; when I tested there were a several places that I updated to use `\%(`....

syntax/go.vim
```diff
@@ -291,7 +291,7 @@ hi def link     goFunctionCall      Type
 
 " Fields;
 if go#config#HighlightFields()
-  syn match goField                 /\.\w\+\([.\ \n\r\:\)\[,]\)\@=/hs=s+1
+  syn match goField                 /\.\w\+\%([}\!=\><\^|&%\/*\-\+.\ \n\r\:\)\[\],]\)\@=/hs=s+1
```
Carpetsmoker: This regular expression is now very hard to understand and/or maintain. It should probably documented better and split over multiple lines.

For example: https://github.com/fatih/vim-go/blob/b81a447a20c59e44fbbcd39e0f8c8c6e1a289afe/syntax/go.vim#L108-L120

(which is still not great, but better than this).

It may also be worth adding `\v` at the start of this pattern for readability.

zkry: That makes sense. I can clean this regex along with updating `\(` with `\%(`. Sorry for having stepped away from this issue for such a long time by the way.

zkry:
I noticed how were weren't using the `\1` - `\9` feature of regexes at all so it seemed safe to update all instances of `\(` to `\%(`. It was a good idea to clean up the regex too. It would always take me forever to understand what it was doing.

bhcleek:
Thanks, @zkry !

PR #1920, opened 2018-08-20
Use the vim-go's async job api for autocompletion

bhcleek:
Use vim-go's async job api for autocompletion in Vim and Neovim.


bhcleek:
Closing for now. I realized that there's an error condition that will not be handled properly by this PR as-is.

PR #1923, opened 2018-08-23
update the supported/tested version of Neovim

bhcleek:
Update the supported and tested version of Neovimn to v0.3.1, the latest
release of Neovim.

### DIFF
--- a/scripts/install-vim
+++ b/scripts/install-vim
@@ -30,7 +30,7 @@ case "$vim" in
 
   "nvim")
     # Use latest stable version.
-    tag="v0.2.0"
+    tag="v0.3.1"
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
Is there a specific reason for this?

bhcleek:
v0.3.1 is the latest version of Neovim. Given that some Neovim apis are different than Vim's, I want to make sure vim-go continues to work with whatever the latest Neovim release is.

bhcleek:
Until Neovim reaches 1.0, it seems prudent test against their latest release, too, as there could be breaking changes between versions until then.

fatih:
lgtm

PR #1925, opened 2018-08-23
fix E734 error with ultisnips

johngrib:
Fixed problem where E734 error in GoUltisnips function.


Carpetsmoker:
It looks like you set `g:UltiSnipsSnippetDirectories` to something other than a list? As far as I understand the docs, it should always be a list?

bhcleek:
> I did not define g:UltiSnipsSnippetDirectories in my .vimrc, but every time I opened the file * .go I got an E734 error on that part.

If it were not defined, then `if !exists("g:UltiSnipsSnippetDirectories")` will evaluate to true, causing `let g:UltiSnipsSnippetDirectories = ["gosnippets/UltiSnips"]` to be executed.

`g:UltiSnipsSnippetDir` has a separate use case from `g:UltiSnipsSnippetDirectories` and has not relevance to vim-go.

Issue #1906, opened 2018-08-06
Support Go modules


bhcleek:
To support Go 1.11 fully, vim-go changes need to be made to handle module mode development.

- [x] support for `go.mod`
- [x] autocompletion (see https://github.com/fatih/vim-go/pull/1853#issuecomment-410776425)
- [x] `:GoImport` tab completion (https://github.com/fatih/vim-go/issues/2212)
- [ ] others....

Obviously, vim-go will be dependent on a number of tools getting updated to complete the support, but there are a number of changes that will need to be made to vim-go itself.

sheng:
+1

But I think this plugin is hard to support the Go modules, at least not until https://github.com/golang/go/issues/24661 is fixed.

fatih:
Added initial support for `:GoModFmt` (auto format the `go.mod` file) and syntax highlighting here: https://github.com/fatih/vim-go/pull/1931

rkuska:
Is it expected that things like auto fmt (and go imports) on save doesn't work outside of GOPATH?

fatih:
@rkuska no, `gofmt` works on a single file and should work outside GOPATH.

rkuska:
Sorry for the noise, I had `let g:go_fmt_fail_silently = 1` and there was syntax errors in my file 🤦‍♂️ but when I `cd`-ed to my original GOPATH location the syntax error wasn't there and go fmt was working so I assumed the above.

makkes:
`:GoDoc` also doesn't seem to work with modules.

powerman:
> `:GoDoc` also doesn't seem to work with modules.

It's not because of modules, it's just Go 1.11 removed this functionality from `godoc`. Looks like we should use `go doc` instead, but it show less information when viewing doc about whole package and thus require request per each single function/method, which means we need very good navigation to make it fast and easier to move between these small chunks of doc. Also I didn't found a way to move from doc to source with `go doc`. Maybe distributing old `godoc` binary with vim-go will solve these issues, but it's unclear if old `godoc` is compatible with modules (unlikely).

thepudds:
Some related discussion in PR #1988

arnottcr:
it looks like there are some issues with the current `go.mod` syntax highlighting processing [pre-release identifiers](https://semver.org/#spec-item-9), want me to fork off a new issue:
[...]

PR #1931, opened 2018-08-30
go.mod file support

fatih:
This adds initial support for the `go.mod` file.  It adds the
followings:

* Syntax highligthing. We highlight keywords, strings, operator and
semver version. It works pretty well for now.
* Auto fmt on save. Command `:GoModFmt` or `Plug(go-mod-fmt)` for custom
  mappings

Before we fully support the semantics of go.mod, I think this initially
will be helpful because I discovered that `go.mod` is read and edited a
lot. So going forward, this will make it easier experimenting with Go
modules.

- [x] Add documentation
- [x] Make sure auto modfmt toggle commands work fine
- [x] Add v1.11 guard for `:GoModFmt`

related: #1906

### DIFF
--- a/autoload/go/mod.vim
+++ b/autoload/go/mod.vim
@@ -1,11 +1,24 @@
+let s:go_major_version = ""
+
 function! go#mod#Format() abort
+  " go mod only exists in `v1.11`
+  if empty(s:go_major_version)
+    let tokens = matchlist(go#util#System("go version"), '\d\+.\(\d\+\) ')
+    let s:go_major_version = str2nr(tokens[1])
+  endif
+
+  if s:go_major_version < "11" 
+    call go#util#EchoError("Go v1.11 is required to format go.mod file")
+    return
+  endif
+
   let fname = fnamemodify(expand("%"), ':p:gs?\\?/?')
 
   " Save cursor position and many other things.
   let l:curw = winsaveview()
 
   " Write current unsaved buffer to a temp file
-  let l:tmpname = tempname() . '.go'
+  let l:tmpname = tempname() . '.mod'
   call writefile(go#util#GetLines(), l:tmpname)
   if go#util#IsWin()
     let l:tmpname = tr(l:tmpname, '\', '/')
@@ -88,13 +101,12 @@ function! s:parse_errors(filename, content) abort
   " list of errors to be put into location list
   let errors = []
   for line in splitted
-    let tokens = matchlist(line, '^\(.\{-}\):\(\d\+\):\(\d\+\)\s*\(.*\)')
+    let tokens = matchlist(line, '^\(.\{-}\):\(\d\+\):\s*\(.*\)')
     if !empty(tokens)
       call add(errors,{
             \"filename": a:filename,
             \"lnum":     tokens[2],
-            \"col":      tokens[3],
-            \"text":     tokens[4],
+            \"text":     tokens[3],
             \ })
     endif
   endfor
@@ -108,7 +120,7 @@ function! s:show_errors(errors) abort
   let l:listtype = go#list#Type("GoModFmt")
   if !empty(a:errors)
     call go#list#Populate(l:listtype, a:errors, 'Format')
-    echohl Error | echomsg "GoModFmt returned error" | echohl None
+    call go#util#EchoError("GoModFmt returned error")
   endif
 
   " this closes the window if there are no errors or it opens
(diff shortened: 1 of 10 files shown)


autoload/go/fix.vim
```diff
@@ -0,0 +1 @@
+/Users/fatih/go/src/github.com/fatih/goautofix/vim/fix.vim
```
bhcleek: This looks like it was mistakenly committed.

autoload/go/mod.vim
```diff
@@ -0,0 +1,128 @@
...
+    call go#list#Populate(l:listtype, a:errors, 'Format')
+    echohl Error | echomsg "GoModFmt returned error" | echohl None
+  endif
+
+  " this closes the window if there are no errors or it opens
+  " it if there is any
+  call go#list#Window(l:listtype, len(a:errors))
+endfunction
+
+function! go#mod#ToggleModfmtAutoSave() abort
```
bhcleek: `fmt` should probably be `Fmt` for consistency...

autoload/go/mod.vim
```diff
@@ -0,0 +1,128 @@
...
+  return errors
+endfunction
+
+" show_errors opens a location list and shows the given errors. If the given
+" errors is empty, it closes the the location list
+function! s:show_errors(errors) abort
+  let l:listtype = go#list#Type("GoModFmt")
+  if !empty(a:errors)
+    call go#list#Populate(l:listtype, a:errors, 'Format')
+    echohl Error | echomsg "GoModFmt returned error" | echohl None
```
bhcleek: You can use `go#util#EchoError` instead...

autoload/go/mod.vim
```diff
@@ -0,0 +1,128 @@
...
+
+  return errors
+endfunction
+
+" show_errors opens a location list and shows the given errors. If the given
+" errors is empty, it closes the the location list
+function! s:show_errors(errors) abort
+  let l:listtype = go#list#Type("GoModFmt")
+  if !empty(a:errors)
+    call go#list#Populate(l:listtype, a:errors, 'Format')
```
bhcleek: Instead of passing in the parsed errors and calling `go#list#Populate` manually, can you use an `errorformat` string and call `go#list#ParseFormat`?

autoload/go/mod.vim
```diff
@@ -0,0 +1,128 @@
...
+endfunction
+
+" parse_errors parses the given errors and returns a list of parsed errors
+function! s:parse_errors(filename, content) abort
+  let splitted = split(a:content, '\n')
+
+  " list of errors to be put into location list
+  let errors = []
+  for line in splitted
+    let tokens = matchlist(line, '^\(.\{-}\):\(\d\+\):\(\d\+\)\s*\(.*\)')
```
bhcleek: This regex is relatively easy to turn into an `errorformat` string. It looks like it's follows the usual pattern, so the `errorformat` string would be `%f:%l:%c:\ %m`

ftplugin/go/commands-fix.vim
```diff
@@ -0,0 +1 @@
+/Users/fatih/go/src/github.com/fatih/goautofix/vim/commands-fix.vim
```
bhcleek: Is this another mistakenly added file?

ftdetect/gofiletype.vim
```diff
@@ -31,4 +31,8 @@ au BufReadPost *.s call s:gofiletype_post()
 
 au BufRead,BufNewFile *.tmpl set filetype=gohtmltmpl
 
+au BufNewFile *.mod setfiletype gomod | setlocal fileencoding=utf-8 fileformat=unix
+au BufRead *.mod call s:gofiletype_pre("gomod")
+au BufReadPost *.mod call s:gofiletype_post()
```
Carpetsmoker: This conflicts with a built-in filetype: https://github.com/vim/vim/blob/master/runtime/filetype.vim#L996-L1002

Would probably be better to explicitly use `go.mod`, and maybe even check if the first line matches `module <ImportPath>`?

bhcleek: Good catch @Carpetsmoker !

`go.mod` seems like the right call to me, maybe _in conjunction_ with check the first line?

autoload/go/fix.vim
fatih: Yeah, it was symlinked (I was lazy) from a personal project for a small tool I wrote and wanted to try out. Removed them :)

ftdetect/gofiletype.vim
fatih: I'll add the checking line 👍  However, can I really use a `.` inside a ft? I checked the link martin gave and a filetype is always a single word. I think we keep it `gomod`. It's not that bad.

autoload/go/mod.vim
fatih: I just tried it, but I think it'll not work because we run the command on a temporary file first (because `go mod edit` does an in-place edit). This means that the initial `%f` is the temporary filename. This is copied from `fmt.vim` and it's the same issue. I'll check if I can do something ( I would prefer `ParseFormat` indeed)

bhcleek: Hmmmm... If you know the temporary filename, then it should be easy enough to do a `substitute(...)` on `a:errors` before passing to `ParseFormat`.

fatih: I think at this point, this is too much a hack. We're having a working version and it really works fine. Not sure if it's worth doing it (same applies for gofmt).

bhcleek:
lgtm, but there's linting problems.

voutasaurus:
> lgtm, but there's linting problems.
https://github.com/fatih/vim-go/pull/1931/files#diff-d7201d5be221f16a9b1825e6a48685e9R1905

It's just one line in the docs that's over the line limit (needs to be wrapped/rewritten). I'd fix it if you want (I don't have access though).

fatih:
@bhcleek @voutasaurus thanks! It was late after I pushed it, so didn't check the Travis results. I fixed and force pushed it. I'll merge it once it's green.

PR #1933, opened 2018-08-31
refactor guru jobs to use vim-go's async api

bhcleek:


autoload/go/guru.vim
```diff
@@ -180,78 +122,35 @@ function! s:async_guru(args) abort
...
-    let l:start_options.in_name = l:tmpname
+    let l:opts.in_io = "file"
+    let l:opts.in_name = l:tmpname
   endif
 
-  call go#statusline#Update(state.status_dir, {
-        \ 'desc': "current status",
-        \ 'type': state.statusline_type,
-        \ 'state': "analysing",
-        \})
```
fatih: Is this removed accidentally? This is a core part of my workflow because otherwise, I don't have any idea what's going on. Same when it's finished (Above it's also removed)

bhcleek: It's removed on purpose. Our job's api supports statusline updates. But, I'd misspelled the key. That's been fixed, and I see the status updates now. PTAL.

fatih: Just tested again and it works fine 👍

autoload/go/job.vim
```diff
@@ -254,6 +267,19 @@ function! go#job#Start(cmd, options)
     let l:options = s:neooptions(l:options)
   endif
 
+  " Verify that the working directory for the job actually exists. Return
+  " early if the directory does not exist. This helps avoid errors when
+  " working with plugins that use virtual files that don't actually exist on
+  " the file system.
+  let dir = expand("%:p:h")
+  if has_key(l:options, 'cwd')
+    if !isdirectory(l:options.cwd)
```
fatih: This can be simplified to `if has_key(l:options, 'cwd')  && !isdirectory(l:options.cwd)` I guess ?

fatih:
lgtm with a nit comment 👍

PR #1934, opened 2018-08-31
Improve go.mod filetype detection

Carpetsmoker:
Comments are allowed in the go.mod file, so previously:

	// Document stuff!

	module foo/bar

	require "other/pkg"

Would not get detected properly.

Also tweak regexp slightly: `module` must be followed by a space and at
least one character.

### DIFF
--- a/ftdetect/gofiletype.vim
+++ b/ftdetect/gofiletype.vim
@@ -31,9 +31,23 @@ au BufReadPost *.s call s:gofiletype_post()
 
 au BufRead,BufNewFile *.tmpl set filetype=gohtmltmpl
 
-" make sure we explicitly look for a `go.mod` and the `module` starts from the
-" beginning
-au BufNewFile,BufRead go.mod
-	\ if getline(1) =~ '^module.*' | set filetype=gomod |  endif
+" Set the filetype if the first non-comment and non-blank line starts with
+" 'module <path>'.
+au BufNewFile,BufRead go.mod call s:gomod()
+
+fun! s:gomod()
+  for l:i in range(1, line('$'))
+    let l:l = getline(l:i)
+    if l:l ==# '' || l:l[:1] ==# '//'
+      continue
+    endif
+
+    if l:l =~# '^module .\+'
+      set filetype=gomod
+    endif
+
+    break
+  endfor
+endfun
 
 " vim: sw=2 ts=2 et


Carpetsmoker:
There is a second problem, by the way. The `modsim3` is still loaded, meaning that settings, highlights, etc. are loaded from that as well, which may be unexpected. I'm not sure how to fix/prevent that.

There is only `syntax/modsim3.vim`, which doesn't do anything strange. Since this is a very old/obscure tech, I don't expect any changes, and this should be fine (but I was curious if it could be done "more correct").

ftdetect/gofiletype.vim
```diff
@@ -31,9 +31,23 @@ au BufReadPost *.s call s:gofiletype_post()
...
+    let l:l = getline(l:i)
+    if l:l ==# '' || l:l[:1] ==# '//'
+      continue
+    endif
+
+    if l:l =~# '^module .\+'
+      set filetype=gomod
+    endif
+
+    break
```
bhcleek: Should this break be moved into the previous `if` block?

Carpetsmoker: No, once we see a non-blank line or a line that is not a non-comment we don't need to look any further. `module pkg` must the first line.

bhcleek:
lgtm

bhcleek:
Can you use `syntax clear` to clear the syntax of `modsim3` ?

Carpetsmoker:
Changing the syntax automatically clears the syntax, just not the highlight groups. `hi clear` doesn't seem to do that.

The real danger is mostly from settings that some filetypes set (e.g. `tabstop`, `expandtab`, `comments`, etc.).

bhcleek:
I think that simply adding adding a `!` to the `au` instruction do what you're looking for in this case.

bhcleek:
Since we're using `go.mod` as the pattern, we probably need to clear the existing groups first by having an explicit `au! BufNewFile,BufRead *.mod` _before_ our `au BufNewFile,BufRead go.mod`

Carpetsmoker:
The filetype is already set, so I don't see how clearing the autogroups will undo that?

Ideally Vim's `filetype.vim` should be fixed/amended; I'll make a PR in the Vim repo, too.

bhcleek:
The filetype is already set because the autogroups were applied. If you clear the autogroup that `filetype.vim` sets up, then its command (which ends up setting the filetype), won't be executed at all....

PR #1936, opened 2018-09-02
cmd: fix status line updated upon completion

fatih:
`statustype` needs to be set in order to complete the statusline update
in job.vim


bhcleek:
lgtm

PR #1937, opened 2018-09-02
Display statuslines consistently

bhcleek:
* Do not update the status line when guru commands are run via
  autocommands.
* Update the status line when gocode is run via :GoInfo.

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -66,15 +66,15 @@ function! go#complete#GetInfo() abort
   return s:sync_info(0)
 endfunction
 
-function! go#complete#Info() abort
+function! go#complete#Info(showstatus) abort
   if go#util#has_job(1) || has('nvim')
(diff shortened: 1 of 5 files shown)


fatih:
I still think we should remove the `codecov/patch` check. It shows things in red where they shouldn't be :/  I think we tweaked it once?

bhcleek:
This handles two UX issues:
1. When running a long-running command (e.g. `:GoReferrers`) with `g:go_auto_sameids=1` or `g:go_auto_type_info=1` , the user would see the statusline updated to say the long-running command was started, but then quickly see the statusline updated with info that the command `CursorHold` autocmd completed successfully, hiding the `started` statusline update for the long-running command.
2. Sets the statusline for `:GoInfo` regardless of the value of `g:go_info_mode`.

fatih:
lgtm for the PR 👍

PR #1939, opened 2018-09-02
highlight predefined identifiers

bhcleek:
#### Match builtins as keywords and highlight with the Identifier group.

This was changed from keywords to the regexp in #605, but I think that
was a mistake.

If you do `new := "foo"` then this is perfectly valid Go, but I don't
think it's a bad thing to add a reminder that you're actually overriding
a global built-in by highlighting the `new`, just as it's not a bad idea
to highlight the `new()` in `func new()` to serve as a similar reminder.

This is a common-ish mistake that most of us have probably made at least
once or twice.

The only case where this fails is in method declaration and calls:

    func () x new() {
      other.new()
    }

But this is the case with the current highlight as well, so it doesn't
make that worse.

_If users want the previous (or different) behavior, they are free to
link the groups to different highlighting groups in their configs._

edit: removed the description of a commit that was removed from the PR.

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -47,12 +47,12 @@ hi def link     goFloats            Type
 hi def link     goComplexes         Type
 
 " Predefined functions and values
-syn match       goBuiltins                 /\<\v(append|cap|close|complex|copy|delete|imag|len)\ze\(/
-syn match       goBuiltins                 /\<\v(make|new|panic|print|println|real|recover)\ze\(/
+syn keyword     goBuiltins                 append cap close complex copy delete imag len
+syn keyword     goBuiltins                 make new panic print println real recover
 syn keyword     goBoolean                  true false
 syn keyword     goPredefinedIdentifiers    nil iota
 
-hi def link     goBuiltins                 Keyword
+hi def link     goBuiltins                 Identifier
 hi def link     goBoolean                  Boolean
 hi def link     goPredefinedIdentifiers    goBoolean
 


bhcleek:
This applies the changes from #1782 with the modifications that I suggested there and extends it just a bit for the predeclared identifiers that vim-go already had groups for, but which weren't being highlighted.

Carpetsmoker:
> The goBoolean and goPredefinedIdentifer groups were the only groups that contained any of the predeclared identifiers that were not highlighted out-of-the-box with vim-go.

They are? Both link back to `Boolean`, which is red in the default colour scheme.

This has been discussed extensively in #1782, as well as in Slack, and a similar change was reverted in #1030. I think your proposal is poor UX. Highlighting e.g. `append` the same as `true` makes no sense. Adding more colours is more subjective, but I see no reason to do that, either; especially since it's a bright colour in the default colour scheme which stands out a lot.

When it comes to syntax highlighting no one cares what the spec says, for all practical purposes `true`, `false`, `nil`, and `iota` are special. That you can do `true := "x"` is correct, but also stupid. Literally no one does that. We should do what is helpful UX, not attempt to parse Go code according to the spec.

bhcleek:
> They are? Both link back to Boolean, which is red in the default colour scheme.

Perhaps it's bright; iTerm2 has has some issues with its solarized color scheme, so that's probably just an issue on my machine. I'll revert that piece.

bhcleek:
I force pushed to remove that second commit.

bhcleek:
Closed since I'm travelling today. Martin's previous PR be reopened and merged to get the desired result.

fatih:
Would like to see a FAQ entry describing how to change the syntax highlighting so the builtins are highlighted as keywords.

fatih:
lgtm

PR #1943, opened 2018-09-03
Fix async job errors

### LINKED ISSUE #1942: :GoBuild doesn't show an error if main function is not declared
```
package main

var T string
```

`:GoBuild` doesn't show an error if `main()` is missing. This is the output of `go build`:

```
$ go build
# demo
runtime.main_main·f: function main is undeclared in the main package
```

Produces by Go version `1.11`.

bhcleek:
* Echo errors when they cannot be parsed. Put the directory at the
beginning of the echoed messages to give context to the messages that
follow. The user may still need to call `:messages` to see the full list of errors....

* Remove redundant setting of statusline

Fixes #1942

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -284,15 +284,6 @@ endfunction
 " ---------------------
 
 function! s:cmd_job(args) abort
-  let status_dir = expand('%:p:h')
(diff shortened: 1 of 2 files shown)


fatih:
lgtm

PR #1946, opened 2018-09-03
add test for go#complete#GetInfo

bhcleek:
Add a test for `go#complete#GetInfo`, because it has guarantees, and some users depend on it.


fatih:
lgtm

PR #1947, opened 2018-09-03
disable other *.mod autocmds

bhcleek:
Disable the highlight settings and syntax groups for other *.mod files.

The one downside of this is that a user that uses vim-go and also works in Module2, LambdaProlog, and Modsim III files will need to set the filetype for those other languages via _another_ autocmd. Is there really any vim-go user that also works one of those three other languages?

### DIFF
--- a/ftdetect/gofiletype.vim
+++ b/ftdetect/gofiletype.vim
@@ -31,6 +31,11 @@ au BufReadPost *.s call s:gofiletype_post()
 
 au BufRead,BufNewFile *.tmpl set filetype=gohtmltmpl
 
+" remove the autocommands for modsim3, and lprolog files so that their
+" highlight groups, syntax, etc. will not be loaded. *.MOD is included, so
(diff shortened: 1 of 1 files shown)


fatih:
Currently only `go.mod` will be matched to `gomod`. Everything else we don't care and the other file detects files will kick in. Why do we need this in this case?

bhcleek:
The default filetype detection in vim sets the files type for all `*.mod` files to one of the three languages I mentioned by using an autocommands. This PR _removes_ those autocommands so that the highlighting and syntax groups for those three files will not be loaded. As our master currently stands, if you run `verbose syntax`, for instance, you'll see the syntax groups that vim-go has created _and_ all the syntax groups for the other languages, too. e.g.:
```
lprologBrackErr xxx links to Error
        Last set from /usr/local/Cellar/vim/8.1.0250/share/vim/vim81/syntax/lprolog.vim
```

With this change, none of those other filetype's highlight groups will be set.

PR #1948, opened 2018-09-04
Homogenize async progress

bhcleek:
#### jobs: improve progress echo

* Only echo status when g:go_echo_command_info is set.
* Echo status when starting a job.

#### improve use of statustype and progress messages

* Make sure all async jobs set the statustype option key so that their
  progress will be echoed appropriately.
* Move some progress messages that were written just prior to starting
  an async job to be written just before executing the similar
  synchronous execution (e.g. in `go#cmd#Test`).
* Move the "analysing with scope" message in guru so that it will be
  written _after_ the async job is started and make sure it's guarded by
  a `go#config#EchoProgress` check.
* Remove remaining progress messages that were written just prior to
  starting async jobs.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -32,10 +32,6 @@ function! go#cmd#Build(bang, ...) abort
 
   " Vim and Neovim async.
   if go#util#has_job() || has('nvim')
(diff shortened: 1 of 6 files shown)


fatih:
lgtm

PR #1949, opened 2018-09-04
job: do not jump to windows if 'for' is set to '_'

### LINKED ISSUE #1944: :GoDef doesn't jump to identifier if it's opened in a new window
Using `:GoDef` on a vertical split doesn't jump to the identifier.  To reproduce:

1. Add following to `vimrc`
```
  autocmd FileType go nmap <silent> <Leader>v <Plug>(go-def-vertical)
```

2. Call `<leader>v` on an identifier
3. Observe how the cursor is still on the same position and didn't jump to the newly opened vertical split.

`gd`  works fine, but `-vertical`, `-split` and `-tab` no more jump to the identifier.

fatih:
fixes #1944

### DIFF
--- a/autoload/go/job.vim
+++ b/autoload/go/job.vim
@@ -91,14 +91,22 @@ function! go#job#Options(args)
   " do nothing in state.complete by default.
   function state.complete(job, exit_status, data)
     if has_key(self, 'custom_complete')
-      let l:winid = win_getid(winnr())
       " Always set the active window to the window that was active when the job
       " was started. Among other things, this makes sure that the correct
       " window's location list will be populated when the list type is
       " 'location' and the user has moved windows since starting the job.
-      call win_gotoid(self.winid)
+      " Make sure to jump only if 'for' was not set explicitly to '_', which
+      " is an indicator that the process will show a list
+      if has_key(self, 'for') && self.for != '_'
+        let l:winid = win_getid(winnr())
+        call win_gotoid(self.winid)
+      endif
+
       call self.custom_complete(a:job, a:exit_status, a:data)
-      call win_gotoid(l:winid)
+
+      if has_key(self, 'for') && self.for != '_'
+        call win_gotoid(l:winid)
+      endif
     endif
 
     call self.show_errors(a:job, a:exit_status, a:data)
--- a/autoload/go/job.vim
+++ b/autoload/go/job.vim
@@ -97,14 +97,14 @@ function! go#job#Options(args)
       " 'location' and the user has moved windows since starting the job.
       " Make sure to jump only if 'for' was not set explicitly to '_', which
       " is an indicator that the process will show a list
-      if has_key(self, 'for') && self.for != '_'
+      if get(self, 'for', '_') != '_'
         let l:winid = win_getid(winnr())
         call win_gotoid(self.winid)
       endif
 
       call self.custom_complete(a:job, a:exit_status, a:data)
 
-      if has_key(self, 'for') && self.for != '_'
+      if get(self, 'for', '_') != '_'
         call win_gotoid(l:winid)
       endif
     endif


autoload/go/job.vim
```diff
@@ -91,14 +91,22 @@ function! go#job#Options(args)
...
     if has_key(self, 'custom_complete')
-      let l:winid = win_getid(winnr())
       " Always set the active window to the window that was active when the job
       " was started. Among other things, this makes sure that the correct
       " window's location list will be populated when the list type is
       " 'location' and the user has moved windows since starting the job.
-      call win_gotoid(self.winid)
+      " Make sure to jump only if 'for' was not set explicitly to '_', which
+      " is an indicator that the process will show a list
+      if has_key(self, 'for') && self.for != '_'
```
Carpetsmoker: You should be able to use `get()` here, which works like in Python:

    if get(self, 'for', '') != ''

It will return the third parameter (`''`) if the `for` key doesn't exist.

More readable, IMHO.

Same below.

fatih: Good call. However I've change it to use `_`, because otherwise this wouldn't work. We shouldn't call the win commands in case it's also set to `_` which is the case when you don't want to use `for`

bhcleek:
`for` should always be set. Instead of trying to deal with it being missing, let's make sure it's set....

bhcleek:
I strongly believe this PR is the wrong thing to do. Here are my concerns:
1. The contract for `complete` is that the window that was active when the job was started will be active when `complete` is called; `for` should not be a consideration for whether that is true or not.
2. `_for` is always set, so there's no need to use a default value with it.
3. If this is causing a problem for #1944 for some reason, then something else is wrong. We should figure out what that something else is and fix that, especially in light of the fact that I'm unable to reproduce #1944 ... 🤔

bhcleek:
Also, there are other cases where `for` is `_`, and the window that started the job expects to be active  when a custom complete is called; this PR would break those. Again, the contract for `complete` is a guarantee that we shouldn't break.

If the jumping to another window is breaking something for you; what's actually happening? Why would changing the window be a problem for you? And finally, perhaps we need to fix the custom complete that's being called in your case for some reason or maybe even wrap the callbacks that come back from `go#job#Options` in the `def` callsite to change the window in some edge case.

fatih:
> _for is always set, so there's no need to use a default value with it.

I don't set anything here. It just make sure `get()` returns a `_` that marks it as empty so the logic handles two cases, when somehow `for` is left out (which never happens of course because it's set to `_job` and the case it's set to `_` which is in the case of guru

> If this is causing a problem for #1944 for some reason, then something else is wrong. We should figure out what that something else is and fix that, especially in light of the fact that I'm unable to reproduce #1944 ... 🤔

This code literally jumps back to the window the job was started. If I remove this piece, everything works fine. I'm not sure how else I can describe it. I git bisected it, and the also gave the SHA that causes the issue.  It just disables jumping away (Which happens because `:GoDef` jumps away. ) Not sure how else I can describe the issue. 

> Also, there are other cases where for is _, and the window that started the job expects to be active when a custom complete is called; this PR would break those.

I'm ok using something else to signal that I don't want to jump back to the window I executed the Job. I used `for` because it makes sense as it assumes a list is created for that particular window.

fatih:
> Why would changing the window be a problem for you? 

Because split, vertical and tab all create a *new* window, which is different than the one you called the job. `gd` works fine here, because you still use the same window.

bhcleek:
> I don't set anything here. It just make sure get() returns a _ 

My point is that the code can be simplified there; you can _rely_ on on `for` always having been set; there's no need for the `get`.

> This code literally jumps back to the window the job was started. If I remove this piece, everything works fine. I'm not sure how else I can describe it. I git bisected it, and the also gave the SHA that causes the issue. It just disables jumping away (Which happens because :GoDef jumps away. ) Not sure how else I can describe the issue.
> Because split, vertical and tab all create a new window, which is different than the one you called the job. gd works fine here, because you still use the same window.

The active window should be the window that started the job _so that correct window_ will be split. Otherwise, the user could do `gd`, jump to another window while the job is running, and then wrong window will be split.

The problem that you're encountering _must_ be in `go#def#jump_to_declaration`.... By chance, do you have `g:go_def_reuse_buffer` set?

fatih:
> The active window should be the window that started the job so that correct window will be split. Otherwise, the user could do gd, jump to another window while the job is running, and then wrong window will be split.

What happens here is:

1. `guru` is called inside `def.vim`
2. `state.complete` is called in `job.vim`
3. Current winid is stored (`1)`
4. Jumps back to the old winid (which is still `1`)
5. `s:custom_complete` callback is finished, which is bound to `jump_to_declaration`
6. A vertical split is open (which has id `2`) (this happens inside `jump_to_declaration`)
7. Inside `jump_to_declaration`, it tries to jump to the correct position and then the function finishes
8 .Inside `state.complete`, `call win_gotoid(l:winid)` is called, which jumps back the original window.

The issue is that `7` and `8` looks like they are serialized, but that's not the case. Either `call cursor()` is slow or something else, it won't work anymore because my cursor stays in windows id `1`.

> The problem that you're encountering must be in go#def#jump_to_declaration.... By chance, do you have g:go_def_reuse_buffer set?

I don't have it. This is my vimrc: 
https://github.com/fatih/dotfiles/blob/2c0029bd7e6d52ba67819d681227b764b6f2d71b/vimrc#L397

I'm using this setup for many months without any issues on jumping. It literally broke after `2eb76d5ece14519cf48cace0799f1787e4a738b1`.

bhcleek:
@fatih, can you try #1950 ?

PR #1950, opened 2018-09-04
def: make sure to make the declaration's window active

### LINKED ISSUE #1944: :GoDef doesn't jump to identifier if it's opened in a new window

bhcleek:
Fixes #1944

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -43,9 +43,10 @@ function! go#def#Jump(mode) abort
     call extend(cmd, ["definition", fname . ':#' . go#util#OffsetCursor()])
 
     if go#util#has_job() || has('nvim')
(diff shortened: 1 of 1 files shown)


fatih:
This fixes the issue. Thanks for fixing it the proper way. lgtm

PR #1952, opened 2018-09-05
fix mappings for go-info and go-sameids

bhcleek:
Pass a value for the showstatus argument to go#tool#Info and
go#guru#SameIds in the go-info and go-sameids mappings.


fatih:
lgtm

PR #1956, opened 2018-09-08
snippet: Add table testing boilerplate snippet

mtyurt:
The snippet only considers string input / output, and it needs to be modified for each test case, but it provides a clean, parallel testing compatible table testing boilerplate code.

Let me know if you have suggestions about tab stops or snippet name.

### DIFF
--- a/gosnippets/UltiSnips/go.snippets
+++ b/gosnippets/UltiSnips/go.snippets
@@ -363,6 +363,28 @@ func Test${1:Function}(t *testing.T) {
 }
 endsnippet
 
+# test table snippet
+snippet tt
+var tests = []struct {
+	name string
+	expected string
+	given string
+}{
+	{"${1}", "${2}", "${3}",},
+}
+for _, tt := range tests {
+	tt := tt
+	t.Run(tt.name, func(t *testing.T){
+		actual := ${0:${VISUAL}}(tt.given)
+		if actual != tt.expected {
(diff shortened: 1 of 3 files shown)


Carpetsmoker:
Thanks!

This seems very specific for your personal use-case though. I'm not sure if adding it to vim-go is a good idea.

(also, as a general point, you should probably use subtests and print errors as aligned output).

Carpetsmoker:
If that is from Effective Go then that document should be updated/changed.

bhcleek:
@mtyurt, if you'll modify it so that each entry in the table has a `name`, the tests are run via `
```
t.Run(tt.name, func(t *testing.T) {
   ...
})
```
and all the snippet is available to all of vim-go's supported engines, I'm sure we can get this merged.

mtyurt:
I have updated the snippet to use subtests and added minisnip / neosnip versions as well.

fatih:
This looks good. Only change I would recommend is to change the name `testtable` (it's a tad long). Let's use `tt`.

mtyurt:
That's quite a better name, I hope it doesn't confuse people :)

Renamed to `tt`

fatih:
@mtyurt true, but then a long name means you need to write a lot, which is an anti-pattern for a snippet. We already have many 2-3 char snippets (i.e: `err`, `errf`, `hf`, etc..). People get used to it so it's not a big issue :)

Issue #1960, opened 2018-09-10
hide empty dispatch message (seems to be a leftover from debugging?)


EVODelavega:
### What did you do? (required. The issue will be **closed** when not provided.)

After updating the vim-go plugin, I noticed the occasional message saying `vim-go: [] dispatched` appearing. I had a quick look and found it to originate from go/job.vim (lines 146,148).

### What did you expect to happen?

If there's nothing to display, then there shouldn't be any output

### What happened instead?

I saw `vim-go: [] dispatched

### Configuration (**MUST** fill this out):

* vim-go version:
Current master:

```
commit fb173c3a849fdc47a267e905cee5e29a88797d61 (HEAD -> master, origin/master, origin/HEAD)
Date:   Thu Sep 6 10:11:49 2018 +0300
     Merge pull request #1954 from bhcleek/add-GoModFmt-list
```

* `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):

Pretty standard setup, WRT vim-go, bar a few keybindings almost all of it is standard:

```
set nu rnu smarttab ai smartindent smartcase wildmenu nocompatible expandtab hlsearch modeline
color elflord
set tabstop=4
set shiftwidth=4
set ch=2
set pastetoggle=<F2>
" Add pwd to path
set path+=**

"" swap files in ~/.vim/tmp
" automatically create the dir if needed
silent execute '!mkdir -p ~/.vim/tmp'
" backups are good
set backup
set writebackup
" set the paths
set backupdir=~/.vim/tmp
set backupskip=/.vim/tmp/*
set directory=~/.vim/tmp

" Fold stuff
" Default to indent fold level
" Based on file type, we will change this to syntax or manual
set foldmethod=indent
set foldlevel=2

filetype plugin on

autocmd FileType go set path +=$GOPATH/src/**
let g:go_fmt_command = "goimports"

```

* Vim version (first three lines from `:version`):

```
VIM - Vi IMproved 8.1 (2018 May 18, compiled Aug 27 2018 11:23:25)
Included patches: 1-328
Modified by <bugzilla@redhat.com>
```

* Go version (`go version`):

go version go1.10.4 linux/amd64

* Go environment (`go env`):

```
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/evo/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
[...]

PR #1961, opened 2018-09-10
Hide nagging message when there is nothing to report

EVODelavega:
Small fix for issue #1960

### DIFF
--- a/autoload/go/job.vim
+++ b/autoload/go/job.vim
@@ -143,7 +143,7 @@ function! go#job#Options(args)
   endif
 
   function! s:start(args) dict
-    if go#config#EchoCommandInfo()
+    if go#config#EchoCommandInfo() && self.statustype != ""
       let prefix = '[' . self.statustype . '] '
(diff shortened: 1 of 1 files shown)


bhcleek:
All async jobs should now be setting the statustype. Which jobs are you experiencing this problem with?

bhcleek:
I see how this is happening now. Thanks for the PR!

LGTM

EVODelavega:
@bhcleek  I've not looked at this in-depth, but I'd be curious to know where this message was coming from. If it's not too much of a pain, could you point me in the right direction? Maybe I can help to fix this in a more elegant way, instead of checking for empty strings

bhcleek:
This was actually the _correct_ fix; the `statustype` being empty is expected in this case.

The job being executed was probably guru as a consequence of having `g:go_auto_type_info` or `g:go_auto_sameids` set, which in turns causes the `statustype` to be empty. See https://github.com/fatih/vim-go/blob/master/autoload/go/guru.vim#L137-L139

PR #1964, opened 2018-09-10
Define call of go#decls#Decls as go-decls-dir for further mappings 

Funfun:
I'd like to call `:GoDeclsDir` via hotkey as follows:
```vim
autocmd FileType go nmap <Leader>l <Plug>(go-decls-dir)
``` 
It appears the command needs to be defined in the mappings before its usage. This PR adds that definition.


fatih:
Hi @Funfun. Thanks for adding the plug mapping. Can you also add the `go-decls` mapping so have both of them?

Funfun:
@fatih Done. Do you merge as it is or squash on merge or should I squash locally?

fatih:
@Funfun please squash them locally, makes things easier for us. Thanks

PR #1967, opened 2018-09-13
default gocode to propose from source

bhcleek:
Although this will make gocode slower by default, it provides the experience that most people expect. Reliability seems more important than performance... 🤔

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -145,7 +145,7 @@ function! go#config#GocodeProposeBuiltins() abort
 endfunction
 
 function! go#config#GocodeProposeSource() abort
-  return get(g:, 'go_gocode_propose_source', 0)
+  return get(g:, 'go_gocode_propose_source', 1)
 endfunction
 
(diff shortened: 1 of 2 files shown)


Carpetsmoker:
I think this makes sense.

fatih:
lgtm

myitcv:
Just an FYI this change causes things to break with https://github.com/stamblerre/gocode, which is the module-aware version of `gocode` that will soon be merged back into https://github.com/mdempsky/gocode

Options here would be to revert this change or to make https://github.com/stamblerre/gocode accept a `-source` flag which is effectively a no-op.

I think it probably makes sense to do the latter; `vim-go` might not be the only thing relying the "API" that is `-source`.

cc @stamblerre @fatih

stamblerre:
Thanks for noticing this, Paul - just added the source flag back since it's definitely the simplest fix (https://github.com/stamblerre/gocode/commit/73d4bb62821e92ba7f8f76c816831fbc38e017cf).

PR #1972, opened 2018-09-14
remove unneeded nvim checks

bhcleek:
Now that vim-go's async api support Neovim, go#util#has_job can handle
checking for Neovim instead of checking it inline.


fatih:
lgtm

PR #1975, opened 2018-09-18
add guru describe information in balloon

### LINKED ISSUE #1963: Add vim balloon support for :GoInfo
Just discovered this feature from a coworker (thanks @lxfontes).  We should add this feature to `:GoInfo` as it would make a lot of sense. Or for other features (need to think about them).

Here is how you can test it (works with `iTerm2` but not `Alacritty`). See `:help balloon-eval` and `:help balloonexpr` for more information.

1. Create file `testballoon.vim`:

```viml
let s:counter = 0
let s:timer = -1

function! MyBalloonExpr()
  let s:counter += 1
  call timer_stop( s:timer )
  let s:message =
          \ 'Cursor is at line ' . v:beval_lnum .
          \', column ' . v:beval_col .
          \ ' of file ' .  bufname(v:beval_bufnr) .
          \ ' on word "' . v:beval_text . '"'
  let s:timer = timer_start( 0, 'RealBalloonExpr' )
endfunction

function! RealBalloonExpr(timer)
  echom 'MyBalloonExpr: ' . s:counter
  call balloon_show( s:message )
endfunction

set mouse=a
set ttymouse=sgr
set balloonexpr=MyBalloonExpr()
set balloondelay=250
set ballooneval
set balloonevalterm
```

2. Run it with:

```
vim -Nu NONE -S testballoon.vim testballoon.vim
```

3. Put your cursor on top of `MyBallonExpr()` and observe what happens:

image

bhcleek:
Teach vim-go how to display balloons with information from guru describe.

During development, I had to experiment a bit with the best way to hook this up. During those experiments I made a few other topical changes:
* DRY out the `:GoInfo` display code so there's only one place that has to change if we modify it again instead of having to remember to do it for both modes.
* Use the config functions in `plugin/go.vim` instead of referencing the config values explicitly.
* Remove some extra spaces.

A few things to note for any testers:
* the balloons work well with iTerm, but not in tmux. It's not clear if there's anyway to get them to work in tmux.
* Neovim does not support balloons.
* I explicitly added an early return when jobs are not available to avoid the performance problems that synchronous calls to populate the balloon would cause...

Fixes #1963

### DIFF
--- a/autoload/go/auto.vim
+++ b/autoload/go/auto.vim
@@ -12,7 +12,7 @@ function! go#auto#template_autocreate()
 endfunction
 
 function! go#auto#echo_go_info()
-  if !get(g:, "go_echo_go_info", 1)
+  if !go#config#EchoGoInfo()
     return
   endif
 
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -207,8 +207,8 @@ function! s:info_filter(echo, result) abort
 endfunction
 
 function! s:info_complete(echo, result) abort
-  if a:echo && !empty(a:result)
-    echo "vim-go: " | echohl Function | echon a:result | echohl None
+  if a:echo
+    call go#util#ShowInfo(a:result)
   endif
 
   return a:result
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -282,8 +282,8 @@ function! go#debug#Stop() abort
   silent! exe bufwinnr(bufnr('__GODEBUG_OUTPUT__')) 'wincmd c'
 
   if has('balloon_eval')
-    set noballooneval
-    set balloonexpr=
+    let &noballooneval=s:ballooneval
+    let &balloonexpr=s:balloonexpr
   endif
 
   augroup vim-go-debug
@@ -466,6 +466,9 @@ function! s:start_cb() abort
   nnoremap <silent> <Plug>(go-debug-print)      :<C-u>call go#debug#Print(expand('<cword>'))<CR>
 
   if has('balloon_eval')
+    let s:balloonexpr=&balloonexpr
+    let s:ballooneval=&ballooneval
+
     set balloonexpr=go#debug#BalloonExpr()
     set ballooneval
   endif
--- a/autoload/go/guru.vim
+++ b/autoload/go/guru.vim
@@ -18,6 +18,7 @@ function! s:guru_cmd(args) range abort
   let format = a:args.format
   let needs_scope = a:args.needs_scope
   let selected = a:args.selected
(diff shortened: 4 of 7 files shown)


tmc:
Would love to see this refreshed

arush-sal:
So, I added the following additional lines in my `.vimrc`
```
let s:counter = 0
let s:timer = -1
set mouse=a
set ttymouse=sgr
set balloonexpr=go#tool#DescribeBalloon()
set balloondelay=250
set ballooneval
set balloonevalterm
```
 & tested this with the following setup:


| OS  | Terminal | Shell | tmux | balloon shows |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| Mac 10.14 | iTerm  | zsh(with oh-my-zsh)  | no  | yes  |
| Mac  10.14 | iTerm  | zsh(with oh-my-zsh)  | yes  | yes  |
| Ubuntu | gnone-teminal | zsh(with oh-my-zsh)  | no  | only after a key press  |
| Ubuntu | gnone-teminal | zsh(with oh-my-zsh)  | yes  | only after a key press  |
| Ubuntu | gnone-teminal | bash  | no  | only after a key press  |
| Ubuntu | gnone-teminal | bash  | yes  | only after a key press  |

On Mac the setup worked like a charm, w/o any issues. But on Ubuntu(with and without tmux) the respective balloon shows up only once if you press a key after you hover the move over the respective element. Even in case of the example mentioned in [1963](https://github.com/fatih/vim-go/issues/1963) the balloon shows up with the value `0` until any key is press, once you press a key then only the correct value is shown instead of `0`.

Any idea what could be causing such a behaviour?

shaneHowearth:
> let s:counter = 0
> let s:timer = -1
> set mouse=a
> set ttymouse=sgr
> set balloonexpr=go#tool#DescribeBalloon()
> set balloondelay=250
> set ballooneval
> set balloonevalterm

Wow, thanks for the .vimrc settings, I am running Ubuntu, terminal, tmux, bash, and with your settings the balloons come up immediately.
I'm on commit 8465d625e9ac8bc2bef90cb22b46669576b57243 

 $ vim --version
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Apr 10 2018 21:31:58)
Included patches: 1-1453
Modified by pkg-vim-maintainers@lists.alioth.debian.org
Compiled by pkg-vim-maintainers@lists.alioth.debian.org

$ tmux -V
tmux 2.6

arush-sal:
@shaneHowearth Lucky you, my friend. I tested the same on a friend's system and everything did work properly and as expected. So, as it seems that the issue is with my system & my setup.

PR #1978, opened 2018-09-19
support and test against Go 1.11

bhcleek:
Update errorformat strings for go test for Go 1.11 test output while
maintaining compatibility with Go 1.10 test output, too. The changes
were validated locally by running the tests in vim-go-test containers
based on the respective versions of Go. I investigated trying to use the
json support, but it doesn't offer any real value for us yet. I may
revisit it later.

Update the go vet expecations; one of the strings changed.

Update to the Dockerfile base image and travis to use Go 1.11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.10.x
+  - 1.11.x
 notifications:
   email: false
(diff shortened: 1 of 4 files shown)


fatih:
lgtm

PR #1981, opened 2018-09-20
Handle indentation in the middle of a function call

dhduvall:


bhcleek:
Thank you for the contribution. In testing, I found two small problems:
1. indentation is incorrect when the cursor is inside a raw string. e.g.:
```
fmt.Println(`some raw string
```
2. indentation is incorrectly determined if there's a mix of braces (admittedly, this would probably be a typo). e.g.:
```
fmt.Println(}
```

dhduvall:
As best I can tell, neither of these behaviors is a regression, though I'm happy to fix what I can.

I've no idea what the right answer should be for the second behavior.

The first is a change, in that prior to my commit, the indent of the next line stays in the same line as the `fmt`, rather than increasing one level, but in neither case does it go back to column 0 (which I assume is what should happen?).

It seems like the simplest thing to fix that is to check if the end of the previous line is in a `goRawString` and that line also contains a backtick (presumably not every line in a raw string should have its indent set to 0), and set the indent to 0 in that case; otherwise, use the indent of the previous line.

This, however, is a regression:
```
        foo = `
first line
```
the next line ends up at the same indentation level as `foo`, rather than using that of the previous line.

I also discovered this, which is unrelated to my change, and I assume is unintentional:
```
foo := `
```
results in the next line being indented to the same level as `foo`, rather than to 0; this seems to be a missing `:\?` in the regex introduced in bf955a35.

bhcleek:
lgtm

Thank you for contributing!

dhduvall:
Hm, so should I submit a new PR? Or do I need to handle the issues w.r.t. raw strings first?

bhcleek:
I've taken care of raw strings in https://github.com/fatih/vim-go/pull/1987. If you want to submit a PR that maintains feature parity with #1981 without noticeably impacting performance, we'll be happy to review it.

PR #1984, opened 2018-09-21
go: install tools with their custom names

fatih:
We're assuming for now to use the same binary name that also matches the
import path. But this is not always the case. To add support for an
experimental `gocode`, we need to install a second binary. This change
will allow us to do it by just adding a line, such as:

      \ 'gocode-gomod':  ['github.com/stamblerre/gocode'],

This PR splits `go get` command into specific download and install
steps.

A follow up PR to add support for gocode-gomod will follow.

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -29,7 +29,7 @@ if
 endif
 
 " these packages are used by vim-go and can be automatically installed if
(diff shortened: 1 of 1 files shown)


bhcleek:
lgtm

PR #1988, opened 2018-09-21
gocode: add support for experimental gocode

fatih:
WIP. Somehow `gocode-gomod` only returns %20 of the time the candidates. To test it you can use `gomodifytags` repo. First pen gocode-gomod in debug mode to see that it doesn't return any candidates:

```
gocode-gomod exit
gocode-gomode -s -debug
```

Clone gomodifytags and play around:

```
git clone https://github.com/fatih/gomodifytags
cd gomodifytags
vim main.go
// try to complete flag. exported identifiers
```

This doesn't work always, not sure why.

For more information see: https://github.com/Microsoft/vscode-go/pull/1932

### DIFF
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -1,5 +1,11 @@
 function! s:gocodeCommand(cmd, args) abort
-  let bin_path = go#path#CheckBinPath("gocode")
+  let l:gocode_bin = "gocode"
+  let l:gomod = go#util#gomod()
+  if filereadable(l:gomod)
+    let l:gocode_bin = "gocode-gomod"
+  endif
+
+  let bin_path = go#path#CheckBinPath(l:gocode_bin)
   if empty(bin_path)
     return []
   endif
--- a/autoload/go/complete.vim
+++ b/autoload/go/complete.vim
@@ -2,6 +2,15 @@ function! s:gocodeCommand(cmd, args) abort
   let l:gocode_bin = "gocode"
   let l:gomod = go#util#gomod()
   if filereadable(l:gomod)
+    " Save the file when in module mode so that go list can read the
+    " imports. If the user doesn't have autowrite or autorwriteall enabled,
+    " they'll need to write the file manually to get reliable results.
+    " See https://github.com/fatih/vim-go/pull/1988#issuecomment-428576989.
+    "
+    " TODO(bc): don't save the file when in module mode once
+    " golang.org/x/tools/go/packages has support for an overlay and it's used
+    " by gocode.
+    call go#cmd#autowrite()
     let l:gocode_bin = "gocode-gomod"
   endif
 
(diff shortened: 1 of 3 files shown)


fatih:
@bhcleek I'm out of town for one week, I'll try to debug it once I'm back. But if you have some spare times, you can also try it out (there is not much code). I posted the instructions. I think it might be due to the Unix socket name, which conflicts when `gocode` and `gocode-gomod` runs at the same time? Not sure but could be worth looking at it.

stamblerre:
@fatih, I'm also happy to take a look at this. It should be pretty similar to the change I made for VSCode. Thank you for starting it off!

stamblerre:
@fatih, I started looking at this, and it seems to be working ok for me. I don't use Vim often, so I may be making a mistake, but I basically tested it out by starting "gocode -s -debug" and "gocode-gomod -s -debug" and then coding in Vim in one $GOPATH directory and a different directory with a go.mod file. It seemed like the completions were being directed to the right version of gocode in both cases.

The only thing I did notice was that to trigger gocode I had to enter the "C-x C-o" keybinding, whereas there were a few times when autocompletions just popped up on my screen. I would attribute this to some incorrect Vim setting on my part, but it did seem to be ~20% of the time those completions appeared, so I wonder if that's related?

bhcleek:
@fatih, overall this seems to work really well. One other thing to note, though: it's probably a good idea to also disable `g:go_gocode_propose_source` while testing to be sure that `-source` isn't being used to find matches.

Like you and @stamblerre , I did see some instances where autocompletion couldn't find any candidates. It seemed to miss when it all exported identifiers from a package should have been returned (e.g. trying to autocomplete `flag.`), but it wasn't anywhere near 20% misses for me, and once it started completing identifiers from a package where it had previously missed, I didn't notice anymore misses from that package.

@stamblerre you should always have to use `C-x C-o` to autocomplete with vim-go. If autocompletion is popping up without that, then either something else is doing it or it's just taking a while for it to pop-up from a previous attempt.

For any readers that may be concerned about running `gocode` and `gocode-mod` simultaneously: @stamblerre's fork use a different unix domain socket than mdempsky's thanks to https://github.com/stamblerre/gocode/commit/b960ec24cf795e8590198eb337c6286d5488e399, so it should be safe to run both `gocode` and `gocode-mod` simultaneously.

fatih:
Paul Jolly shared the following notes:

```
The one thing to point out with respect to gocode in module mode is that files will have to be saved to disk before completions can work reliably

(or jumping to definitions for that matter)
Reason being, go list needs to see the file for the imports etc to resolve
This is only a short term issue, but it might be worthwhile having the jump-to-definition and completion shortcuts automatically save a file in the meantime (in module mode)
```

bhcleek:
Did Paul provide any expectations for how long `short term` will be?

myitcv:
> Did Paul provide any expectations for how long short term will be?

No specific timelines from the tools team for now, no.

Context: https://docs.google.com/document/d/1oEknhf60Cdg9p_i17ESIm3zjTuVK7Adr-lTw78D0Qrc/edit#bookmark=id.ko11cb4rz28i

bhcleek:
Thanks, @myitcv. Fatih is going to put us in touch next week, and I'll be sure to have reviewed the docs by then.

PR #1990, opened 2018-09-22
Fix missing extension for Windows

blyoa:
The binaries installed via `:GoUpdateBinaries` cannot be executable on Windows since the binaries do not have an extension.

This pull-request adds the ".exe" extension to the binary file name.


bhcleek:
Thank you.

lgtm

PR #1993, opened 2018-09-23
Fix highlighting version on gomod

rhysd:
Hi,

I started using Go modules with Go 1.11. And I found some highlights are not correct as follows:

- Version may have `+incompatible` suffix. I'm not sure that other words than 'incompatible' are used as suffix. For safety, I hard-coded `+incompatible` in the highlight definition.
- Comment after version is highlighted as part of version

I took screenshot of my `go.mod` before/after this fix.

- Before

before

- After

after


bhcleek:
lgtm

PR #1995, opened 2018-09-24
match gomod versions correctly

bhcleek:
* don't match +incompatible when major version is < 2.
* match pseudo versions correctly.

### DIFF
--- a/autoload/go/highlight_test.vim
+++ b/autoload/go/highlight_test.vim
@@ -1,0 +1,95 @@
+function! Test_gomodVersion_highlight() abort
+  try
+    syntax on
+
+    let l:dir= gotest#write_file('gomodtest/go.mod', [
+          \ 'module github.com/fatih/vim-go',
+          \ '',
+          \ '\x1frequire (',
+          \ '\tversion/simple v1.0.0',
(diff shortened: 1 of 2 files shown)


Carpetsmoker:
I fixed the semver match with this a while ago:

    " Semver as 'v1.1.1' and versions as 'v.0.0.0-date-commit'.
    syn match gomodVersion /\vv\d+\.\d+\.\d+(-\d{14}-[0-9a-f]{12})?/

Which seems to work well; but I didn't extensively test.

I'd strongly recommend adding `\v` and/or splitting stuff over multiple lines btw. It's going to be a major PITA to maintain like this.

Carpetsmoker:
Also, I had been thinking of re-doing this entire approach, as we know that everything after the first space and before `//` is going to be the version. I hadn't read the full spec (if any?) but I think all this exact version matching isn't really needed.

bhcleek:
`syn match gomodVersion /\vv\d+\.\d+\.\d+(-\d{14}-[0-9a-f]{12})?/` will not catch all the versions.

> we know that everything after the first space and before // is going to be the version

I disagree. I did this explicitly _because_ that's not the rule; vim-go should not highlight versions that don't match the spec.

PR #1998, opened 2018-09-24
Semver regex

wolf99:
Per recommended regex semver/semver#460

I'm not up on vim script regex support so this will likely require testing by reviewers (the regex itself is good, see the above linked PR and also semver/semver#232)


Carpetsmoker:
Also see #1995.

Personally I would consider a 168-character regexp without any comments or newlines to be unmaintainable btw.

bhcleek:

go.mod versions are semvers, but not all semvers are valid versions for go.mod. This attempts to cast the net too wide.

PR #2006, opened 2018-10-03
preserve shellcmdflag and normalize for /bin/sh

mirza-s:
Fix shell related problems when `shellcmdflag` contains flags not understood by `/bin/sh` by resetting it to a safe value and restoring it at the end.

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -137,12 +137,13 @@ endfunction
 " so that we always use a standard POSIX-compatible Bourne shell (and not e.g.
 " csh, fish, etc.) See #988 and #1276.
 function! s:system(cmd, ...) abort
-  " Preserve original shell and shellredir values
+  " Preserve original shell, shellredir and shellcmdflag values
   let l:shell = &shell
   let l:shellredir = &shellredir
+  let l:shellcmdflag = &shellcmdflag
 
   if !go#util#IsWin() && executable('/bin/sh')
-      set shell=/bin/sh shellredir=>%s\ 2>&1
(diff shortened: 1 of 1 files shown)


mirza-s:
Related https://github.com/fatih/vim-go/pull/988

bhcleek:
Can you give a concrete example of what this fixes for you?

mirza-s:
On Ubuntu 18.04 if you set `shellcmdflag=-lc` it breaks all operations that use the shell such as gofmt.
So you can't even save *.go files because gofmt is called before each save.

```
Error detected while processing function <SNR>24_fmt_autosave[3]..go#fmt#Format[41]..go#fmt#run[5]..go#util#Exec[17]..<SNR>41_exec[7]..<SNR>41_system:
line   10:
E484: Can't open file /tmp/vcxetHD/1
Error detected while processing function <SNR>24_fmt_autosave:
line    3:
E171: Missing :endif
```

This is because /bin/sh does not understand -lc (on Ubuntu `sh` is linked to `dash`)

```
$ sh -lc echo foo
foo: 66: foo: Syntax error: "(" unexpected
```

This does not happen on OSX because `sh` is actually `bash` which understands the provided `shellcmdflag`

So we reset shellcmdflag to -c which is understood by `sh` and restore it after invocation.

bhcleek:
Given that vim-go does not set `shellcmdflag` at all, it sounds like this is better taken care of in your vimrc than in vim-go.

mirza-s:
But it replaces the shell that uses shellcmdflag.

bhcleek:

mirza-s:
Thank you.

PR #2013, opened 2018-10-09
debug: Fix dlv call

jcvenegas:
delve has new flag to define the output[1].
For recent delve installed vim-go fails.

[1] https://github.com/derekparker/delve/pull/1192


jcvenegas:
Duplicated of https://github.com/fatih/vim-go/pull/1992/files

PR #2014, opened 2018-10-09
Use guibg instead of ctermbg if user is using neovim 

railsdev:
If it is neovim, goStatusLineColor must be using guibg and guifg to reflect correct colors. I am not expert in vim script. Hence please let me know if we can refactor this code.

### DIFF
--- a/autoload/go/statusline.vim
+++ b/autoload/go/statusline.vim
@@ -53,22 +53,12 @@ function! go#statusline#Show() abort
 
   " only update highlight if status has changed.
   if status_text != s:last_status
-    if (has("nvim"))
-      if status.state =~ "success" || status.state =~ "finished" || status.state =~ "pass"
-        hi goStatusLineColor cterm=bold guibg=#5fd700 guifg=#005f00
-      elseif status.state =~ "started" || status.state =~ "analysing" || status.state =~ "compiling"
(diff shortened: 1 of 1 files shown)


autoload/go/statusline.vim
```diff
@@ -53,12 +53,22 @@ function! go#statusline#Show() abort
...
-      hi goStatusLineColor cterm=bold ctermbg=196 ctermfg=52
+    if (has("nvim"))
+      if status.state =~ "success" || status.state =~ "finished" || status.state =~ "pass"
+        hi goStatusLineColor cterm=bold guibg=#5fd700 guifg=#005f00
+      elseif status.state =~ "started" || status.state =~ "analysing" || status.state =~ "compiling"
+        hi goStatusLineColor cterm=bold guibg=#ff8700 guifg=#870000
+      elseif status.state =~ "failed"
+        hi goStatusLineColor cterm=bold guibg=#ff0000 guifg=#5f0000
+      endif
+    else
```
Carpetsmoker: You don't need an `if`/`else` here, I think? You can just add `guifg` and `guibg` to the `hi [..]` statements. That will also make it work for gVim and Vim with `termguicolors`.

railsdev: Thanks for your input. I will make that change

bhcleek:
lgtm

PR #2025, opened 2018-10-17
Debugger: optional highlighting

phux:
The default highlighting of `GoDebugCurrent` makes the current line unreadable with my color scheme.
Overriding it in my `init.vim` or `ftplugin/go.vim` did not work.

PR introduces an option to skip the highlighting in `syntax/go.vim`, so one can specify their preferred colors in their configuration via `let g:go_debug_default_highlighting = 0`.


bhcleek:
Thanks. I made a few adjustments and merged your change.

Carpetsmoker:
Are you sure this can't be overridden from an `autocmd`? e.g. the `colorscheme` one?

PR #2034, opened 2018-10-30
Make $PATH take precedence over go_bin_path

cespare:
We're using wrapper scripts at the front of $PATH to ensure that everyone is using a particular version of gofmt/goimports. I was surprised that these weren't picked up by vim-go. This change makes vim-go respect the user's preferences by using $PATH first and only falling back to go_bin_path if the requested binaries aren't in $PATH at all.

### DIFF
--- a/autoload/go/path.vim
+++ b/autoload/go/path.vim
@@ -155,7 +155,7 @@ function! go#path#CheckBinPath(binpath) abort
   if !empty(go_bin_path)
     " append our GOBIN and GOPATH paths and be sure they can be found there...
     " let us search in our GOBIN and GOPATH paths
-    let $PATH = go_bin_path . go#util#PathListSep() . $PATH
+    let $PATH = $PATH . go#util#PathListSep() . go_bin_path
   endif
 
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
I don't think we should do any frobbery with `$PATH` at all, and just rely on users setting a sane `$PATH`. No matter what you do, you're going to surprise *someone*.

cespare:
I'd be fine with that as well. It seems like the current code is made to handle the case where the user hasn't added their Go bin directory to their $PATH, which seems like it is going to cause them issues outside of vim anyway.

bhcleek:
Thank you for contributing.

Because of the way `:GoInstallBinaries` and `:GoUpdateBinaries` work, this change would break many users; they might suddenly be using different binaries than they were previously, and `:GoInstallBinaries` and `:GoUpdateBinaries` could not be used to install the binaries that would be used by vim-go.

Instead, I'd suggest that you simply set `g:go_bin_path` or `$GOBIN` to the directory that contains the files you want to use.

Issue #2040, opened 2018-11-02
Add some option to search $PATH before go_bin_path for tool binaries


cespare:
This is a feature request, not an issue.

## The problem

If you're working on a team, it's useful to have everyone using the exact same version of gofmt/goimports. For example, the Go 1.11 upgrade changed the default formatting; if everyone is invoking a standard version of goimports, that can be upgraded in one central place rather than everyone needing to update their tool version.

Our solution at my company is to use a wrapper script for goimports in a standard location. This script installs and runs a particular version of goimports. The script is in our monorepo so we can upgrade it for everyone all at once.

Other companies use different solutions but they amount to the same thing: ensure that everyone invokes a particular version of the tool that can be upgraded centrally. (For example, maybe the tool is a binary on a network mounted drive.)

This all conflicts with the way vim-go handles binaries. In vim-go, `go_bin_path` specifies a path for vim-go's tools; it is used *both* as an installation location *and* the location to search first when finding a binary to run. In the scenarios I described, this won't work: if `go_bin_path` is set to be the shared, central tool location, then `:GoInstallBinaries`/`:GoUpdateBinaries` will try to trample on the centralized tool; if `go_bin_path` is set to some other directory (such as its default of `$GOBIN` or `$HOME/go/bin`), then vim-go will use its own installed version of the tool instead of the central version.

## Solutions

I created #2034 which made $PATH take precedence over `go_bin_path` when searching for the tool to run, but @bhcleek didn't like that solution because it could confuse people who ended up accidentally getting a different tool version than the one vim-go installed.

A workaround I'm using now is to set `go_bin_path` to some nonexistent path; this completely breaks `:GoInstallBinaries`/`:GoUpdateBinaries`, but then at least I get the correct tool version from my $PATH.

 Another approach would be to introduce a new boolean option to let `$PATH` take precedence over `go_bin_path`. This would let people who know what they're doing set up a $PATH that makes sense while still allowing `go_bin_path` to be useful for other tools installed via `:GoInstallBinaries` which don't need to be versioned.

I'm thinking something like

```
                                                *'g:go_search_bin_path_first'*


This option lets |'g:go_bin_path'| (or its default value) take precedence over
$PATH when invoking a tool command such as |:GoFmt| or |:GoImports|.

Enabling this option ensures that the binaries installed via
|:GoInstallBinaries| and |:GoUpdateBinaries| are the same ones that are
invoked via the tool commands.

By default it is enabled.
>
[...]

PR #2041, opened 2018-11-03
Add the go_search_bin_path_first option

cespare:
This allows the user to let their $PATH take precedence over go_bin_path when looking for tools if they so choose (but defaults to keeping the current behavior). For context, see #2040.


bhcleek:
Thanks!

Issue #2047, opened 2018-11-07
:GoDeclsDir open correct file with wrong content.


konstantinzolotarev:
### What did you do? (required. The issue will be **closed** when not provided.)
Run `:GoDeclsDir` command and jumped to struct/function declaration.

### What did you expect to happen?
File with this struct/function declaration opened

### What happened instead?
File opened but with content:
```
package main

import "fmt"

func main() {
  fmt.Println("vim-go")
}
```

And for test packages:
```
package main

import "testing"

func TestHelloWorld(t *testing.T) {
  // t.Fatal("not implemented")
}
```

After `:edit` command (tried to refresh file content) same file with correct content opened in new buffer.
Switching back to buffer opened by `:GoDeclsDir` this message appears
```
:bnext
W13: Warning: File "pkg/mailer/sendgrid.go" has been created after editing started
[O]K, (L)oad File:
``` 

### Configuration (**MUST** fill this out):

* vim-go version:
 Same issue with `master` branch and `v1.19` and also `v1.18`

**But it works as expected in `v1.17` and less**

* `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):
https://gist.github.com/konstantinzolotarev/d4c218a864de503c7007428ad49aeb2d

* Vim version (first three lines from `:version`):
```
NVIM v0.3.1
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/local/Homebrew/Library/Homebrew/shims/mac/super/clang -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -W
extra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_
GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/tmp/neovim-20181019-20909-1hlaj6i/neovim-0.3.1/build/config -I/tm
p/neovim-20181019-20909-1hlaj6i/neovim-0.3.1/src -I/usr/local/include -I/usr/local/opt/gettext/include -I/usr/include -I/tmp/neovim-20181019-20909-1hlaj6i/neo
[...]

PR #2048, opened 2018-11-07
Fix for #2047

konstantinzolotarev:
Fixes https://github.com/fatih/vim-go/issues/2047

because if we use filename with path `fnamemodify(decl.filename, ":p:.")` (line 120) we don't need to `execute cd . s:current_dir` (line 37)


bhcleek:
Thank you @konstantinzolotarev

PR #2077, opened 2018-11-28
Incorporate the Go logo into doc header

nebloc:
Non functional change, just thought the hyphen looked like the [Go logo](https://blog.golang.org/go-brand) and decided to try incorporate it for fun.


bhcleek:
Thank you for the contribution. I took a look, but I think this change negatively impacts the readability fo the ascii art logo.

PR #2079, opened 2018-11-30
add go_gocode_enabled option

fannheyward:
I love vim-go and use it with development every day. Current I'm trying go language server for auto-completion and lint diagnostic etc. After I disabled vim-go from my workflow, I really missing commands like GoTest/GoImport, so I add `g:go_gocode_enabled` option to disable gocode completion. And now I can talk to language server for code completion/lint, and keep other vim-go useful stuff.

This small feature may be a very personal requirement. If it doesn't meet vim-go's goal, just discard it.

Thanks again for the great plugin!


fannheyward:
Thanks for your suggestion.

PR #2080, opened 2018-12-01
Add syntax highlight to 'godoc' filetype

rhysd:
I added syntax highlighting to `godoc` filetype.

This patch highlights from head of the buffer until the second blank line as `go` filetype.


rhysd:
Oh, I see. It looks not easy to get Go snippets in document.

Then I think I need to enable this highlight on `K` (or `:GoDoc` with no argument). I guess it's possible.

PR #2082, opened 2018-12-02
-fallback-to-source needs gocode to be running the cache importer

andyjack:
- therefore pass the -cache option to gocode command calls, so the
  client will start the server with the -cache option as well.
  Needs [this PR](https://github.com/mdempsky/gocode/pull/93) in gocode


bhcleek:
Thank you for contributing

PR #2088, opened 2018-12-09
Remove duplicated l:importPath

yaojingguo:
`l:importPath` has been added twice  in the following code. It should be added once.

```go
let l:build_cmd = ['go', 'build', '-o', go_bin_path . go#util#PathSep() . bin, l:importPath]
let [l:out, l:err] = go#util#Exec(l:build_cmd + [l:importPath])
```


bhcleek:
thank you

PR #2090, opened 2018-12-11
Restore &cpo even when finishing early

dhduvall:
If a script calls `finish`, the restoration of `&cpo`, which is at the
end of the script, never gets called.

As a follow-up to #2055, this change moves the place where `&cpo` is
preserved until after the places where `finish` is called.  I believe
this is safe, since those places are near the beginning of their
respective scripts, and there's nothing earlier that's dependent on the
default setting of `&cpo`.


dhduvall:
I don't have the tagbar plugin installed, so I was getting my `&cpo` reset to the vim defaults. This fixes that problem for me.

bhcleek:
thanks you

PR #2091, opened 2018-12-15
use modified buffer as input to gocode-gomod

bhcleek:
Saving a modified buffer to disk is no longer necessary for
autocompletion with github.com/stamblerre/gocode;
golang.org/x/tools/go/packages now supports overlays, and support for
them was merged to github.com/stamblerre/gocode recently.


myitcv:
Awesome news. Thanks

Issue #2096, opened 2018-12-17
Template creation and BufNewFile rules do not contemplate non-modifiable buffers


mixedCase:
### What did you do? (required. The issue will be **closed** when not provided.)

Tried to open a .go file in a non-modifiable buffer, specifically, vim-gina's Blame's.

### What did you expect to happen?

vim-go not to interfere.

### What happened instead?

vim-go attempted to set fileencoding, and when patching that out, attempts to instantiate the go template.

### Configuration (**MUST** fill this out):

* vim-go version: https://github.com/fatih/vim-go/commit/189676476a638364e76fce9998c37f7ac79f7056

* `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):

Assuming vim-plugged installed:
```
call plug#begin('~/.config/nvim/plugged')
Plug 'fatih/vim-go', { 'for': 'go' }
Plug 'lambdalisue/gina.vim'
call plug#end()
```
Open a .go file, run `:Gina blame`, first time works, try again, next time fails.

* Vim version (first three lines from `:version`):
```
NVIM v0.3.1
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wconversion -O2 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -Wno-array-bounds -D
INCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/build/neovim/src/build/config -I/build/neovim/src/neovim-0.3.1/src -I/usr/include -I/build/neovim/src/build/src/nvim/auto -I/build/neovim/src/build/include
```

* Go version (`go version`):

```go version go1.11.3 linux/amd64```

* Go environment (`go env`):

```
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/mixedcase/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/mixedcase/.go"
[...]

PR #2097, opened 2018-12-17
Fix breaking non-modifiable buffers in some cases

mixedCase:
Fixes fatih/vim-go#2096

### DIFF
--- a/autoload/go/template.vim
+++ b/autoload/go/template.vim
@@ -5,9 +5,6 @@ set cpo&vim
 let s:current_file = expand("<sfile>")
 
 function! go#template#create() abort
-  if !&modifiable
-    return
-  endif
   let l:go_template_use_pkg = go#config#TemplateUsePkg()
   let l:root_dir = fnamemodify(s:current_file, ':h:h:h')
 
--- a/autoload/go/template.vim
+++ b/autoload/go/template.vim
@@ -5,6 +5,9 @@ set cpo&vim
 let s:current_file = expand("<sfile>")
 
 function! go#template#create() abort
+  if !&modifiable
+    return
+  endif
   let l:go_template_use_pkg = go#config#TemplateUsePkg()
   let l:root_dir = fnamemodify(s:current_file, ':h:h:h')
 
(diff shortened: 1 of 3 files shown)


autoload/go/template.vim
```diff
@@ -5,6 +5,9 @@ set cpo&vim
 let s:current_file = expand("<sfile>")
 
 function! go#template#create() abort
+  if !&modifiable
+    return
+  endif
```
bhcleek: Given that this function does not write the file, why is this change necessary?

mixedCase: Added some clarification on related issue; but maybe the check should be somewhere else in the call stacK?

bhcleek: Instead of returning early here, it's may be better to return early in https://github.com/fatih/vim-go/blob/master/plugin/go.vim#L278-L283

ftdetect/gofiletype.vim
```diff
@@ -25,7 +25,7 @@ function! s:gofiletype_post()
 endfunction
 
 " Note: should not use augroup in ftdetect (see :help ftdetect)
-au BufNewFile *.go setfiletype go | setlocal fileencoding=utf-8 fileformat=unix
+au BufNewFile *.go setfiletype go | if &modifiable | setlocal fileencoding=utf-8 fileformat=unix | endif
```
bhcleek: As mentioned in #2096, the fileencoding and fileformat options probably shouldn't be set at all here, but this is an acceptable short term solution until I can refactor `ftdetect/gofiletype.vim`, `ftplugin/go.vim` and `plugin/go.vim`.

Would you mind adding a similar guard to the `au BufNewFile *.s` line for consistency until I can do the refactor?

mixedCase: Done

autoload/go/template.vim
mixedCase: Done. Opted for an additional check instead of early return since I felt it was still clear given the var name, but please let me know if you'd still prefer it as an early return.

PR #2099, opened 2018-12-19
Allow for custom template directory to be set

EVODelavega:
It can be useful to have project-specific `.vimrc` files, which set up different templates to use. Rather than having to keep them all in my `~/.vim/plugged/vim-go/templates` directory, I'd quite like to be able to keep the template path alongside my custom `.vimrc` files. for that reason, I've added a `g:go_template_dir` variable that defaults to the current templates directory, so it shouldn't break anything, but I can use `g:go_template_dir` and point to the correct templates easily. Quite a simple change, but IMHO, one worth adding.

Haven't tested this yet, but wanted to get some feedback on what the best approach would be in case of a missing template file. I see 3 options:

* Copy over from the default templates dir (probably adding `go#config#DefaultTemplateDir()` to get it
* Don't handle it differently (i.e. PR as it is)
* Add an additional var like `g:go_template_dir_init` (default to 0), that copies over the default templates from the templates dir should they not exist. 

The third option might yield the best UX, but it's a bit too handhold-y IMO. Pretty soon, we'd end up checking the custom `g:go_template_file`, in the custom dir, and if it doesn't exist, we could end up copying over the default `hello_world.go` to the custom directory, and storing it with a new name. I'm more of the mindset that if people don't understand that the files are expected to exist, they shouldn't be messing with those vars in the first place, but I'm open to suggestions, of course.

Should this PR be accepted, I'll squash all the _"fix typo"_ and _"resolve issue"_ commits, of course

### DIFF
--- a/autoload/go/template.vim
+++ b/autoload/go/template.vim
@@ -24,7 +24,13 @@ function! go#template#create() abort
       else
         let l:template_file = go#config#TemplateFile()
       endif
-      let l:template_path = go#util#Join(l:root_dir, "templates", l:template_file)
+
+      " If template_file is an absolute path, use it as-is. This is to support
+      " overrides pointing to templates outside of the vim-go plugin dir
+      if fnamemodify(l:template_file, ':p') != l:template_file
+        let l:template_path = go#util#Join(l:root_dir, "templates", l:template_file)
+      endif
+
       silent exe 'keepalt 0r ' . fnameescape(l:template_path)
     endif
   else
--- a/autoload/go/template.vim
+++ b/autoload/go/template.vim
@@ -9,6 +9,7 @@ function! go#template#create() abort
   let l:root_dir = fnamemodify(s:current_file, ':h:h:h')
 
   let l:package_name = go#tool#PackageName()
+  let l:filename = expand('%:t')
 
   " if we can't figure out any package name (i.e. no Go files in the directory)
   " from the directory create the template or use the directory as the name.
@@ -18,7 +19,6 @@ function! go#template#create() abort
       let l:content = printf("package %s", l:path)
       call append(0, l:content)
     else
-      let l:filename = expand('%:t')
       if l:filename =~ "_test.go$"
         let l:template_file = go#config#TemplateTestFile()
       else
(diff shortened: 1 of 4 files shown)


bhcleek:
Thank you for contributing. 😄 

This approach was submitted once before, but the feedback wasn't implemented. Would you be willing to implement it as suggested in https://github.com/fatih/vim-go/pull/1746#issuecomment-378750382 ?

EVODelavega:
@bhcleek OMW, down to the name of the config var :smile:, didn't know that'd been submitted before. I can implement it as suggested for sure, however implementing it like that might be a breaking change, no?
If someone has set the existing vars to a custom file, and that file were to be interpreted as an absolute path, their setup wouldn't work.

Suppose they've added subdirectories in the templates directory, to keep things organised, and have `g:go_template_file = "foo/bar/main.go"` in there. There's no reliable way to distinguish that from an absolute path (thinking about across different OS's here). On top of that, having the path be a separate variable does make it easier to automate swapping templates with a bit of scripting in your vimrc, and (for example) depending on the path, setting the `g:go_template_dir` variable to a different path.

If you disagree, I'll make the requested changes tomorrow, but just thought I'd share my thoughts on the comment you linked.

bhcleek:
You raise some good points. I think, though, that we can deal with this simply enough: `fnamemodify()` can tell you whether a file path is absolute or not. e.g.:
```
diff --git a/autoload/go/template.vim b/autoload/go/template.vim
index 3f5e538..c4b1a44 100644
--- a/autoload/go/template.vim
+++ b/autoload/go/template.vim
@@ -24,8 +24,12 @@ function! go#template#create() abort
       else
         let l:template_file = go#config#TemplateFile()
       endif
-      let l:template_path = go#util#Join(l:root_dir, "templates", l:template_file)
-      silent exe 'keepalt 0r ' . fnameescape(l:template_path)
+
+      if fnamemodify(l:template_file, ':p') != l:template_file
+        let l:template_file = go#util#Join(l:root_dir, "templates", l:template_file)
+      endif
+
+      silent exe 'keepalt 0r ' . fnameescape(l:template_file)
     endif
   else
     let l:content = printf("package %s", l:package_name)
```

EVODelavega:
Right-o, forgot about `:p` on `fnamemodify`, I'll just use `fnamemodify` and `expand`, I'll set about making the changes somewhere today (I'm flying out shortly, so a couple of hours waiting at the terminal should do the trick, if I'm not too tired).

EVODelavega:
@bhcleek Done. I've pretty much implemented exactly what your suggested patch was. I did think about moving the `fnamemodify` calls to the config file. Given that getters should just return the value as-is, I decided against it, though.

autoload/go/template.vim
```diff
@@ -9,6 +9,7 @@ function! go#template#create() abort
   let l:root_dir = fnamemodify(s:current_file, ':h:h:h')
 
   let l:package_name = go#tool#PackageName()
+  let l:filename = expand('%:t')
```
bhcleek: There's no need to move this; it's only used; keep its declaration close to where it's used aids in readability.

bhcleek:
Thanks!

Merged via #2434

PR #2104, opened 2018-12-22
Changed templates

UlisseMini:
I've changed the template contents, when i started using the plugin it annoyed me to no end (until i figured out how to change it)<br>
i've also changed the names of the files, i think `default.go` is a better name then `hello_world.go`
My changes here are for making vim-go work better without configuration, using `hello_world.go` kind of forced people into changing the templates whereas with more generic templates they might never need to :)

EDIT: Looks like travis does not like my changes, i don't see how i could have screwed anything up since all i did was change the file names and contents


bhcleek:

Travis is failing, because there are tests for the templates.

Issue #2063, opened 2018-11-18
support modules in go#package#FromPath


bhcleek:
`go#package#FromPath` should be module-aware

bhcleek:
`go#package#FromPath` uses `go list`. `go list` supports modules. Therefore, nothing needs to be done for this.

PR #2113, opened 2019-01-02
Add staticcheck/GoStaticCheck command

EVODelavega:
I've been enjoying the staticcheck tool because of its completeness as a linter, and the additional suggestions it makes (e.g. unused global variables and types, which aren't uncommon in my tests). Because `:GoErrCheck` has the `-abspath` flag hard-coded, it's not possible to change the `g:go_errcheck_bin` variable, so I decided to take a stab at adding the `:GoStaticCheck` command.

Most of the code is a straight copy-paste from `go#lint#Errcheck`, though I thought it best not to add the function in the `lint` file. To me `staticcheck` is something I use not as a linter. It's more of a brush used to polish code a bit more. Anyway, depending on the feedback I'll see what I can do to add support for this tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ This plugin adds Go language support for Vim, with the following main features:
   quickfix or location list.
 * Lint your code with `:GoLint`, run your code through `:GoVet` to catch static
   errors, or make sure errors are checked with `:GoErrCheck`.
+* In addition to linters, run `:GoStaticCheck` to get some suggestions to improve
+  overall code quality
 * Advanced source analysis tools utilizing `guru`, such as `:GoImplements`,
   `:GoCallees`, and `:GoReferrers`.
 * ... and many more! Please see [doc/vim-go.txt](doc/vim-go.txt) for more
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -258,6 +258,10 @@ function! go#config#ErrcheckBin() abort
   return get(g:, "go_errcheck_bin", "errcheck")
 endfunction
 
+function! go#config#StaticcheckBin() abort
+  return get(g:, "go_staticcheck_bin", "staticcheck")
+endfunction
+
 function! go#config#MetalinterAutosave() abort
   return get(g:, "go_metalinter_autosave", 0)
 endfunction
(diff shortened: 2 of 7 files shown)


bhcleek:
Thanks for another contribution!

vim-go will already leverage `staticcheck` via `gometalinter` (e.g. `:GoMetaLinter`) when `g:go_metalinter_enabled` includes `staticcheck`. Do you have a practical justification for why `staticcheck` should be runnable via vim-go on its own while some of the other linters that `gometalinter` can run that aren't enabled by default should not be?

fatih:
Just want to chime in. I believe staticcheck is a full blown linter with many options and knobs, and therefore is better suited for CI/CD systems. I think we need to think about if this makes sense to include such a tool :)

EVODelavega:
@bhcleek The main reason for my wanting to begin work on the integration of `staticcheck` is because `gometalinter` does support it, but not by default. By default it relies on a string of tools that have been deprecated and incorporated into the main `staticcheck` tool.

With that said, @fatih  does raise a good point. Because of the tool being so feature complete, running that tool any time you lint a project, it becomes cumbersome and slow. That's the main reason why I decided to create a dedicated command for it. Allowing me to run something like `:GoStaticCheck github.com/foo/project/internal/...`. It's true, though, that because of the scope it's probably best to give this some more thought, and see whether or not there's actually a demand for a feature like this.

For the time being, it's quite easy for me to just do something like this:

```
:new
:read !staticcheck github.com/foo/pkg/internal/...
```

Following the mantra _"Copy with pride"_, I can take a similar approach as `:GoErrCheck` to parse the output and use quickfix independently of this plugin.

Feel free to close this PR, or suggest some changes and I'd be more than happy to work on them. I've been an avid user/evangelist of this plugin, and I'm eager to contribute, despite my being a vimL beginner.

bhcleek:
vim-go will already support staticcheck simply by setting the right option: `let g:go_gometalinter_enable=['staticcheck']`. From what I can tell, that should be completely sufficient. One could also just run `staticcheck` _instead_ of `gometalinter` by using `let g:go_metalinter_command='staticcheck'`.

> Feel free to close this PR, or suggest some changes and I'd be more than happy to work on them. I've been an avid user/evangelist of this plugin, and I'm eager to contribute, despite my being a vimL beginner.

Thank you for the offer. Your help would be greatly appreciated. https://github.com/fatih/vim-go/issues/2063 is a good place to start. I'm also working on lsp integration. Once the foundation is in place, any help you could provide maturing it would be great.

PR #2114, opened 2019-01-02
Prefer the system PATH for some commands

euank:
The default $PATH is a better place to look than $GOPATH/bin for some
tools. This is especially true for 'go', 'diff', and 'uname'.

If someone incidentally creates a go binary named 'go' or the like, but
has configured their $PATH to prefer the system 'go' binary to the one
in their $GOPATH, it seems unexpected to go ahead and run it anyway.

For the tools that are expected to be in the $GOPATH/bin folder, like
'goimports', it does make sense to just go ahead and check that folder
first.

-------------------

In case you're curious what motivates this change, I'll briefly explain the situation I ran into. I created some language benchmarks, which meant I had folders named `lambda-bench/rust` and [`lambda-bench/go`](https://github.com/euank/lambda-bench/tree/benchmark-2018-09/go) and the like.

Unfortunately, this meant my go package import path ended in `go`, so when I typed `:GoInstall` it of course ended up as a binary named `$GOPATH/bin/go`. After doing this, `vim-go` basically would fall over since it would execute the wrong go binary.

My `$PATH` does include `$GOROOT/bin` well before `$GOPATH/bin`, but that didn't help since `vim-go` goes out of its way to prefer `$GOPATH`.

I'm sure you could say "just don't name a folder that contains go code 'go' ever", and that might also be a reasonable outcome... but it seems a little nicer to me to remove that footgun entirely.

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -175,7 +175,7 @@ function! go#tool#ExecuteInDir(cmd) abort
   let dir = getcwd()
   try
     execute cd . fnameescape(expand("%:p:h"))
-    let [l:out, l:err] = go#util#Exec(a:cmd)
+    let [l:out, l:err] = go#util#ExecSystem(a:cmd)
   finally
     execute cd . fnameescape(l:dir)
   endtry
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -55,7 +55,7 @@ endfunction
 function! go#doc#Open(newmode, mode, ...) abort
   " With argument: run "godoc [arg]".
   if len(a:000)
-    let [l:out, l:err] = go#util#Exec(['go', 'doc'] + a:000)
+    let [l:out, l:err] = go#util#ExecSystem(['go', 'doc'] + a:000)
   else " Without argument: run gogetdoc on cursor position.
     let [l:out, l:err] = s:gogetdoc(0)
     if out == -1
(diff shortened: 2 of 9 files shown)


bhcleek:
Thank you for contributing again.

From your description, it sounds like `let g:go_search_bin_path_first=0` would satisfy your use case without requiring additional changes to vim-go.

euank:
Yeah, while digging into this I did notice that toggle.

It would work for me yes, but I don't think it's ideal. I would still prefer `$GOPATH/bin` to be used for things like `golint` and `goimports`, even if I had those binaries somewhere earlier in my normal path

The other reason I felt a pull request might be appropriate is because I think this change provides slightly better defaults.

It's possible the scenario I described above is enough of an edgecase that the bit of extra complexity it adds to vim-go isn't worth it, and I totally understand if that's the consensus.

autoload/go/tool.vim
```diff
@@ -175,7 +175,7 @@ function! go#tool#ExecuteInDir(cmd) abort
   let dir = getcwd()
   try
     execute cd . fnameescape(expand("%:p:h"))
-    let [l:out, l:err] = go#util#Exec(a:cmd)
+    let [l:out, l:err] = go#util#ExecSystem(a:cmd)
```
bhcleek: `ExecuteInDir` is used for other tools, too, and shouldn't be changed to use `ExecSystem` instead of `Exec`.

euank:
Fair enough, thanks for the prompt response 👍

PR #2116, opened 2019-01-05
Fix highlighting for function return types when highlight_function_arguments is enabled

pocke:
# Problem

vim-go does not highlight function return types with brackets correctly when `g:go_highlight_function_arguments = 1` is setted.

## actual highlight

vim-go does not highlight the returned types of `b` function, which are `string` and `error`.
It should be highlighted as `goType` syntax group (green color in this color scheme), but they actually have `goSimpleArguments` syntax group.

note: If `let g:go_highlight_function_arguments = 0`, it is highlighted below.

## configuration to reproduce

`test.vim`. Need to replace `/path/to/vim-go/` with the actual path.

```vim
let g:go_highlight_function_arguments = 1
set nocompatible
set rtp^=/path/to/vim-go/
syntax on
```

`test.go`

```go
package main

type T struct{}

func a() error {
	return nil
}
func b(x int) (string, error) {
	return "", nil
}
func c(y bool, z T) (str string, err error) {
	return
}
```

```bash
$ vim --clean -u test.vim test.go
```

# cause

`goSImpleArtuments` syntax matching definition has a problem.
It matches a function argument (e.g. `(a int)`, `(a int, b bool)`).

it specifys `nextgroup=goSimpleArguments`, it means the group repeats itself. Return types with bracktes are concealed by this `nextgroup` definition.

```
func b(x int) (string, error) {
  //  ^^^^^^^                 goSimpleArguments
  //          ^^^^^^^^^^^^^^^ repeated goSimpleArguments
```

I'm not sure why the nextgroup is specified. I think it's unnecessary because I guess go syntax does not allow an argument list after an argument list.
If someone knows the reasons of the nextgroup, please tell me it :pray:

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -270,7 +270,7 @@ if go#config#HighlightFunctions() || go#config#HighlightFunctionArguments()
   syn match goFunction          /\w\+/ nextgroup=goSimpleArguments contained skipwhite skipnl
   syn match goReceiverType      /\w\+/ contained
 if go#config#HighlightFunctionArguments()
-  syn match goSimpleArguments   /(\(\w\|\_s\|[*\.\[\],\{\}<>-]\)*)/ contained contains=goArgumentName nextgroup=goSimpleArguments skipwhite skipnl
+  syn match goSimpleArguments   /(\(\w\|\_s\|[*\.\[\],\{\}<>-]\)*)/ contained contains=goArgumentName skipwhite skipnl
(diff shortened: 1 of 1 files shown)


bhcleek:
Thank you for contributing 🙇 

>  I think it's unnecessary because I guess go syntax does not allow an argument list after an argument list

Without this, named arguments may not be highlighted correctly. ref: https://github.com/fatih/vim-go/pull/1587 

Before this PR:
image

This PR:
image

bhcleek:
I've come up with an additional fix to address the named return value situation; you can leave this PR as-is, and I'll build on it.

Thanks again for contributing!

Issue #632, opened 2015-11-27
Remote editing wipes out .go files with go_fmt_autosave enabled.


mk23:
Bisecting revealed this commit to be the culprit:  54946eafacdb6721cad4fd50b803580445d3eb68

I use vim as my IDE and often use netrw(scp) for editing files on remote servers.  I recently updated vim-go bundle and now every time I write the buffer out (ZZ or :w) the resulting file on the remote is completely wiped out.  This only happens with auto gofmt is enabled.
##### Prerequisites
1. any machine with ssh+key auth
2. vim-go installed with `g:go_fmt_autosave = 1`
3. gofmt or goimports doesn't make a difference
##### Steps
1. $ `vim scp://localhost/foo.go`
2. enter some content, i.e. `package main`
3. save buffer (:w)
##### Observed
- The resulting document is blank
##### Expected
- My code, that I worked so hard on, must continue exist but with better formatting. :grin:

guns:
Didn't think about netrw with that commit! Looks like this can be hacked around by writing to `b:netrw_tmpfile`. I'll take a stab at it now.

fatih:
That's an interesting issue :) @guns are you're working on it? If not let me know please.

guns:
@fatih: Yes. I have found a couple of options, but I'm afraid I won't have time to write a solid patch until tomorrow.

fatih:
@guns no worries! Take your time. I just wanted to be sure if you're working on it :) Thanks on tackling this problem :+1:

codeactual:
Just in case it's helpful for the patch, I also ran into revert/wipe symptoms in local file scenarios.

#### My setup

- `vim 7.4.052-1ubuntu3`
- `vim-go` settings are default except for `g:go_highlight_*` are all enabled
- `syntastic` bundle removed entirely just for testing this

#### Scenarios

These appear 100% reproducible for me in the cited 54946ea (but not in the prior commit) and current head (4de4eeb).

##### New file

Open a new file and add some code. A save (`:w`) or format (`:GoFmt`) operation will empty the buffer.

##### Existing file

Open an existing file that only contains "hi" and add some code. `:w`/`:GoFmt` will revert the buffer back to "hi". "hi" can also be replaced by pre-`gofmt`-ed and compilable code.

#### Caveats

- Errors prevent it. If `gofmt` finds an error like `expected 'package', found 'func'`, nothing is wiped/reverted erroneously.
[...]

PR #2132, opened 2019-01-26
Prevent go fmt autosave from attempting to write to fugitive buffers

washtubs:
Fixes [this issue](https://github.com/tpope/vim-fugitive/issues/1128#issuecomment-446087055) from the vim-go side. Not sure if this is the way to go, but it works for me.

### DIFF
--- a/autoload/go/auto.vim
+++ b/autoload/go/auto.vim
@@ -47,7 +47,9 @@ endfunction
 function! go#auto#fmt_autosave()
   " Go code formatting on save
   if get(g:, "go_fmt_autosave", 1)
-    call go#fmt#Format(-1)
+    if filereadable(expand("%@"))
+      call go#fmt#Format(-1)
+    endif
   endif
 endfunction
 


bhcleek:
Thank you for the contribution. 🙇 

It's preferable to add a more general solution instead of adding a special case for one other plugin. There was a previous attempt to with vim-fugitive, #1516, that we ended up rejecting for similar reasons. The correct fix in that case was #1527, and it shows the way for skipping formatting when writing a fugitive buffer, too. A solution for this that's similar to the fix in #1527 would be welcome.

Alternatively (and perhaps a better solution), change https://github.com/fatih/vim-go/blob/master/autoload/go/fmt.vim#L108 to read in the formatted temp files contents into the current buffer and then write the buffer so that fugitive can write it correctly when the buffer doesn't actually exist (check it with `filereadable(a:target)`).

Solving this will also likely solve #632.

Are you willing to implement one of the solutions I proposed here?

washtubs:
Using `filereadable` makes *way* more sense. TY. I will amend my PR.

washtubs:
I think I may have glossed over your last paragraph and just honed in on checking `filereadable`.

I think bailing on auto-save as I have it is the way to go at least for these special fugitive buffers. (Not sure what other special cases the `filereadable` check may capture). Anyways I would find that behavior pretty unexpected, and unwanted even because when I write to staging, I would want to ensure that the changes I'm making are very targeted. Formatting the file could cause changes to be staged that I didn't intend for that commit.

bhcleek:
Thank you for the contribution. This change is included in #2143.

PR #2141, opened 2019-02-10
fix error at go#def#jump_to_declaration when vim-go uses godef

dzruyk:
Godef outputs only one line when user tries to jump to module definition,
because of that vim-go shows following error:

```
Error detected while processing function go#def#Jump[70]..go#def#jump_to_declaration:
line   17:
E684: list index out of range: 1
E15: Invalid expression: parts[1]
```

This PR fixes the issue and adds few additional checks.


bhcleek:
Thank you

PR #2148, opened 2019-02-22
Add '_internal_test.go' to alternate files

unkiwii:
Usually tests written in `_test.go` files are in another package so the tests can be written as a user of the package, but sometimes when we need to test package internals then the suggested way is to separate those tests in a file ending with `_internal_test.go` and leave them in the same package so the internals can be tested.

I added the possibility to alternate between a file and it's internal tests the following way:

| Files you have | Current file | Current file after go#alternate#Switch|
| --- | --- | --- |
| x.go, x_test.go, x_internal_test.go | x.go | x_test.go |
| x.go, x_test.go, x_internal_test.go | x_test.go | x_internal_test.go |
| x.go, x_test.go, x_internal_test.go | x_internal_test.go | x.go |
| x.go, x_test.go | x.go | x_test.go |
| x.go, x_test.go | x_test.go | x.go |
| x.go, x_internal_test.go | x.go | x_internal_test.go |
| x.go, x_internal_test.go | x_internal_test.go | x.go |
| x_test.go, x_internal_test.go | x_test.go | x_internal_test.go |
| x_test.go, x_internal_test.go | x_internal_test.go | x_test.go |
| x.go | x.go | [error] |
| x_test.go | x_test.go | [error] |
| x_internal_test.go | x_internal_test.go | [error] |

### DIFF
--- a/autoload/go/alternate.vim
+++ b/autoload/go/alternate.vim
@@ -8,24 +8,32 @@ function! go#alternate#Switch(bang, cmd) abort
   if empty(file)
     call go#util#EchoError("no buffer name")
     return
+  elseif file =~# '^\f\+_internal_test\.go$'
+    let l:root = split(file, '_internal_test.go$')[0]
+    let l:alt_files = [l:root . ".go", l:root . '_test.go']
   elseif file =~# '^\f\+_test\.go$'
     let l:root = split(file, '_test.go$')[0]
-    let l:alt_file = l:root . ".go"
+    let l:alt_files = [l:root . "_internal_test.go", l:root . '.go']
   elseif file =~# '^\f\+\.go$'
     let l:root = split(file, ".go$")[0]
-    let l:alt_file = l:root . '_test.go'
+    let l:alt_files = [l:root . '_test.go', l:root . '_internal_test.go']
   else
     call go#util#EchoError("not a go file")
     return
   endif
-  if !filereadable(alt_file) && !bufexists(alt_file) && !a:bang
-    call go#util#EchoError("couldn't find ".alt_file)
-    return
-  elseif empty(a:cmd)
-    execute ":" . go#config#AlternateMode() . " " . alt_file
-  else
-    execute ":" . a:cmd . " " . alt_file
-  endif
+  let l:alt_file = ""
+  for alt_file in l:alt_files
+    if !filereadable(alt_file) && !bufexists(alt_file) && !a:bang
+      continue
+    elseif empty(a:cmd)
+      execute ":" . go#config#AlternateMode() . " " . alt_file
+      return
+    else
+      execute ":" . a:cmd . " " . alt_file
+      return
+    endif
+  endfor
+  call go#util#EchoError("couldn't find " . join(l:alt_files, " or "))
 endfunction
 
 " restore Vi compatibility settings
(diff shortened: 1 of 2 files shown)


bhcleek:
Where is that suggested?

unkiwii:
> > but sometimes when we need to test package internals then the suggested way
> 
> Where is that suggested?

Is not official, but I see it (and use it) in several projects, I will left here some examples:

> If you do need to unit test some internals, create another file with _internal_test.go as the suffix. Internal tests will necessarily be more brittle than your interface tests — but they’re a great way to ensure internal components are behaving, and are especially useful if you do test-driven development.
> https://stackoverflow.com/questions/32180341/go-test-internal-functions

> If a test must touch the internals of some package foo, that test can be in package foo, but must keep its dependencies to a minimum, and must have a name that ends with _internal_test.go.
> https://vanadium.github.io/community/coding-guidelines.html

> _internal_test.go
上面的package_test是用来测试对外可见的方法。如果，我们要测试内部方法，就创建*_internal_test.go这样的文件。使用同一个package。这种对于tdd，确保每个方法的测试很有用。
可以参考标准库 time包里的internal_test.go
对了，go build的时候 不会编译以_test.go结尾的文件，只有go test的时候才会编译。
> https://fedomn.bitcron.com/articles/backend/go-testing#toc_1

Translated from chinese:
> _internal_test.go
The above package_test is used to test externally visible methods. If we want to test the internal methods, we create a file like *_internal_test.go. Use the same package. This is useful for tdd to ensure that each method is tested.
Can refer to internal_test.go in the standard library time package
By the way, go build does not compile files ending with _test.go, only compile when go test.

> [13:07] <perrito666> what is the difference between _test.go and _whitebox_test.go ?
[13:08] <fwereade> perrito666, the _whitebox_ ones are in-package
[13:08] <fwereade> perrito666, ie `package foo` not `package foo_test`
[13:09] <fwereade> perrito666, they don't get compiled into proper builds because their names end in _test.go
[13:09] <fwereade> perrito666, but they get all sorts ofinadvisable access to packageinternals
[13:09] <natefinch> perrito666: _whitebox isn't a Go thing, it's just a visual indicator for us.   _test.go is the magic that makes them only compiled during go test
[13:10] <fwereade> natefinch, perrito666: I suspect there may be a couple named _internal_test.go
[13:10] <natefinch> fwereade, perrito666 : I've only seen _whitebox in a few places... it's something I'd rather avoid, since it seems to just add visual noise.  I don't think it really matters that much if tests are internal or not
[13:11] <natefinch> fwereade: I much prefer internal tests anyway, because you can make the tests a lot more focused, so when something breaks, it's more obvious where the breaks are, rather than having to trace through reams and reams of code
[13:12] <fwereade> natefinch, well, it's inevitably testing the implementation not the interface, to some degree at least, and is therefore a bit nasty and makes things harder to change
[13:12] <fwereade> natefinch, so I'd prefer tests in general to be in a different package to the code
[13:13] <fwereade> natefinch, but if they have to be, I also rather agree it's not *that* big a deal if a few internal tests really help
> https://irclogs.ubuntu.com/2014/03/27/%23juju-dev.txt

I also hear it in some tutorials on testing like https://testwithgo.com/ (need to suscribe to view videos, I don't think I can link them here)

I hope those examples are enough

bhcleek:
Thank you for the contribution 🙇 

I'm generally 👎 on this change. These aren't widely accepted standards in the community; if we add this, then someone is going to come along and want to accommodate their new favorite pattern, too. It's easy enough to edit the file you want using `:e`.

PR #2150, opened 2019-02-22
Run godef from buffer path

ldelossa:
This PR stemmed from this issue:
https://github.com/fatih/vim-go/issues/2149

We now execute `godef` in the current directory of the buffer.

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -175,7 +175,7 @@ function! go#tool#ExecuteInDir(cmd, ...) abort
   let dir = getcwd()
   try
     execute cd . fnameescape(expand("%:p:h"))
-    let [l:out, l:err] = call call('go#util#Exec', [a:cmd] + a:000)
+    let [l:out, l:err] = call('go#util#Exec', [a:cmd] + a:000)
   finally
     execute cd . fnameescape(l:dir)
   endtry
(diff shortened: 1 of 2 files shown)


autoload/go/tool.vim
```diff
@@ -175,7 +175,7 @@ function! go#tool#ExecuteInDir(cmd) abort
   let dir = getcwd()
   try
     execute cd . fnameescape(expand("%:p:h"))
-    let [l:out, l:err] = go#util#Exec(a:cmd)
+    let [l:out, l:err] = go#util#Exec(a:cmd, a:000)
```
bhcleek: Since `a:000` is a list that may or may not be populated we have to make a dynamic call to `go#util#Exec` with something like (untested): `let [l:out, l:err] = call call('go#util#Exec', [a:cmd] + a:000)`

ldelossa: Yes I saw that in stack overflow: https://stackoverflow.com/questions/11703297/how-can-i-pass-varargs-to-another-function-in-vimscript but didn't quite wrap my head around it. I'll send another commit with this change.

ldelossa:
@bhcleek had to use the latest commit in order to get the tests to pass. Maybe the double call is not legal when gathering the output ?

bhcleek:
yeah, since the return values are assigned to variables, you don't need `call` statement.

bhcleek:
Thank you for contributing again 🙇

ldelossa:
No problem I rely on vim-go

PR #2156, opened 2019-02-27
correct import path of plugin delve

aofei:
The delve project has been transferred (from derekparker to go-delve). So we should also correct it.

Visiting https://github.com/derekparker/delve will be redirected to https://github.com/go-delve/delve.

This PR is necessary because when `GO111MODULE` is on by default (starting with Go 1.13), the old URL will cause an unexpected module path error.


aofei:
Explain: In the case I described, the unexpected module path error will occur when executing `:GoInstallBinaries` or `:GoUpdateBinaries`.

bhcleek:
thank you

PR #2157, opened 2019-03-01
Add the go_metalinter_config option

BronDo:
Add the go_metalinter_config option


bhcleek:
Why is this needed?

BronDo:
Exclude some warning messages like "exported func XXX should have comment or be unexported"

bhcleek:
Given that gometalinter will pick up a config in the working directory or any of its parents and that gometalinter is deprecated, I don't see much value in adding this.

Thank you for contributing.

PR #2163, opened 2019-03-11
add initial support for gopls

### LINKED ISSUE #2146: implement support for gopls
implement support for [`gopls`](https://godoc.org/golang.org/x/tools/cmd/gopls).

The initial implementation will provide the framework and foundational support in vim-go along with options to use` gopls` for `g:go_def_mode`. Follow on work will provide options to use `gopls` instead of `guru`, for `g:go_info_mode`, and other functions as `gopls` matures, with the ultimate goal of using `gopls` by default and eventually dropping all support for other tools whose functionality is going to be provided by `gopls`.

bhcleek:
Add initial support for gopls. This PR adds the foundation for transitioning to use gopls and introduces a new option for `g:go_def_mode`, `gopls`.

##### extract URI encoding function
Extract function to URI encode strings and add the decoding function,
too.

Use the new uri encoding function when reporting a GitHub issue.

changes. Lines starting

##### teach vim-go to get definition information from gopls

Fixes #2146

### DIFF
--- a/autoload/go/lsp.vim
+++ b/autoload/go/lsp.vim
@@ -1,0 +1,302 @@
+" don't spam the user when Vim is started in Vi compatibility mode
+let s:cpo_save = &cpo
+set cpo&vim
+
+scriptencoding utf-8
+
+let s:lspfactory = {}
+
+function! s:lspfactory.get() dict abort
+  if !has_key(self, 'current')
+    " TODO(bc): check that the lsp is still running.
+    let self.current = s:newlsp()
+  endif
+
+  return self.current
+endfunction
+
+function! s:lspfactory.reset() dict abort
+  if has_key(self, 'current')
+    call remove(self, 'current')
+  endif
+endfunction
+
+function! s:newlsp()
+  if !go#util#has_job()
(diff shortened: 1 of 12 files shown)


autoload/go/lsp.vim
```diff
@@ -0,0 +1,302 @@
...
+          call self.handleInitializeResult(l:response)
+        endfor
+      endif
+
+      if self.ready is 1
+        for l:response in l:responses
+          if has_key(l:response, 'id') && has_key(self.handlers, l:response.id)
+            try
+              if has_key(l:response, 'error')
+                call go#util#EchoError(l:response.message)
```
larrylv: @bhcleek Should this be `call go#util#EchoError(l:response.error)`? I'm getting this when the response has `error`:
```
E716: Key not present in Dictionary: message)
E116: Invalid arguments for function go#util#EchoError
```

btw, thank you for adding the gopls support!

bhcleek: yes! I've updated master accordingly.

larrylv: Thank you!!

PR #2165, opened 2019-03-13
add :GoDefType

bhcleek:
Add a new function, `:GoDefType`, to jump to type definitions. Because
it relies on gopls, `:GoDefType` will only work in Vim8 and Neovim.

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -5,7 +5,7 @@ set cpo&vim
 let s:go_stack = []
 let s:go_stack_level = 0
 
-function! go#def#Jump(mode) abort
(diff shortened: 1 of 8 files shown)


fatih:
What's the difference between :GoDef and :GoDefType? Wasn't `:GoDef` not jumping to type definition already? Maybe I'm missing something :)

bhcleek:
:GoDef will jump to the definition of the identifier under the cursor. When the identifier is a type, it jumps to the type declaration.

:GoDefType will jump to the type declaration of the identifier under the cursor, even when the identifier is a variable.

PR #2173, opened 2019-03-18
Update README.md to notify code completion support via gopls

pankona:
According to this commit https://github.com/fatih/vim-go/commit/579d926ab9b16546febaa43bf104651634c20cbb , vim-go supports code completion using `gopls`.
I think it should be reflected on README.md.


pankona:
Hi,
This is very very small change, but I hope this will help for you.
Thanks!

PR #2182, opened 2019-03-19
add support for golangci-lint

micahcoffman:
Throwing this PR out here to get a discussion started on how we want golangci-lint defaults to look. It's a bit different than gometalinter, but a lot of the gometalinter code copies over nicely. Right now I set the defaults on autosave to be all the "fast" linters and it works pretty well. This seems to be how the other integrations with golangci-lint are working.

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -240,16 +240,32 @@ function! go#config#SetTemplateAutocreate(value) abort
   let g:go_template_autocreate = a:value
 endfunction
 
+function! go#config#Metalinter() abort
+  return get(g:, "go_metalinter", "gometalinter")
+endfunction
+
 function! go#config#MetalinterCommand() abort
   return get(g:, "go_metalinter_command", "")
 endfunction
 
 function! go#config#MetalinterAutosaveEnabled() abort
-  return get(g:, 'go_metalinter_autosave_enabled', ['vet', 'golint'])
+  let l:default_enabled = ["vet", "golint"]
+
+  if go#config#Metalinter() == "golangci-lint"
+    let l:default_enabled = ["deadcode", "errcheck", "ineffassign", "structcheck", "typecheck", "varcheck", "golint"]
+  endif
+
+  return get(g:, "go_metalinter_autosave_enabled", default_enabled)
 endfunction
 
 function! go#config#MetalinterEnabled() abort
-  return get(g:, "go_metalinter_enabled", ['vet', 'golint', 'errcheck'])
+  let l:default_enabled = ["vet", "golint", "errcheck"]
+
+  if go#config#Metalinter() == "golangci-lint"
+    let l:default_enabled = ["deadcode", "errcheck", "ineffassign", "structcheck", "typecheck", "varcheck", "gosimple", "govet", "staticcheck", "unused", "golint"]
+  endif
+
+  return get(g:, "go_metalinter_enabled", default_enabled)
 endfunction
 
 function! go#config#MetalinterDisabled() abort
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -247,8 +247,8 @@ endfunction
 function! go#config#MetalinterAutosaveEnabled() abort
   let l:default_enabled = ["vet", "golint"]
 
-  if go#config#Metalinter() == "golangci-lint"
-    let l:default_enabled = ["deadcode", "errcheck", "ineffassign", "structcheck", "typecheck", "varcheck", "golint"]
+  if go#config#MetalinterCommand() == "golangci-lint"
+    let l:default_enabled = ["govet", "golint"]
   endif
 
   return get(g:, "go_metalinter_autosave_enabled", default_enabled)
@@ -257,8 +257,8 @@ endfunction
 function! go#config#MetalinterEnabled() abort
   let l:default_enabled = ["vet", "golint", "errcheck"]
 
-  if go#config#Metalinter() == "golangci-lint"
-    let l:default_enabled = ["deadcode", "errcheck", "ineffassign", "structcheck", "typecheck", "varcheck", "gosimple", "govet", "staticcheck", "unused", "golint"]
+  if go#config#MetalinterCommand() == "golangci-lint"
+    let l:default_enabled = ["govet", "golint"]
   endif
 
   return get(g:, "go_metalinter_enabled", default_enabled)
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -240,12 +240,8 @@ function! go#config#SetTemplateAutocreate(value) abort
   let g:go_template_autocreate = a:value
 endfunction
 
-function! go#config#Metalinter() abort
-  return get(g:, "go_metalinter", "gometalinter")
-endfunction
-
 function! go#config#MetalinterCommand() abort
-  return get(g:, "go_metalinter_command", "")
+  return get(g:, "go_metalinter_command", "gometalinter")
 endfunction
 
 function! go#config#MetalinterAutosaveEnabled() abort
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -61,7 +61,7 @@ function! go#lint#Gometa(bang, autosave, ...) abort
     "   <file>:<line>:<column>:<severity>: <message> (<linter>)
     "   <file>:<line>::<severity>: <message> (<linter>)
     " This can be defined by the following errorformat:
-    let errformat = "f:%l:%c:%t%*[^:]:\ %m,%f:%l::%t%*[^:]:\ %m"
+    let errformat = "%f:%l:%c:%t%*[^:]:\ %m,%f:%l::%t%*[^:]:\ %m"
   else
     " Golangci-lint can output the following:
     "   <file>:<line>:<column>: <message> (<linter>)
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -256,6 +256,10 @@ function! s:golangcilintcmd(bin_path)
   let cmd += ["run"]
   let cmd += ["--print-issued-lines=false"]
   let cmd += ["--disable-all"]
+  " do not use the default exclude patterns, because doing so causes golint
+  " problems about missing doc strings to be ignored and other things that
+  " golint identifies.
+  let cmd += ["--exclude-use-default=false"]
   return cmd
 endfunction
 
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -9,21 +9,13 @@ function! go#lint#Gometa(bang, autosave, ...) abort
     let goargs = a:000
   endif
 
+  let l:metalinter = go#config#Metalinter()
+
   if empty(go#config#MetalinterCommand())
-    let bin_path = go#path#CheckBinPath("gometalinter")
-    if empty(bin_path)
+    let cmd = s:metalintercmd(l:metalinter)
+    if empty(cmd)
       return
     endif
-
-    let cmd = [bin_path]
-    let cmd += ["--disable-all"]
-
-    " gometalinter has a --tests flag to tell its linters whether to run
-    " against tests. While not all of its linters respect this flag, for those
-    " that do, it means if we don't pass --tests, the linter won't run against
-    " test files. One example of a linter that will not run against tests if
-    " we do not specify this flag is errcheck.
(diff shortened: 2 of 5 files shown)


autoload/go/lint.vim
```diff
@@ -95,6 +95,87 @@ function! go#lint#Gometa(bang, autosave, ...) abort
...
+    let cmd += ["--fast"]
+
+    " linters
+    let linters = a:autosave ? go#config#GolangciLintAutosaveEnabled() : go#config#GolangciLintEnabled()
+    for linter in linters
+      let cmd += ["--enable=".linter]
+    endfor
+
+    for linter in go#config#GolangciLintDisabled()
+      let cmd += ["--disable=".linter]
```
micahcoffman: `--disable` conflicts with the `--disable-all` flag when run together, so need to hash out some more things around that

autoload/go/lint.vim
```diff
@@ -95,6 +95,87 @@ function! go#lint#Gometa(bang, autosave, ...) abort
...
+  if empty(go#config#GolangciLintCommand())
+    let bin_path = go#path#CheckBinPath("golangci-lint")
+    if empty(bin_path)
+      return
+    endif
+
+    let cmd = [bin_path]
+    let cmd += ["run"]
+    let cmd += ["--print-issued-lines=false"]
+    let cmd += ["--fast"]
```
micahcoffman: this flag makes it so only the "fast" linters run, probably don't want to hardcode this

autoload/go/config.vim
```diff
@@ -256,6 +265,22 @@ function! go#config#MetalinterDisabled() abort
...
+function! go#config#GolangciLintAutosaveEnabled() abort
+  return get(g:, 'go_golangci_lint_autosave_enabled', ['deadcode', 'errcheck', 'ineffassign', 'structcheck', 'typecheck', 'varcheck'])
+endfunction
+
+function! go#config#GolangciLintEnabled() abort
+  return get(g:, "go_golangci_lint_enabled", ['deadcode', 'errcheck', 'ineffassign', 'structcheck', 'typecheck', 'varcheck', 'gosimple', 'govet', 'staticcheck', 'unused'])
+endfunction
+
+function! go#config#GolangciLintDisabled() abort
+  return get(g:, "go_golangci_lint_disabled", ['megacheck'])
```
micahcoffman: golangci-lint gives a nasty warning message about not being able to run megacheck on code that can't compile unless megacheck is explicitly disabled. there's probably a better way of getting that warning message to go away

micahcoffman:
@bhcleek I'm +1 on all those ideas, just wasn't sure where you all wanted to head with gometalinter, so I spun this up without touching any of that code! I can start in on some of those improvements.

#3 is a great idea

bhcleek:
I appreciate you being careful 🙇 

FWIW, gometalinter is deprecated and soon to be archived: https://github.com/alecthomas/gometalinter#go-meta-linter. By providing an option to allow a user to choose between `gometalinter` and `golangci-lint`, we can be assured that it will be very easy to remove the unnecessary bits for `gometalinter` once we're sure everything works as expected with `golangci-lint`.

Because `gometalinter` and `golangci-lint` are a meta linter, I'm comfortable with using `golangci-lint` via `:GoMetaLinter`, because it's _not_ `:GoGoMetaLinter` 😄

micahcoffman:
@bhcleek I updated the code to your suggestions, defaulting to `gometalinter` as the default `g:go_metalinter` so no one's config breaks. Take a look and let me know what you think. I'd still like to take a few to consider the best default flags for golangci-lint.

autoload/go/config.vim
```diff
@@ -240,20 +240,42 @@ function! go#config#SetTemplateAutocreate(value) abort
...
 function! go#config#MetalinterCommand() abort
   return get(g:, "go_metalinter_command", "")
 endfunction
 
 function! go#config#MetalinterAutosaveEnabled() abort
-  return get(g:, 'go_metalinter_autosave_enabled', ['vet', 'golint'])
+  let l:default_enabled = ["vet", "golint"]
+
+  if go#config#Metalinter() == "golangci-lint"
+    let l:default_enabled = ["deadcode", "errcheck", "ineffassign", "structcheck", "typecheck", "varcheck"]
```
bhcleek: Why have you chosen these as the defaults?

At a minimum, I'd expect the defaults for golangci-lint to at least use `golint` and `govet` to maintain parity with gometalinter for now.

autoload/go/config.vim
```diff
@@ -240,20 +240,42 @@ function! go#config#SetTemplateAutocreate(value) abort
...
+
+  return get(g:, "go_metalinter_autosave_enabled", default_enabled)
 endfunction
 
 function! go#config#MetalinterEnabled() abort
-  return get(g:, "go_metalinter_enabled", ['vet', 'golint', 'errcheck'])
+  let l:default_enabled = ["vet", "golint", "errcheck"]
+
+  if go#config#Metalinter() == "golangci-lint"
+    let l:default_enabled = ["deadcode", "errcheck", "ineffassign", "structcheck", "typecheck", "varcheck", "gosimple", "govet", "staticcheck", "unused"]
```
bhcleek: A similar question here: why have you chosen these as the defaults?

At a minimum, I'd expect the defaults for golangci-lint to at least use `golint`, `govet`, and `errcheck` to maintain parity with gometalinter.

autoload/go/lint.vim
```diff
@@ -44,15 +36,17 @@ function! go#lint#Gometa(bang, autosave, ...) abort
...
-    let include = [printf('--include=^%s:.*$', fnamemodify(expand('%:p'), ":."))]
-    if go#util#has_job()
-      let include = [printf('--include=^%s:.*$', expand('%:p:t'))]
+    if l:metalinter == "gometalinter"
+      " Include only messages for the active buffer for autosave.
+      let include = [printf('--include=^%s:.*$', fnamemodify(expand('%:p'), ":."))]
+      if go#util#has_job()
+        let include = [printf('--include=^%s:.*$', expand('%:p:t'))]
+      endif
+      let cmd += include
```
bhcleek: How can we ensure that only messages for the active buffer are included when using golangci-lint?

autoload/go/config.vim
micahcoffman: These are the "fast" defaults of golangci-lint https://github.com/golangci/golangci-lint#enabled-by-default-linters. There's also a `--fast` flag available to use, but we need to pass it in as another config because the non-autosave command shouldn't be restricted to "fast" linters.

I personally use `golint`, but it's not part of their defaults, so I didn't throw it in. I would be happy putting it in.

Their `govet` seems a bit odd to be honest, I haven't gotten it to mimic the output of a regular `go vet ./...`. It seems like they do some exclude duplicate checks between linters, so typecheck which is default takes over some of the govet behaviors. And then their `govet` just does some other source code analysis.

autoload/go/config.vim
micahcoffman: these are just straight from the golangci-lint defaults https://github.com/golangci/golangci-lint#enabled-by-default-linters

autoload/go/lint.vim
micahcoffman: they unfortunately don't have an `include` like gometalinter. i'll take another look at their flag options to see if there's something we can do

autoload/go/config.vim
bhcleek: Yeah, let's put in `golint`.

autoload/go/config.vim
bhcleek: That makes sense; I'm just considering whether vim-go should try to maintain parity for the default linters regardless of whether it's using `gometalinter` or `golangci-lint` or whether it's ok to vary the default linters depending on which metalinter is being used.

autoload/go/config.vim
micahcoffman: added

autoload/go/lint.vim
micahcoffman: figured this out, golangci-lint can be run per file so I added https://github.com/fatih/vim-go/pull/2182/commits/f574b1f06dee7a23483e79f55c956c83763ba866

autoload/go/lint.vim
```diff
@@ -3,28 +3,24 @@ let s:cpo_save = &cpo
...
   if a:0 == 0
-    let goargs = [expand('%:p:h')]
+    if l:metalinter == "gometalinter"
+      let goargs = [expand('%:p:h')]
+    else
+      let goargs = [expand('%:p')]
+    endif
   else
     let goargs = a:000
   endif
```
bhcleek: This block changes things somewhat. When `a:autosave` is not set, then both gometalinter and golangci-lint should be passed the directory. When `a:autosave` is set then `golangci-lint` should be passed the file to check.

bhcleek: To be clear, `gometalinter` also needs to be passed the file to check when `a:autosave` is set; that already happens for `gometalinter` a little further down the file. Perhaps this block could be restored to its previous glory, and the section below that ensure that `gometalinter` is given the `--include` option when `a:autosave` is set could be extended to append the file path to `goargs` when `golangci-lint` is used.

micahcoffman: Cool, that makes sense to me. I did this https://github.com/fatih/vim-go/pull/2182/commits/77243e3c68c32696a62146d06e8f106567f4b3cd instead, do you think it could just be `let goargs = [expand('%p')]`? What all do we expect goargs could be when autosave is enabled?

micahcoffman:
I also updated errorformat to include a better format for golangci-lint. It's slightly different than gometalinter.

autoload/go/config.vim
```diff
@@ -240,16 +240,40 @@ function! go#config#SetTemplateAutocreate(value) abort
...
+  return get(g:, "go_metalinter", "gometalinter")
+endfunction
+
+function! go#config#MetalinterConfigEnabled() abort
+  return get(g:, "go_metalinter_config_enabled", 0)
+endfunction
+
+function! go#config#MetalinterGrepNew() abort
+  return get(g:, "go_metalinter_grep_new", 0)
+endfunction
```
bhcleek: Neither `go#config#MetalinterConfigEnabled` nor `go#config#MetalinterGrepNew` is needed. vim-go doesn't need to support every possible option. If people want an escape hatch, there's already one in the form of `g:go_metalinter_command`.

micahcoffman: Sorry, agreed, I didn't intend to push that. I'm playing around with the best way to interact with the golangci-lint config. Config values are trumping flags which is a bit different than I expected.

bhcleek: let's not worry about it; there's an escape hatch, and people can use the config as they see fit.

bhcleek:
> @bhcleek I'm ok with this PR as is now.

Once you remove those 85d4505 and squash all these commits, let me know. Do you mind if I push to your branch to update docs and make some small tweaks before merging?

micahcoffman:
Should be good to go. You're welcome to push to my branch.

bhcleek:
This looks great! Thank you for the contribution.

I'll make some small tweaks, update docs, and run with this locally for a while before I merge it.

tommyknows:
Hi,
I just updated to use `golanglint-ci` and have a question;
right now, when enabling specific linters (for example `typecheck`) I get an error when having declarations outside the file that's currently in my buffer.
This is because the linter only checks one file, and not the whole package.
Is there an option to lint the whole package instead?

```
> golangci-lint run --disable-all --enable 'typecheck' pkg/x/y.go
pkg/x/y.go:71:2: undeclared name: `procRequests` (typecheck)
        procRequests.Inc()
        ^
> golangci-lint run --disable-all --enable 'typecheck' pkg/x/
>
```

windvalley:
> Hi,
> I just updated to use `golanglint-ci` and have a question;
> right now, when enabling specific linters (for example `typecheck`) I get an error when having declarations outside the file that's currently in my buffer.
> This is because the linter only checks one file, and not the whole package.
> Is there an option to lint the whole package instead?
> 
> ```
> > golangci-lint run --disable-all --enable 'typecheck' pkg/x/y.go
> pkg/x/y.go:71:2: undeclared name: `procRequests` (typecheck)
>         procRequests.Inc()
>         ^
> > golangci-lint run --disable-all --enable 'typecheck' pkg/x/
> >
> ```

solution:
add 'call ale#Set('go_golangci_lint_package', 1)' to your ~/.vimrc

PR #2230, opened 2019-04-14
Remove mention of vim-godebug from documentation

velovix:
Now that vim-go has first-class support for Delve, there doesn't seem a reason to recommend another plugin to do the same thing. If there's something this plugin does that the native Delve support does, I can add mention of that to this section instead.


velovix:
Sorry, meant to say: "If there's something this plugin does that the native Delve support *does not do*, I can add mention of that in this section instead."

bhcleek:

Issue #2185, opened 2019-03-21
doc: vim-go + deoplete + deplete-go + gopls


dnwe:
### What did you do? (required. The issue will be **closed** when not provided.)

Upgraded to latest vim-go and wanted to experiment with the new `gopls` completion

### What did you expect to happen?

I expected `gopls` to be used for both the built-in omnifunc binding (<C-x><C-o>) and the `.` real-time completion from deoplete/deoplete-go

### What happened instead?

I found that `gopls` was working via omnifunc but deoplete-go was still using gocode

### Configuration (**MUST** fill this out):

* vim-go version: origin/master@cf0b9c9

* `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):

```
Plug 'fatih/vim-go', { 'do': ':GoUpdateBinaries', 'for': 'go' }
Plug 'shougo/deoplete.nvim', { 'do': ':UpdateRemotePlugins' }
Plug 'zchee/deoplete-go', { 'do': 'make' }

let g:go_def_mode = "gopls"
```

* Vim version (first three lines from `:version`): NVIM v0.3.4
* Go version (`go version`): go version go1.12 darwin/amd64

dnwe:
p.s. thank you _so_ much for the hard work of implementing `gopls` support. It is sooo good

dnwe:
So I _thought_ I had fixed this by adding these config options:
```
    let g:deoplete#sources#go = ['vim-go']
    let g:deoplete#sources#go#gocode_binary = '/dev/null'
```
But after experimenting with adding go_debug for 'lsp' I still don't believe deoplete is using gopls via vim-go

Could we add a small note in the `:help vim-go-plugins` section to mention a recommended configuration for using this via deoplete?

bhcleek:
This isn't an issue for vim-go. An issue on deoplete asking them how to configure deoplete for Go would be appropriate. vim-go sets up an omnifunc that uses gopls.

velovix:
I've opened [an issue on Deoplete](https://github.com/Shougo/deoplete.nvim/issues/965) for this.

velovix:
As per the issue on deoplete, you need to set an Omni pattern, which tells deoplete when to call the omnifunc:

```
call deoplete#custom#option('omni_patterns', {
\ 'go': '[^. *\t]\.\w*',
\})
```

[...]

PR #2231, opened 2019-04-15
Add docs for using deoplete with gopls completion

velovix:
This recommends configuring the "omni_patterns" parameter, which instructs deoplete to call the omnifunc when a certain regular expression is matched.

I found this solution here: https://github.com/Shougo/deoplete.nvim/issues/965
The idea of adding docs for this was discussed here: https://github.com/fatih/vim-go/issues/2185

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -138,8 +138,16 @@ The following plugins are supported for use with vim-go:
   https://github.com/Shougo/neocomplete.vim
 
 * Real-time completion (Neovim and Vim 8):
-  https://github.com/Shougo/deoplete.nvim and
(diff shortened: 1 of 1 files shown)


bhcleek:
Vim-go supports both `gocode` and `gopls` are currently supported for completion. Though `gopls` is used by default in Vim 8 and Neovim, it can be switched to use `gocode` by setting `omnifunc` to `go#complete#GocodeComplete` in an `after/ftplugin/go.vim` file. Is there any reason to recommend `https://github.com/zchee/deoplete-go`?

velovix:
Good point! I've removed mention of deoplete-go and rephrased the section to not specifically mention either completion back end.

PR #2307, opened 2019-05-24
Fix errors

mattn:
I'm using Go tip and got error.
(Also fixes another error when return empty lines from format)

### DIFF
--- a/autoload/go/mod.vim
+++ b/autoload/go/mod.vim
@@ -8,7 +8,11 @@ function! go#mod#Format() abort
   " go mod only exists in `v1.11`
   if empty(s:go_major_version)
     let tokens = matchlist(go#util#Exec(['go', 'version']), '\d\+.\(\d\+\)\(\.\d\+\)\? ')
-    let s:go_major_version = str2nr(tokens[1])
+    if len(tokens) > 0
+      let s:go_major_version = str2nr(tokens[1])
+    else
+      let s:go_major_version = 13
+    endif
   endif
 
   if s:go_major_version < "11" 
--- a/autoload/go/mod.vim
+++ b/autoload/go/mod.vim
@@ -11,11 +11,11 @@ function! go#mod#Format() abort
     if len(tokens) > 0
       let s:go_major_version = str2nr(tokens[1])
     else
-      let s:go_major_version = 13
+      let s:go_major_version = ""
     endif
   endif
 
-  if s:go_major_version < "11" 
+  if !empty(s:go_major_version) && s:go_major_version < "11"
     call go#util#EchoError("Go v1.11 is required to format go.mod file")
     return
   endif
--- a/autoload/go/package.vim
+++ b/autoload/go/package.vim
@@ -163,7 +163,11 @@ function! go#package#FromPath(arg) abort
     return -1
   endif
 
-  let l:importpath = split(l:out, '\n')[0]
+  let l:lines = split(l:out, '\n')
+  if empty(l:lines)
+    return -1
+  endif
+  let l:importpath = l:lines[0]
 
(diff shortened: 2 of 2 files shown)


autoload/go/mod.vim
```diff
@@ -8,7 +8,11 @@ function! go#mod#Format() abort
   " go mod only exists in `v1.11`
   if empty(s:go_major_version)
     let tokens = matchlist(go#util#Exec(['go', 'version']), '\d\+.\(\d\+\)\(\.\d\+\)\? ')
-    let s:go_major_version = str2nr(tokens[1])
+    if len(tokens) > 0
+      let s:go_major_version = str2nr(tokens[1])
+    else
+      let s:go_major_version = 13
```
mattn: Do you prefer 0 and ignore checking below?

bhcleek:
Thanks @mattn.

What are `go version` and `go list` outputting?

mattn:
```
C:\>go version
go version devel +1ab063ce53 Fri May 17 22:32:30 2019 +0000 windows/amd64
```

And `go list` outside GOPATH should be an error.

bhcleek:
`go list` outside `GOPATH` with `*.go` files present and without a go.mod file (i.e. a null module) returns the absolute path with a `_` prefix. With no `*.go` files present `go list` already returns a non-zero exit code and `go#package#FromPath` returns -1 in that case on line 163 (https://github.com/fatih/vim-go/blob/f45093914a9dcfa8d39d5849c100b0118c310fcd/autoload/go/package.vim#L162-L164).

Can you provide more information about your use case for the change to `go#package#FromPath`?

mattn:
Trying outside GOPATH.

terminal7

bhcleek:
Thanks! So to be clear, you're solving for the case of trying to use completion of packages when in a null module?

mattn:
Yes, it is.

autoload/go/package.vim
```diff
@@ -82,13 +86,21 @@ function! s:vendordirs() abort
...
     let [l:dir, l:err] = go#util#ExecInDir(['go', 'list', '-f', '{{.Dir}}'])
     if l:err != 0
       return []
     endif
-    let l:dir = split(l:dir, '\n')[0]
+    let l:lines = split(l:dir, '\n')
+    if empty(l:lines)
+      return {}
+    endif
+    let l:dir = l:lines[0]
```
bhcleek: Why is this second new block necessary? Assuming that `go list` in a null modules exits 0, is there some condition where `.Root` will have a value, but `.Dir` will not?

autoload/go/package.vim
```diff
@@ -163,7 +175,11 @@ function! go#package#FromPath(arg) abort
     return -1
   endif
 
-  let l:importpath = split(l:out, '\n')[0]
+  let l:lines = split(l:out, '\n')
+  if empty(l:lines)
+    return -1
+  endif
+  let l:importpath = l:lines[0]
```
bhcleek: How can `go list` with no arguments exit 0 _and_ not have any lines in its output?

autoload/go/package.vim
mattn: No. go list return `.` not null.

image

mattn: go list possibly return empty and 0.

image

mattn: Oops. Sorry. This is my fault.

mattn:
Sorry, This is my mistake. Ignore this.

bhcleek:
I think there was some good stuff in here. Are you sure you want to close it?

mattn:
Ah, checking tip is required.

mattn:
Rebased from master branch & removed my some changes about empty outputs.

bhcleek:
Thanks, @mattn

PR #2347, opened 2019-06-10
support vim8 popup window when run GoDoc

skanehira:
When vim8 has patch `patch-8.1.1513` we can use popup window.
set `let go_doc_popup_window = 1` and run `:GoDoc` , document will display on the popup window

image

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -178,6 +178,9 @@ function! go#config#DocUrl() abort
   return godoc_url
 endfunction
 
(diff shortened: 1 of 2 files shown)


bhcleek:
Thanks. This looks great. The CI tests are failing, but I think it's unrelated. I'll merge this as soon as I can confirm that the failing CI tests aren't at all related.

vitaly-zdanevich:
Is it possible to show this popup on mouse hover, like in ALE?

PR #2358, opened 2019-06-17
snippets: rename snippet errn->err

sebastianst:
This PR renames the snippet `errn` to `err`, since `errn` is already in use for the return of two variables. It also makes sense to give the simple return of `err` the shortest snippet name `err`.


bhcleek:

fatih:
And I was wondering why this is broken for me :/ Added a revert for this: https://github.com/fatih/vim-go/pull/2438

PR #2367, opened 2019-06-26
fix(golangci-lint): Run linter on the directory

gnuletik:
Fixes false positives from `unused` & `typecheck` linters.

For example:
mypackage/a.go has function `myFunc()`
mypackage/b.go use function `myFunc()`

threw a `unused` error in mypackage/a.go and a `undef` error in mypackage/b.go

This now checks the files in all directories instead of the current file/buffer

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -44,7 +44,7 @@ function! go#lint#Gometa(bang, autosave, ...) abort
       endif
       let cmd += include
     elseif l:metalinter == "golangci-lint"
-      let goargs[0] = expand('%:p')
+      let goargs[0] = expand('%:p:h')
     endif
   endif
 


bhcleek:
`golangci-lint` is only given the file to lint when linting on save. In that case, we _only_ want to lint the file being saved.

bhcleek:
I understand why you want to do this, though. Perhaps the right solution is to filter the errors coming back from golangci-lint on the filenames. That's relatively easy to do with the job support in vim-go...

bhcleek:
Thank you for the PR. I'm going to merge this soon and extend it to support vim-go expected behavior.

gnuletik:
@bhcleek Thanks for having a look !

Yes filtering the errors on the current filename seems totally appropriate.
If you know how this can be done, okay :) I'm farily new to vimscript.

Thanks !

bhcleek:
Doing this as-is causes other problems, too. When not in GOPATH, `golangci-lint run` doesn't seem to report any problems when then argument it's given is a directory instead of a file.

bhcleek:
I created https://github.com/golangci/golangci-lint/issues/592 to demonstrate the problem that golangci-lint is having. Since the problem is limited to null modules, I'll likely merge this PR with some additional changes soon.

Issue #2396, opened 2019-07-13
NVIM GoTestFunc runs tests multiple times


ldelossa:
### What did you do? (required: The issue will be **closed** when not provided)
Have a test file which I run "GoTestFunc" on. This is an integration test which writes and reads from a database.

### What did you expect to happen?
The test is performed in a single time and in the correct order for the read/write/teardown of the database to occur

### What happened instead?
Postgres logs indicate the test is being ran more than once. This is not reproducible on the command line running go test directly. 

### Configuration (**MUST** fill this out):
https://github.com/ldelossa/dotfiles/blob/master/vimrc

#### vim-go version:
MASTER 

####  `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):
https://github.com/ldelossa/dotfiles/blob/master/vimrc

#### Vim version (first three lines from `:version`):

NVIM v0.3.8
Build type: Release
LuaJIT 2.0.5

#### Go version (`go version`):

go version go1.12.1 darwin/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GOARCH="amd64"
GOBIN="/Users/louis/git/gopath/bin"
GOCACHE="/Users/louis/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/louis/git/gopath"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/louis/git/go/libcscan/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/mc/xwfdd41j1j74xfmcmphbwpgw0000gn/T/go-build977618147=/tmp/go-build -gno-record-gcc-switches -fno-common"

</pre></details>
[...]

PR #2397, opened 2019-07-13
fix test terminal handling

ldelossa:
fixes https://github.com/fatih/vim-go/issues/2396

tested and working.


bhcleek:
Thank you !

PR #2398, opened 2019-07-14
add a feature to generate return statement

110y:
Add a new command `GoReturn` which generates `return` statement with zero values for current function. 
The command uses [goreturn](https://github.com/110y/goreturn) which was also created by me based on [iferr](https://github.com/koron/iferr).

### e.g.

```go
func doSomething() (string, error) {
}
```

becomes:

```go
func doSomething() (string, error) {
    return "", nil
}
```

### DIFF
--- a/autoload/go/return.vim
+++ b/autoload/go/return.vim
@@ -1,0 +1,26 @@
+" don't spam the user when Vim is started in Vi compatibility mode
+let s:cpo_save = &cpo
+set cpo&vim
+
+function! go#return#Generate()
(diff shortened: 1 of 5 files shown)


autoload/go/return.vim
```diff
@@ -0,0 +1,26 @@
+" don't spam the user when Vim is started in Vi compatibility mode
```
110y: same as `go#iferr#Generate` excepts executed command.

PR #2402, opened 2019-07-15
fix off by one error in text placement

ldelossa:
This fixes issue where autocomplete would not replace the first character after a "."

For example when attempting to auto complete `assert.Equals`
```
assert.E<auto complete>
```
You'd wind up with:
```
assert.EEqual
```

Looks like this was an 0 index issue.

### DIFF
--- a/autoload/go/lsp/lsp.vim
+++ b/autoload/go/lsp/lsp.vim
@@ -38,7 +38,7 @@ function! go#lsp#lsp#PositionOf(content, units) abort
   let l:remaining = a:units
   let l:str = ""
   for l:rune in split(a:content, '\zs')
-    if l:remaining < 0
+    if l:remaining <= 0
       break
     endif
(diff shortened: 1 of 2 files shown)


bhcleek:
I can duplicate the problem that you're having, and I think I know where to fix it, but this solution is not the right one and will break other things. The important thing to know is that text positions in LSP are 0-based, while character positions in Vim are 1-based. `go#lsp#lsp#PositionOf` converts from the LSP position to the Vim position, and changing it as you have here will break other things (e.g. `:GoInfo`, `:GoDef`, etc.).

I'm closing this for now and will work out the correct fix. I'm pretty sure that the problem is at https://github.com/fatih/vim-go/blob/master/autoload/go/complete.vim#L284. There are a couple of non-obvious edge case that I need to test before I create a PR, though.

PR #2409, opened 2019-07-17
toggle whether term closes

ldelossa:
This PR introduces a way to quickly toggle whether the nvim terminal is closed after a quickfix menu is populated. 

I found this helpful when needing to add `log.Printf()` statements in my tests as the term formatter function does not capture stdout. 

I am open to discussion here.

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -47,19 +47,12 @@ function! go#config#TermMode() abort
   return get(g:, 'go_term_mode', 'vsplit')
 endfunction
 
-function! go#config#TermExit() abort
-  return get(g:, 'go_term_exit', 1)
+function! go#config#TermCloseOnExit() abort
+  return get(g:, 'go_term_close_on_exit', 1)
 endfunction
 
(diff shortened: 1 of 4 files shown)


autoload/go/config.vim
```diff
@@ -47,6 +47,21 @@ function! go#config#TermMode() abort
...
+endfunction
+
+function! go#config#ToggleTermExit() abort 
+  let l:cur = get(g:, 'go_term_exit', 0)
+  if l:cur == 0
+    silent! unlet g:go_term_exit
+    let g:go_term_exit = 1
+  else 
+    silent! unlet g:go_term_exit
+    let g:go_term_exit = 0
```
bhcleek: There's no need to set it again; `unlet` will suffice just fine.

ldelossa:
Dont mind at all.

bhcleek:
@ldelossa I'm not a heavy Neovim user. Would you mind testing my changes to make sure they work for you as you'd expect?

ldelossa:
@bhcleek confirmed working. Thanks!!

PR #2411, opened 2019-07-19
Check function exists

mattn:
```
:let $XXX = 1
:echo go#util#env("XXX")
```
Does not work since does not check function `go#util#xxx` exists.


bhcleek:
Which go environment variables from `go env` are used in vim-go, but don't have a `go#util#xxx` function?

PR #2426, opened 2019-07-27
lsp: send initialized notification

bhcleek:
Send the initialized notifcation to gopls after receiving its initialize
response so that configuration will be processed and the message
indicating where the gopls diagnostic server is available will be
output.


jason-riddle:
@bhcleek Is there a way to suppress this message?

PR #2437, opened 2019-08-01
plugin: redraw before updating/installing each binary

bhcleek:
Redraw before updating or installing each binary so that users of small
terminals don't get prompted to hit Enter a bunch.


bhcleek:
Closing in favor of a change that will `set nomore` before installing all the plugins.

PR #2438, opened 2019-08-02
snippets: revert #2358

fatih:
There were two of them, one is `errn` and `errn,` (note the coma at the
end). We didn't use `err` because that is sometimes a variable itself
and you don't want to autoexpand it. Or sometimes the completer catches
it.

We should revert it,  because `errn` is not in use at all,
so this change was wrong in the first place and this breaks things for
many people who were using `errn` before. It's not a backwards compatible
change.

### DIFF
--- a/gosnippets/UltiSnips/go.snippets
+++ b/gosnippets/UltiSnips/go.snippets
@@ -151,7 +151,7 @@ if err := ${1:condition}; err != nil {
 endsnippet
 
 # error snippet
-snippet err "Error return" !b
+snippet errn "Error return" !b
 if err != nil {
 	return err
 }


bhcleek:
@fatih why does it matter that `err` is a real variable value sometimes? Since the use of snippets requires that you expand the snippet by pressing `<TAB>` or some other key combo, what's the concern with naming this one `err`? What am I missing?

fatih:
By default, most of the completion plugins have multiple sources (not just `omnifunc). For example in my case, my autocompletion plugin also provides candidates from variables in the local buffer. `err` is a very common variable name, hence this happens. When I start to write an expression that starts with:

```
err := 
```

after the last `r` character, suddenly now I have two completions. Some people use `tab` as to show multiple sources. In my case it's all ok, I reverted it because the change is not backwards compatible.

PR #2439, opened 2019-08-03
Update popup window's judgement

kazukazuinaina:
I think this patch is better about judgement of can use or not use popup window.


bhcleek:
Thank you for contributing 🙇

PR #2440, opened 2019-08-03
Add GoAddBinaryToInstall() to install extra tools.

tobinjt:
I'd like to use GoInstallBinaries to install other Golang tools, e.g.
goimports for use by Syntastic.  This commit adds a function
GoAddBinaryToInstall() that extends s:packages with more tools to
install so that GoInstallBinaries can install them.

### DIFF
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -510,6 +510,15 @@ CTRL-t
 
     Set |'g:go_get_update'| to disable updating dependencies.
 
+                                                      *GoAddBinaryToInstall()*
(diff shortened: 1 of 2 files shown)


bhcleek:
Thank you for the contribution. This is out of scope for vim-go, though. The tools that it installs are based on vim-go's needs, and is not intended to be a Go tool manager per se.

FWIW, though, `goimports` is already installed by vim-go: https://github.com/fatih/vim-go/blob/master/plugin/go.vim#L57

PR #2446, opened 2019-08-08
lsp: do not strip godoc from hover content

bhcleek:
Now that vim-go uses ClientCapabilities to ask for hover content with
NoDocumentation, there's no need to attempt to strip godoc from hover
content in infoFromHoverContent.

### DIFF
--- a/autoload/go/lsp.vim
+++ b/autoload/go/lsp.vim
@@ -680,16 +680,6 @@ function! s:infoFromHoverContent(content) abort
 
   let l:content = a:content[0]
 
-  " strip godoc summary
-  " Hover content with godoc summary will have the godoc summary in the first
-  " line, and the second line will not have leading whitespace. When there is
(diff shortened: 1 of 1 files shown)


autoload/go/lsp/message.vim
```diff
@@ -19,7 +19,7 @@ function! go#lsp#message#Initialize(wd) abort
               \ },
               \ 'textDocument': {
                 \ 'hover': {
-                  \ 'contentFormat': ['plaintext'],
```
stamblerre: I think this is still correct here. The "NoDocumentation" configuration should be set in the LSP config that `gopls` parses [here](https://github.com/golang/tools/blob/e713427fea3f98cb070e72a058c557a1a560cf22/internal/lsp/general.go#L230).

bhcleek: Good call. Thanks for the extra set of eyes :bow:

PR #2447, opened 2019-08-09
Improve some godoc's popup window

skanehira:
I improve some :)

- Only close godoc's popup window
- Can to be able scroll popup window
- Display a window in the center of the scree

[asciicast](https://asciinema.org/a/261678)


bhcleek:
Thank you for the contribution. When the current tab is split into two windows, though, I suspect that this won't work quite the way you might expect. A few things to consider:

* a user has two windows open and tries to show godoc in each.
* the window in which the popup will be shown is 10 lines tall or less
* the window in which the popup will be shown is 10 columns wide or less

skanehira:
@bhcleek 
>a user has two windows open and tries to show godoc in each.
I think if user want to open more one godoc, use buffer is more better.
also, I think that there is no problem because the user can choose whether or not to use the popup window.

>the window in which the popup will be shown is 10 lines tall or less
>the window in which the popup will be shown is 10 columns wide or less

How about fixing the side of the popup window to 10 and using `go_doc_max_height` for the vertical?

bhcleek:
> I think if user want to open more one godoc, use buffer is more better.

The option is global, though, and doesn't take into account how many windows are open, nor does it take into account how many tabs are open, it simpy closes the last opened godoc popup window. I think this can be handled better by making `s:last_popup_window` an object whose keys are the windows that contain the popup windows and the values are the popup window ids.


```
 \ 'maxheight': &lines,
 \ 'maxwidth': &columns
```
 
I'll think about ☝️  and get back to you

skanehira:
>The option is global, though, and doesn't take into account how many windows are open, nor does it take into account how many tabs are open, it simpy closes the last opened godoc popup window. I think this can be handled better by making s:last_popup_window an object whose keys are the windows that contain the popup windows and the values are the popup window ids.

It's easy to display multiple popup windows, but how do you think it should be displayed and controlled?

I think the following controls are needed.

- requires key binding to focus between multiple popup windows
- since user cannot do any thing the current window while displaying popup windows, user must close all popup windows or disable popup window's filter function when want to do something.
- when disable popup window's filter after, if user want to scroll godoc in popup window, must run some Ex command to focus to popup window.

In this way, the operation becomes complicated, so i think that usability will be reduced.
If you have a good idea, please let me know

bhcleek:
I'll play with this and get back to you.

skanehira:
@bhcleek Hi, how about this PR?

bhcleek:
@skanehira I'll get to this soon. I have a couple of things in front of it that I need to get done first. I'll update here as soon as I've had a chance to review more thoroughly.

skanehira:
@bhcleek I got it :)

bhcleek:
@skanehira should I close this?

skanehira:
@bhcleek I'm sorry for being late.

>Can you tell me more about why you prefer the doc window to be in the center of the screen instead of near the cursor?

If window popup at the screen center, window can use more screen space for long doc.
And, use `popup_menu` can scroll long doc.

>Have you encountered some problems with the current implementation that motivated you to set maxheight and maxwidth?

I want to watch long doc using more screen space, so I used maxheight and max width.

bhcleek:
`popup_menu` sets more options than are appropriate for this. Can you switch it to `popup_atcursor` instead? You should still be able to get the scrolling behavor you desire.

skanehira:
@bhcleek Ok, I'll use `popup_atcursor` 👍

bhcleek:
@skanehira I'm wondering whether you plan to finish this up? If not, that's ok; let me know so I can decide what to do with this PR.

skanehira:
@bhcleek Sorry, I'm so late..
after much consideration,I thought that this PR was unnecessary.
Therefore, it will be closed.

PR #2449, opened 2019-08-10
config: add test for setting build tags

bhcleek:
Add test to verify that gopls works as expected with build tags.

This test is not expected to pass until https://github.com/golang/tools/commit/09f9cfa88292dd2b01eb8fc08387898ff3561f27 is released.


bhcleek:
All the tests passed. The build failed due to other reasons related to the coverage python package.

PR #2450, opened 2019-08-10
highlight function calls as functions

neta540:
To me, it makes more sense to highlight goFunctionCall as Function instead of Type.

https://github.com/fatih/vim-go/blob/ed63c87496c030d436507c4eb211a923aa5dc90b/syntax/go.vim#L291

My expectation would be: 
`hi def link     goFunctionCall      Function`


bhcleek:
Thank you for taking the time to contribute and trying to improve vim-go.

These are mostly function definitions, not actually function calls. Since function are first class citizens in Go and can be used as types, Type is appropriate.

PR #2451, opened 2019-08-11
Add support for DocPopupWindow in Neovim

georgijd:
Basic support for floating windows in Neovim. It's using minimal style to disable some UI components like end-of-buffer region or special columns.

Imgur

### DIFF
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -107,9 +107,9 @@ function! s:GodocView(newposition, position, content) abort
             \ 'col': 0,
             \ 'width': width,
             \ 'height': height,
-            \ 'style': 'minimal'
             \ }
       call nvim_open_win(buf, v:true, opts)
+      setlocal nonumber norelativenumber nomodified nomodifiable syntax=go
 
       " close easily with CR, Esc and q
       noremap <buffer> <silent> <CR> :<C-U>close<CR>
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -77,14 +77,45 @@ endfunction
 
 function! s:GodocView(newposition, position, content) abort
   " popup window
-  if go#config#DocPopupWindow() && exists('*popup_atcursor') && exists('*popup_clear')
-    call popup_clear()
+  if go#config#DocPopupWindow()
+    if exists('*popup_atcursor') && exists('*popup_clear')
+      call popup_clear()
 
-    call popup_atcursor(split(a:content, '\n'), {
-          \ 'padding': [1, 1, 1, 1],
-          \ 'borderchars': ['-','|','-','|','+','+','+','+'],
-          \ "border": [1, 1, 1, 1],
-          \ })
+      call popup_atcursor(split(a:content, '\n'), {
+            \ 'padding': [1, 1, 1, 1],
+            \ 'borderchars': ['-','|','-','|','+','+','+','+'],
+            \ "border": [1, 1, 1, 1],
+            \ })
+    elseif has('nvim') && exists('*nvim_win_set_config')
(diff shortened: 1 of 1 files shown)


skanehira:
@georgijd Hi!
I improve some GoDoc popup window in #2447.
We are discussing the use case of the popup window, so if you like, please tell us your opinion.

autoload/go/doc.vim
```diff
@@ -77,14 +77,45 @@ endfunction
...
+
+      let buf = nvim_create_buf(v:false, v:true)
+      call nvim_buf_set_lines(buf, 0, -1, v:true, lines)
+      let opts = {
+            \ 'relative': 'cursor',
+            \ 'row': 1,
+            \ 'col': 0,
+            \ 'width': width,
+            \ 'height': height,
+            \ 'style': 'minimal'
```
frescoraja: `nvim_open_win` options seems not to support `style` key anymore. see `help nvim_open_win`, i use nvim v0.4.0

could i also suggest setting buffer syntax to `go` for the doc popup window?

thanks

georgijd: Hi, thank you for your feedback! My bad on the `style` key, I'm using `v0.4.0-dev`, which is probably why it works for me. I've updated the diff with syntax highlighting as per your suggestion and I've set some other options to mimic the `minimal style` option.

frescoraja: cool, thanks for taking the time to implement!

delphinus: Wait. `nvim_open_win()` can take `style` in options. It is documented in [HEAD][]. See `:h api-floatwin`

[HEAD]: https://github.com/neovim/neovim/blob/632ee8d2ed6e307cc4d9ca2cf9e3ba0531a8af51/runtime/doc/api.txt#L307

It does not appear on `:h nvim_open_api` as @frescoraja says, but on [src/nvim/api.c][src].

[src]: https://github.com/neovim/neovim/blob/632ee8d2ed6e307cc4d9ca2cf9e3ba0531a8af51/src/nvim/api/vim.c#L1072-L1081

Neovim v0.4.0 is still in development and documents are not enough. But `style` will be included in v0.4.0 release, I think, so it is better to use it.

georgijd: Aaah, that explains it then :). Thanks for digging @delphinus . We could have a branch with the `style` option set and merge it once v0.4.0 is released.

autoload/go/doc.vim
```diff
@@ -77,14 +77,45 @@ endfunction
...
+      call nvim_buf_set_lines(buf, 0, -1, v:true, lines)
+      let opts = {
+            \ 'relative': 'cursor',
+            \ 'row': 1,
+            \ 'col': 0,
+            \ 'width': width,
+            \ 'height': height,
+            \ }
+      call nvim_open_win(buf, v:true, opts)
+      setlocal nonumber norelativenumber nomodified nomodifiable syntax=go
```
bhcleek: `syntax=go` -> `filetype=godoc`

autoload/go/doc.vim
```diff
@@ -77,14 +77,45 @@ endfunction
...
+      let height = 0
+      let width = 0
+      for line in lines
+        let lw = strdisplaywidth(line)
+        if lw > width
+          let width = lw
+        endif
+        let height += 1
+      endfor
+      let width += 1 " right margin
```
bhcleek: Some kind of maximum is needed for height so that large godoc doesn't take up the whole screen.

bhcleek:
LGTM

Thank you for contributing!

frescoraja:
@georgijd thanks man ! now we just need to make `:GoDoc` async operation ;)

PR #2455, opened 2019-08-15
Some new UltiSnips

andrewstuart:
I keep writing these and figure they're probably useful for others (I use them all the time), so figured I might as well make a PR. :-)

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+patreon: bhcleek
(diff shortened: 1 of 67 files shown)


bhcleek:
Thank you for contributing again. We try to keep the snippets that vim-go provides limited to those that only depend on the standard library; several of those that you provided here will be useful, and I'd like to integrate them. We also try to keep the snippets consistent for all the engines that vim-go supports. Would you be willing to refactor this PR to:
1. limit the snippets to those that only make use of the standard library
2. provide analogous snippets for the other two snippets engines that vim-go supports?

If not, that's ok, too, and I'll eventually merge this PR locally and refactor accordingly.

bhcleek:
Thank you @andrewstuart. Since this PR has continued to have things added to it, I've extracted the snippets that will only require the standard library, made some small modifications, and created #2623.

PR #2457, opened 2019-08-18
plugin: reset more while installing binaries

bhcleek:
Reset 'more' while installing binaries. Set it back to its original
value before returning. If necessary, Vim will only prompt the user for
more once, at the end of the process instead of multiple times during.

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -136,6 +136,9 @@ function! s:GoInstallBinaries(updateBinaries, ...)
     let l:platform = 'windows'
   endif
 
+  let l:oldmore = &more
+  let &more = 0
+
   for [l:binary, l:pkg] in items(l:packages)
     let l:importPath = l:pkg[0]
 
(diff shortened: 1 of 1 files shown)


arp242:
You should really use `try .. finally` for this kind of stuff, as returning or errors inside the function means the reset will never be called. Same applies for some other stuff that gets reset in this function.

bhcleek:
True. I made this decision deliberately, because there are no returns between the new statements and this code is getting a refactor soon anyway.

arp242:
All these external commands – many of which access the network – are very apt to failure, so returning early is probably happening for some people.

Why "deliberately" write code with a predisposition for buggy behaviour if it's just as much effort to do it in a non-buggy way? I don't understand this at all... It's not a massive refactor we're talking about here, just a wrapping in a try..finally block. In a previous change you put `Restore_modules` just outside of an existing `try` block ... :-/

PR #2458, opened 2019-08-18
use go#util#Echo* functions consistently

bhcleek:
##### complete: remove debugging message

Remove an echo statement that only existed for debugging purposes.

##### debug: use go#util#EchoInfo for debugging message

replace echom with go#util#EchoInfo for debugging message.

##### package: use go#util#EchoError instead of echomsg

Replace echomsg with go#util#EchoError for displaying an error message.

##### play: use go#util#Echo* instead of echom or echo

Replaces uses of echomsg and echo with the appropriate go#util#Echo*
message for consistency.

##### plugin: use go#util#Echo* functions

Use go#util#Echo* functions instead of calling echom directly.

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -90,9 +90,7 @@ function! s:GoInstallBinaries(updateBinaries, ...)
   endif
 
   if go#path#Default() == ""
-    echohl Error
-    echomsg "vim.go: $GOPATH is not set and 'go env GOPATH' returns empty"
-    echohl None
+    call go#util#EchoError('$GOPATH is not set and `go env GOPATH` returns empty')
     return
   endif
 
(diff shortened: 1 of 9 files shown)


plugin/go.vim
```diff
@@ -26,14 +26,14 @@ function! s:checkVersion() abort
...
     if l:unsupported == 1
-      echohl Error
-      echom "vim-go requires Vim 7.4.2009 or Neovim 0.3.2, but you're using an older version."
-      echom "Please update your Vim for the best vim-go experience."
-      echom "If you really want to continue you can set this to make the error go away:"
-      echom "    let g:go_version_warning = 0"
-      echom "Note that some features may error out or behave incorrectly."
-      echom "Please do not report bugs unless you're using Vim 7.4.2009 or newer or Neovim 0.3.2."
-      echohl None
+      call go#util#EchoError([
```
arp242: This used echom on purpose since older Vim version may error out on autoloading the `autoload/go/util.vim` file and/or executing the `EchoError` function.

bhcleek: Again, I changed this deliberately. If it errors trying to load `autoload/go/util.vim`, then things are messed up pretty badly anyway likely the `go#config#VersionWarning` already failed.

arp242: Probably a good reason to not use `go#config#VersionWarning()` here, too.

I didn't do this for the fun of it, but in response to real errors. The entire reason this exists is to present a nice "your Vim is too old" message instead of confusing parse errors and the like.

bhcleek: yeah, fair. I'll revert this bit.

PR #2464, opened 2019-08-28
fix: stepOut typo to stepout

xbsura:
which cause go-debug-stepout not work


bhcleek:
Thank you for contributing.

PR #2472, opened 2019-09-04
Bugfix - Job Fails Silently When Path Contains Spaces

Rican7:
# Identification

I noticed that certain commands, such as _all_ of the `GoTest` variants (`GoTest`, `GoTestCompile`, `GoTestFunc`, etc) weren't working for only **some** of my projects. Interestingly, when they failed, they failed silently, with no logs, messages, or errors.

After trying these commands in other projects, some in module mode, some in my GOPATH, etc, I realized that the consistent issue was for paths that had certain characters, such as spaces.

I narrowed the issue down to an over-eager escaping of a file-path in the `go#job#Options` function:

https://github.com/fatih/vim-go/blob/c6273c5ca6a67f02bd6afdba657194253a0e5d85/autoload/go/job.vim#L64

... which was causing the early return due to the condition that intends to squelch error messages for otherwise "virtual" files:

https://github.com/fatih/vim-go/blob/c6273c5ca6a67f02bd6afdba657194253a0e5d85/autoload/go/job.vim#L286-L287

I then noticed that not only was the escaping eager, but it was inconsistent with the later attempt to grab the same path, if one wasn't provided:

https://github.com/fatih/vim-go/blob/c6273c5ca6a67f02bd6afdba657194253a0e5d85/autoload/go/job.vim#L285

# Fix

This PR fixes the over-eager escaping of the file-path of the job directory in `go#job#Options` that caused paths with certain characters and spaces to fail (silently) to be able to execute a job.

I also made sure to properly escape the path later on, where it was necessary, as its potentially used in a raw command later on that would require this level of escaping.

### DIFF
--- a/autoload/go/job.vim
+++ b/autoload/go/job.vim
@@ -61,7 +61,7 @@ function! go#job#Options(args)
   let state = {
         \ 'winid': win_getid(winnr()),
         \ 'dir': getcwd(),
-        \ 'jobdir': fnameescape(expand("%:p:h")),
+        \ 'jobdir': expand("%:p:h"),
         \ 'messages': [],
         \ 'bang': 0,
         \ 'for': "_job",
@@ -195,7 +195,7 @@ function! go#job#Options(args)
     let l:cd = exists('*haslocaldir') && haslocaldir() ? 'lcd' : 'cd'
     try
       " parse the errors relative to self.jobdir
-      execute l:cd self.jobdir
+      execute l:cd fnameescape(self.jobdir)
       call go#list#ParseFormat(l:listtype, self.errorformat, out, self.for)
       let errors = go#list#Get(l:listtype)
     finally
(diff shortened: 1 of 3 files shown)


bhcleek:
Thank you for the fix! I'll take a close look and probably add some tests before merging. Do you mind if I push commits that add tests to your branch?

Rican7:
> Thank you for the fix!

Wow! You're quick! No problem!

I've actually been plagued by this issue for months, and just couldn't figure out what the problem was, and finally decided to spend the time to look into it. 😅

> I'll take a close look and probably add some tests before merging. Do you mind if I push commits that add tests to your branch?

That would be great! I wouldn't mind at all!

I'm sorry that I didn't include tests myself. I usually **never** open a PR without including tests, but I couldn't think of a good way to actually test running in a path with spaces.

(Also, I've never actually written a test in VimL 😆... I'm still new to the plugin contribution world.)

Rican7:
I appreciate your hard-work on this project!

Thanks again!

image

Rican7:
Thanks again, @bhcleek!! 🙌

Issue #2470, opened 2019-09-03
syntax/gomod.vim: add syntax for "go 1.12" line


dolmen:
### What did you do? (required: The issue will be **closed** when not provided)

When editing `go.mod`, the "go 1.12" line is not colored

### What did you expect to happen?

The `go` keyword should be colored like `module` or `require`.

### What happened instead?

The `go` keyword is not colored like `module` or `require`.

#### vim-go version:

1.20 (but no change on syntax/gomod.vim on master)

#### Go version (`go version`):
1.12

PR #2473, opened 2019-09-04
syntax: highlight go directive and version

carnott-snap:
Go1.12 started writing the go directive and language version into go.mod
files. These were not previously highlighted. The go keyword is only
highlighted at top level, not in import paths, and the version assumes
go2 will never be released.


bhcleek:
Fixes #2470

Issue #2489, opened 2019-09-12
drop support for gometalinter


bhcleek:
gometalinter is deprecated and archived; it's time for vim-go to fully drop support for it.

rbisewski:
If no one else is working on this, I'll see if I can make a pull request later today.

PR #2494, opened 2019-09-16
Dropping support for the depreciated gometalinter tool

rbisewski:
As per issue #2489, since the gometalinter tool is deprecated and archived, I think a number of references to the gometalinter tool, as well as a number of lines of dedicated logic, can be removed. 

I believe additional logic could be taken out (see the TODO added to line 194 in the `plugin/go.vim` file), however, in the interest of keeping the changeset minimal and to avoid possible regressions, I have been conservative in the amount of lines deleted.

Also probably some of the documentation could be adjusted to mention that primary supported meta-linting tool is [golangci-lint](https://github.com/golangci/golangci-lint), but likely that can be left for a future ticket.

Let me know what you think or whether I overlooked anything :)

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -57,7 +57,6 @@ let s:packages = {
       \ 'goimports':     ['golang.org/x/tools/cmd/goimports'],
       \ 'golint':        ['golang.org/x/lint/golint'],
       \ 'gopls':         ['golang.org/x/tools/gopls@latest', {}, {'after': function('go#lsp#Restart', [])}],
-      \ 'gometalinter':  ['github.com/alecthomas/gometalinter'],
       \ 'golangci-lint': ['github.com/golangci/golangci-lint/cmd/golangci-lint'],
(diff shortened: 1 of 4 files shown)


plugin/go.vim
```diff
@@ -192,6 +191,8 @@ function! s:GoInstallBinaries(updateBinaries, ...)
         endif
 
         " GO111MODULE must be off to install gometalinter.
+        " TODO: determine if this line is still needed now that gometalinter
+        " has been depreciated
```
bhcleek: Please leave this line in, but we can do with out the TODO _and_ the preceding line now.

rbisewski: You mean delete my TODO line and the original `GO111MODULE` comment line? Done.

rbisewski:
Thanks for reviewing my work :)

I have since made the changes requested, and while the linter tests pass, I see that the individual vim version tests fail.

I'll take a quick look into this and see if it's an easy fix...

rbisewski:
Tests work again and docs are updated.

Let me know what you think...

Issue #2488, opened 2019-09-12
drop support for Vim 7.4


bhcleek:
Drop support for Vim 7.4

- [ ] update build scripts
- [ ] update warning messages
- [ ] update checks for Vim 7.4

rbisewski:
I'll see if I can make a pull request soon enough.

PR #2495, opened 2019-09-16
Removing logic needed to support Vim 7.4 as it has been dropped.

rbisewski:
As per issue #2488, this branch removes Vim 7.4 support by deleting a number of checks for features that the Vim 7.4 series did not support. This also adjusts the tests and the Makefile, as well as small tweaks to the documentation in order to clarify that Vim 8.0.1542 or newer is the supported version; with that particular version of Vim 8 being chosen on account of its prior use in the `scripts/install-vim` file.

Let me know what you think...

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -9,30 +9,22 @@ let s:cpo_save = &cpo
 set cpo&vim
 
 function! s:checkVersion() abort
-  " Not using the has('patch-7.4.2009') syntax because that wasn't added until
-  " 7.4.237, and we want to be sure this works for everyone (this is also why
-  " we're not using utils#EchoError()).
-  "
-  " Version 7.4.2009 was chosen because that's greater than what the most recent Ubuntu LTS
-  " release (16.04) uses and has a couple of features we need (e.g. execute()
-  " and :message clear).
-
   let l:unsupported = 0
   if go#config#VersionWarning() != 0
     if has('nvim')
       let l:unsupported = !has('nvim-0.3.2')
     else
-      let l:unsupported = (v:version < 704 || (v:version == 704 && !has('patch2009')))
+      let l:unsupported = v:version < 800
     endif
 
     if l:unsupported == 1
       echohl Error
-      echom "vim-go requires Vim 7.4.2009 or Neovim 0.3.2, but you're using an older version."
+      echom "vim-go requires Vim 8+ or Neovim 0.3.2, but you're using an older version."
       echom "Please update your Vim for the best vim-go experience."
       echom "If you really want to continue you can set this to make the error go away:"
       echom "    let g:go_version_warning = 0"
       echom "Note that some features may error out or behave incorrectly."
-      echom "Please do not report bugs unless you're using Vim 7.4.2009 or newer or Neovim 0.3.2."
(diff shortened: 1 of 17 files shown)


rbisewski:
Sounds good, and glad I could help :)

So for context you mean you want code reverted?

* `autoload/go/cmd.vim`
    * revert all changes

* `autoload/go/coverage.vim`
    * revert the`if !exists("*matchaddpos")` lines

* `autoload/go/guru.vim`
    * revert all the `if !exists("*json_decode")` and `if !exists("*matchaddpos")` lines

I suppose I could simplify the comments on those above lines as well to remove references to the Vim 7.4 and the various commit links since I figure they are less important as Vim 8 / NeoVim supports more features by default.

Unless you'd prefer to keep them for historic value?

bhcleek:
> So for context you mean you want code reverted?

Yes. Your list of things to revert is exactly what I was thinking.

> I suppose I could simplify the comments on those above lines as well to remove references to the Vim 7.4 and the various commit links since I figure they are less important as Vim 8 / NeoVim supports more features by default.
> 
> Unless you'd prefer to keep them for historic value?

Simplifying those comments make sense to me, but it's not essential; I'll leave it to your discretion whether to leave them as-is or simplify them.

rbisewski:
I have since restored the code and simplified the comments and debug output messages, so I believe this is done now.

plugin/go.vim
```diff
@@ -9,30 +9,22 @@ let s:cpo_save = &cpo
...
-  " and :message clear).
-
   let l:unsupported = 0
   if go#config#VersionWarning() != 0
     if has('nvim')
       let l:unsupported = !has('nvim-0.3.2')
     else
-      let l:unsupported = (v:version < 704 || (v:version == 704 && !has('patch2009')))
+      let l:unsupported = v:version < 800
     endif
```
arp242: Checking for just "Vim 8" isn't really enough, since it spans almost 2 years of development. You should pick a specific patch version like it did previously.

In the documentation above `8.0.1542` was chosen; why that version? The previous policy was always "what the latest Ubuntu LTS has", which is now Vim 8.0.1453 (18.04).

arp242: Oh, I see now that you probably choose 8.0.1542 because that's in the Travis file? It has no special meaning, it just happened to be the version installed on my Arch laptop when I wrote that.

bhcleek: We needed to pick something, and Vim8 is sufficient for now.

arp242: So will Vim 8,0.0001 work? Probably not, since: https://github.com/fatih/vim-go/search?q=patch-8.0&unscoped_q=patch-8.0

These patch checks should probably also be removed, by the way. I don't know why it's such a fight to get all these details right every time... It's not even harder or anything :man_shrugging:

bhcleek: My thinking here is that previously vim-go checked that it was at least Vim 7.4.2009. All we're doing is dropping support for Vim 7.4. Another patch to make sure it's at least 8.0.1453 makes sense. I'll add that later and probably drop the unneeded patch checks at the same time (if you're not motivated to submit it first), but didn't see any need to hold up the improvements that @rbisewski made.

Issue #2483, opened 2019-09-09
format with gopls


bhcleek:
Add support for formatting with gopls instead of `gofmt` or `goimports`.

##### acceptance criteria
* when `g:go_fmt_command` is `gopls` vim-go uses `gopls` to format Go source files
* when `g:go_fmt_command` is set to `gopls`, `gopls` should be used regardless of the value of  `g:go_fmt_experimental`.
* `g:go_fmt_experimental` should be respected regardless of the value of `g:go_fmt_command`.

rbisewski:
I'll take a shot at doing this, if nobody else is.

bhcleek:
This will be a bit tricky to get right, but the basics are that `gopls` will provide a list of changes to make. Those changes need to be applied starting in reverse order (i.e. starting at the last line), because of the way the LSP protocol defines them.

Let me know if you have any questions or need a hand.

rbisewski:
So you think a simple `gopls format -w /path/to/file.go` and some new tests won't suffice? I've been testing this on my own code for the last day or so and it seems to format correctly; to my knowledge anyway, I will admit I am far from a gopls expert.

If it's alright with you, I have pushed my changes to a branch, so I'll make a pull request in a few minutes and we can continue the discussion there :)

bhcleek:
Yeah, as I mentioned on the PR, `gopls format` doesn't add any value over `gofmt`, because they both use the same underlying packages to do the formatting. However, `gopls` regularly outputs a  formatted file.  Basically, any time it's sent the `document/didChange` or `document/didOpen` request, it will send a diagnostic with the formatted output.

As I've been thinking about it more, I'm starting to lean towards:
1. leaving `go_fmt_command` as-is so that `goimports` can be used on save, because `gopls` doesn't add imports.
2. Introduce a new config value, `g:go_gopls_fmt_enabled`, that controls whether the formatting provided by `gopls` automatically updates. We might need to modify the `workspace/didChange` integration to use incremental updates.
3. When `g:go_gopls_fmt_enabled` is set, update formatting as the user types (maybe on pressing Enter?)

I think it's worth closing the #2499 and collaborating to define how this should work. What do you think?

rbisewski:
Agreed and yeah, `gopls` doesn't add imports, so probably your suggestion is the most straightforward way it could be done :)

bhcleek:
FWIW, I think that `gopls` will add imports, but `gopls format` does not.

bhcleek:
[...]

PR #2499, opened 2019-09-18
Implement logic to set gopls as a formatting tool for vim-go.

rbisewski:
As per #2483, this offers the ability of the end-user to set a variable like the below in their vimrc file to enable `gopls` code formatting:

```
let g:go_fmt_command = "gopls"
```

Essentially all this does is change the command used to format code into something like:

```
gopls format -w /path/to/whatever/file/you/are/editting.go
```

I have also included a test case, similar to how there are tests for `gofmt` and `goimports` and it seems to pass. I have also been testing this on my own go code and have yet to experience any problems. However, testing from others would also be welcome.

All of this said, the topic of the issue seems to mention other variables, so please let me know if I am overlooking something obvious :)

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -3,7 +3,7 @@
 " license that can be found in the LICENSE file.
 "
 " fmt.vim: Vim command to format Go files with gofmt (and gofmt compatible
-" toorls, such as goimports).
+" tools, such as goimports).
 
 " don't spam the user when Vim is started in Vi compatibility mode
 let s:cpo_save = &cpo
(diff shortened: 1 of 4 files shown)


bhcleek:
`gopls format` doesn't add any value over `gofmt`. #2483 is really about doing dynamic formatting without requiring the file to be saved or perhaps using the output from `gopls` when it prints its diagnostic on save.

Since `gopls` is already running, there shouldn't be any need to run `gopls format`, but the lsp integration will need to be enhanced to process the diagnostics from gopls.

We can start by using the diagnostic message for a file after saving instead of trying to do it continuously.

rbisewski:
Closing...

PR #2506, opened 2019-09-20
Add documentation for go_doc_popup_window

ian-howell:


bhcleek:
Thank you for adding to the docs. CI failed for two reasons, only one of which is legitimate: the help linter we use doesn't know about the popup tag (perhaps because it runs using Vim 8.0, and popup wasn't introduced until 8.1). I don't see a good way to change the linter to ignore that tag 🤔 

Can you remove the popup tag and also change the `let g:go_doc_popup_window = 1` to `let g:go_doc_popup_window = 0`, because vim-go's help documents the default values of config items.

ian-howell:
I botched some git stuff while fixing this, so ignore any stray travis builds. The changes you requested have been implemented.

bhcleek:
Thank you for the contribution

PR #2512, opened 2019-09-25
Use `minimal` style for GoDoc popup windows in Neovim

delphinus:
This patch uses “minimal” style for floating windows to introduce more readability on `:GoDoc`.

See [:h nvim_open_win()](https://neovim.io/doc/user/api.html#nvim_open_win()).
See https://github.com/fatih/vim-go/pull/2451#discussion_r321139543

### DIFF
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -111,9 +111,10 @@ function! s:GodocView(newposition, position, content) abort
             \ 'col': 0,
             \ 'width': width,
             \ 'height': height,
+            \ 'style': 'minimal',
             \ }
       call nvim_open_win(buf, v:true, opts)
-      setlocal nonumber norelativenumber nomodified nomodifiable filetype=godoc
+      setlocal nomodified nomodifiable filetype=godoc
 
       " close easily with CR, Esc and q
       noremap <buffer> <silent> <CR> :<C-U>close<CR>


bhcleek:
It sounds like `style` was introduced sometime after `nvim_open_win`. If that's the case, then this won't work for all the versions of Neovim that vim-go users may be running.

delphinus:
@bhcleek Neovim v0.4.0 introduced `style: minimal` at the same time as `nvim_open_win()` itself ([link to v0.4.0 code][link]). The code in #2451 was written before developers in Neovim decided the final spec of `nvim_open_win()`.

So when they come across errors for this PR, it means they are using development versions of Neovim. We should write the code for the released spec, not for its development version, I think.

[link]: https://github.com/neovim/neovim/blob/v0.4.0/src/nvim/api/vim.c#L1072-L1081

bhcleek:
Thanks for the clarification on the timing of the introduction of `minimal` and `nvim_open_win`.  It sound like it makes `setlocal nonumber norelativenumber` redundant, though. Am I reading that right?

delphinus:
Yes. In addition to them, `style: minimal` has more meanings. (I noticed this I saw the chars by `set list`.)

```vim
                              •  `style` : Configure the appearance of the window.
                                Currently only takes one non-empty value:
                                • "minimal" Nvim will display the window with
                                  many UI options disabled. This is useful
                                  when displaying a temporary float where the
                                  text should not be edited. Disables
                                  'number', 'relativenumber', 'cursorline',
                                  'cursorcolumn', 'foldcolumn', 'spell' and
                                  'list' options. 'signcolumn' is changed to
                                  `auto` . The end-of-buffer region is hidden
                                  by setting `eob` flag of 'fillchars' to a
                                  space char, and clearing the |EndOfBuffer|
                                  region in 'winhighlight'.
```

bhcleek:
In that case, can you remove the additions to the `setlocal` line that you added?

delphinus:
I myself did not add it. `setlocal nomodified nomodifiable` derives from #2451.

These two settings are effective even with `style: minimal` and are not necessary. But they are useful to avoid unintended behaviors, don't you think?

bhcleek:
Weird. It's showing up as an addition in this PR. 🤷‍♂ oh well. I'll merge this soon.

PR #2521, opened 2019-09-28
wrap SetEnv's undo in try/catch

bhcleek:


bhcleek:
closing in favor of two PRs thats split this work into the two concerns addressed here.

PR #2529, opened 2019-10-01
Fix #2524: Use own interval to call AutoTypeInfo/AutoSameids

### LINKED ISSUE #2524: Suggestion: Use separated interval for AutoTypeInfo/AutoSameids other than updatetime
### What did you do?

1. Prepare vimrc

   ```vim
   set runtimepath+=~/test/vim-go
   ```

2. Launch (Neo)Vim with clean env.

   ```sh
   nvim -N -u vimrc  -i NONE new_file.txt
   ```

2. You can see the default value set to `updatetime`.

   ```vim
   :set updatetime?
   4000
   ```

3. Open `*.go` files in a new buffer.

   ```vim
   :sp another.go
   ```

4. You can see the value from `vim-go` set to `updatetime`.

   ```vim
   :set updatetime?
   800
   ```

5. Press `<C-w>w` and return to the original window: `new_file.txt`.

6. You can see `updatetime` set to the value from `vim-go`!

   ```vim
   :set updatetime?
   800
   ```

### What is the problem?

#1055 introduced the `g:go_updatetime` variable and @svanharmelen should have intended to limit the scope of `updatetime` into a buffer-local one.

But `updatetime` is a global option. You cannot use it as buffer-local.

### Suggestion

I think this is not the case to use `updatetime`. The option is used for many other plugins and `vim-go` should not depend on it.

I suggest to rewrite the logic with the `timer` feature.

1. Call func on `CursorMoved`.
2. In the func, call `timer_start()` and wait some interval.
3. After interval, do the job (ex. `AutoTypeInfo`, `AutoSameids`)

(I wrote a tiny plugin with using similar logic: [vim-auto-cursorline](https://github.com/delphinus/vim-auto-cursorline).)

I will write PR if you approve this plan.

#### vim-go version:

e50c575

####  `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):

(see above)

#### Vim version (first three lines from `:version`):

```
NVIM v0.4.2
Build type: Release
LuaJIT 2.0.5
```

delphinus:
See #2524 

----

now developing

### TODO

* [x] Deal with vim74 that has no timer support?

### DIFF
--- a/autoload/go/auto.vim
+++ b/autoload/go/auto.vim
@@ -34,21 +34,23 @@ endfunction
 
 let s:timer_id = 0
 
-function! go#auto#cursor_moved()
-  if (!go#config#AutoTypeInfo() && !go#config#AutoSameids()) || !isdirectory(expand('%:p:h'))
-    return
+function! go#auto#update_autocmd()
+  execute 'augroup vim-go-buffer-auto-' . bufnr()
+    autocmd! * <buffer>
+    if go#config#AutoTypeInfo() || go#config#AutoSameids()
+      autocmd CursorMoved <buffer> call s:cursor_moved()
+      autocmd BufLeave <buffer> call s:timer_stop()
+    else
+      call s:timer_stop()
+    endif
+  augroup END
+endfunction
+
+function! s:cursor_moved()
+  if isdirectory(expand('%:p:h'))
+    call s:timer_stop()
+    call s:timer_start()
   endif
-
-  call s:timer_stop()
-  call s:timer_start()
-endfunction
-
-function! go#auto#win_enter()
-  call s:timer_stop()
-endfunction
-
-function! go#auto#win_leave()
-  call s:timer_stop()
 endfunction
 
 function! s:timer_stop()
--- a/autoload/go/auto.vim
+++ b/autoload/go/auto.vim
@@ -33,7 +33,7 @@ function! go#auto#echo_go_info()
 endfunction
 
 function! go#auto#update_autocmd()
-  execute 'augroup vim-go-buffer-auto-' . bufnr()
+  execute 'augroup vim-go-buffer-auto-' . bufnr('')
     autocmd! * <buffer>
     if go#config#AutoTypeInfo() || go#config#AutoSameids()
       autocmd CursorMoved <buffer> call go#auto#timer_restart()
--- a/autoload/go/auto.vim
+++ b/autoload/go/auto.vim
@@ -32,22 +32,45 @@ function! go#auto#echo_go_info()
   redraws! | echo "vim-go: " | echohl Function | echon item.info | echohl None
 endfunction
 
-function! go#auto#auto_type_info()
-  if !go#config#AutoTypeInfo() || !isdirectory(expand('%:p:h'))
+let s:timer_id = 0
+
+function! go#auto#cursor_moved()
+  if (!go#config#AutoTypeInfo() && !go#config#AutoSameids()) || !isdirectory(expand('%:p:h'))
     return
   endif
 
-  " GoInfo automatic update
-  call go#tool#Info(0)
+  call s:timer_stop()
+  call s:timer_start()
 endfunction
 
-function! go#auto#auto_sameids()
-  if !go#config#AutoSameids() || !isdirectory(expand('%:p:h'))
-    return
-  endif
+function! go#auto#win_enter()
+  call s:timer_stop()
+endfunction
 
-  " GoSameId automatic update
-  call go#guru#SameIds(0)
+function! go#auto#win_leave()
+  call s:timer_stop()
+endfunction
+
+function! s:timer_stop()
+  if s:timer_id
+    call timer_stop(s:timer_id)
+    let s:timer_id = 0
+  endif
+endfunction
+
+function! s:timer_start()
+  let time = get(g:, 'go_updatetime', 800)
+  let s:timer_id = timer_start(time, function('s:handler'))
+endfunction
+
+function! s:handler(timer_id)
+  if go#config#AutoTypeInfo()
+    call go#tool#Info(0)
+  endif
+  if go#config#AutoSameids()
+    call go#guru#SameIds(0)
+  endif
+  let s:timer_id = 0
 endfunction
 
 function! go#auto#fmt_autosave()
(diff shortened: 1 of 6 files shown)


bhcleek:
no need to worry about Vim 7.4. vim-go no longer supports it.

bhcleek:
Thank you for contributing. This looks great.

Are you waiting for something specific before promoting it from a draft?

delphinus:
@bhcleek

Nope, sorry for late response. I have been considering one problem.

This patch makes it start on `CursorMove` event. This occurs very frequently, so it calls `go#auto#cursor_moved()` many times. This func is lightweight and almost users do not notice any delay. But it should not be called at all when users have disabled AutoTypeInfo/AutoSameids.

So I am planning to implement with a strategy below.

* Set `autocmd CursorMove` in `plugin/go.vim` only when `g:go_auto_type_info == 1` or `g:go_auto_sameids == 1`.
* Set / unset the `autocmd` when either `:GoAutoTypeInfoToggle` or `:GoSameIdsAutoToggle` is called.

Perhaps this works good, but introduces breaking changes.

<dl>
<dt>Before</dt>
<dd>It toggles the features immediately when either value of <code>g:go_auto_type_info</code> / <code>g:go_auto_sameids</code> is changed.</dd>
<dt>After</dt>
<dd>It does nothing when either value of <code>g:go_auto_type_info</code> / <code>g:go_auto_sameids</code> is changed. Users must call <code>:GoAutoTypeInfoToggle</code> / <code>:GoSameIdsAutoToggle</code> to toggle the features.</dd>
</dl>

I think users rarely change the values for the purpose of toggling the features, so this is a permissiable change. How do you think?

bhcleek:
The performance of this PR will definitely be key to getting it merged.

Instead of only checking at startup time, perhaps use `CursorHold` to check `g:go_auto_type_info` and `g:go_auto_sameids` and hook up the autocmd if it was previously off and the variables are now set.

delphinus:
It seems good! OK, I try to implement with both `CursorHoold` and `CursorMove` events.

ftplugin/go.vim
```diff
@@ -74,9 +74,8 @@ if get(g:, "go_textobj_enabled", 1)
   xnoremap <buffer> <silent> [[ :<c-u>call go#textobj#FunctionJump('v', 'prev')<cr>
 endif
 
-if go#config#AutoTypeInfo() || go#config#AutoSameids()
-  let &l:updatetime= get(g:, "go_updatetime", 800)
-endif
+" Set up autocmd to call :GoSameIds / :GoInfo automatically
+call go#auto#update_autocmd()
```
bhcleek: Given that there's a CursorHold autocmd that calls `go#auto#update_autocmd`, is this call necessary?

autoload/go/auto.vim
```diff
@@ -32,22 +32,46 @@ function! go#auto#echo_go_info()
...
+    autocmd! * <buffer>
+    if go#config#AutoTypeInfo() || go#config#AutoSameids()
+      " vint: -ProhibitAutocmdWithNoGroup
+      autocmd CursorMoved <buffer> call go#auto#timer_restart()
+      autocmd BufLeave <buffer> call s:timer_stop()
+      " vint: +ProhibitAutocmdWithNoGroup
+    else
+      call s:timer_stop()
+    endif
+  augroup END
```
bhcleek: Since this function will be executed on `CursorHold` and to make sure that duplicate events aren't configured, shouldn't it _first_ check to see whether there's an existing timer, and then
1. return early if there isn't an existing timer and neither auto type info nor nor auto same ids is enabled
1. return early when there is a timer and either one or both of auto type info and auto same ids is enabled
1. clear the group and stop the timer when there is a timer and neither auto type info nor auto same ids is enabled
1. configure the group when there isn't a timer and either one or both of auto type info and auto same ids is enabled

?

autoload/go/auto.vim
```diff
@@ -32,22 +32,46 @@ function! go#auto#echo_go_info()
   redraws! | echo "vim-go: " | echohl Function | echon item.info | echohl None
 endfunction
 
-function! go#auto#auto_type_info()
-  if !go#config#AutoTypeInfo() || !isdirectory(expand('%:p:h'))
-    return
-  endif
+function! go#auto#update_autocmd()
+  execute 'augroup vim-go-buffer-auto-' . bufnr('')
```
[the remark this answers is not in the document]
delphinus: That's right. I tested and recognized `vim-go-suffer-auto` is just enough. Fixed on bb1d460

ftplugin/go.vim
delphinus: Not exactly because `BufEnter` does. I removed this line: 3b576be.

autoload/go/auto.vim
delphinus: Good point. It (re)defines augroup's redundantly. I fix it to do more effectively.

I arranged your suggestion in a table.

| # | has_timer | should_enable | do                                 |
|---|-----------|---------------|------------------------------------|
| 1 | false     | false         | return early                       |
| 2 | true      | true          | return early                       |
| 3 | true      | false         | clear the group and stop the timer |
| 4 | false     | true          | configure the group                |

Fixed on 1a5ff26

bhcleek:
This looks great. Thank you for the contribution!

I'll play with this for a couple of weeks before merging it.

bhcleek:
@delphinus this has been working great on my machine for a while now, and I'm planning to merge it soon. I'm taking a moment to step back, though, and ask if we really need it. Playing devil's advocate: using `updatetime` and `CursorHold` is sufficient, while this makes the code more complicated. What's the argument against using `updatetime` as you see it?

bhcleek:
The reason to not use `updatetime` was laid out pretty clearly in #1055. I'm going to merge this.

bhcleek:
Thank for all the hard work @delphinus

delphinus:
Yeah, I noticed this problem with using [airblade/vim-gitgutter][gg], that recommends `set updatetime=100` to refresh the gutters more frequently. To update AutoTypeInfo/AutoSameids less frequently than the gutters, this feature is quite useful for me, at least. ;)

[gg]: https://github.com/airblade/vim-gitgutter

PR #2531, opened 2019-10-02
added GoInit command

esoytekin:
when you create files using nerdtree, vim-go does not create file
template.

i found this command useful for these kinds of situations


bhcleek:

While I understand that this is useful for you, I don't think this adds enough value to include in vim-go. Users can add this simple command to an `after/ftplugin/go.vim` file to get this easily.

PR #2546, opened 2019-10-21
Escape backslashes

mattn:
Completion of `:GoImport` does not work since the path is not escaped.

### DIFF
--- a/autoload/go/package.vim
+++ b/autoload/go/package.vim
@@ -283,7 +283,7 @@ function! go#package#Complete(ArgLead, CmdLine, CursorPos) abort
         endif
 
         if dir !=# path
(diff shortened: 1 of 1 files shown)


bhcleek:
I'm having a hard time find a failure mode for the current implementation. How can I duplicate the problem that this PR solves?

mattn:
As far as I walked issues, there are no filed issues will be fixed by this PR.

bhcleek:
Have you seen a problem in the wild? I'm trying to figure out how to duplicate the problem that this would fix 🤔

bhcleek:
@mattn Is this trying to solve a problem on Windows?

Is `'\\'` necessary or can it be shortened to `'\'`?

Issue #2555, opened 2019-10-31
GoTest with go_term_mode enabled does not reset pwd


drewbailey:
### What did you do? (required: The issue will be **closed** when not provided)

Ran `:pwd` with autochdir off, output -> `/home/drew/go/src/github.com/hashicorp/nomad`

Open a test file `/home/drew/go/src/github.com/hashicorp/nomad/commanda/agent/monitor/monitor_test.go`
`:pwd` -> `/home/drew/go/src/github.com/hashicorp/nomad`

run `:GoFuncTest!`, close terminal split vim-go opens

`:pwd` -> `/home/drew/go/src/github.com/hashicorp/nomad/commanda/agent/monitor/monitor_test.go` instead of `/home/drew/go/src/github.com/hashicorp/nomad/`

### What did you expect to happen?

vim-go wouldn't change directory, or would revert back after test run

### What happened instead?

`:pwd` gets updated to test file

### Configuration (**MUST** fill this out):

```
let g:go_def_mode='gopls'
let g:go_info_mode='gopls'

let g:go_term_enabled = 1
let g:go_term_mode='vsplit'
let g:go_term_close_on_exit = 0
let g:go_list_autoclose = 1
let g:go_fmt_fail_silently = 0
let g:go_fmt_command = "goimports"
let g:go_list_type = "quickfix"
let g:go_list_type_commands={"GoTestFunc!": "", "GoTestFunc": ""}
let g:go_test_show_name=1
```

#### vim-go version: master

####  `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):
<details><summary>vimrc</summary><br><pre>
Plug 'fatih/vim-go', { 'do': ':GoUpdateBinaries' }
Plug 'neoclide/coc.nvim', { 'do': 'yarn install --frozen-lockfile' }

</pre></details>

#### Vim version (first three lines from `:version`):
<!-- :version -->
→ nvim --version                                                                                                                                                                                                                                                                       NVIM v0.3.4
Build type: Release
LuaJIT 2.1.0-beta3

#### Go version (`go version`):go version go1.13.3 linux/amd64
<!-- go version -->

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>
<!-- go env -->
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/drew/.cache/go-build"
GOENV="/home/drew/.config/go/env"
[...]

PR #2556, opened 2019-10-31
revert back to original pwd after GoTestFunc!

drewbailey:
fixes https://github.com/fatih/vim-go/issues/2555


drewbailey:
Sure thing, updated

PR #2557, opened 2019-10-31
Add syntax support for binary literals

romeovs:
Adds syntax support for binary literals, [introduced in go 1.3](https://golang.org/doc/go1.13#language).

These will literals will now be highlighted as integers:

```go
x := 0b1100
y := 0B0100
```

And these will be marked as errors for obvious reasons:

```
x := 0b1102
y := 0B8001
```

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -166,11 +166,15 @@ syn match       goDecimalInt        "\<-\=\d\+\%([Ee][-+]\=\d\+\)\=\>"
 syn match       goHexadecimalInt    "\<-\=0[xX]\x\+\>"
 syn match       goOctalInt          "\<-\=0\o\+\>"
 syn match       goOctalError        "\<-\=0\o*[89]\d*\>"
+syn match       goBinaryInt         "\<-\=0[bB][01]\+\>"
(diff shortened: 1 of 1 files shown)


syntax/go.vim
```diff
@@ -166,11 +166,15 @@ syn match       goDecimalInt        "\<-\=\d\+\%([Ee][-+]\=\d\+\)\=\>"
 syn match       goHexadecimalInt    "\<-\=0[xX]\x\+\>"
 syn match       goOctalInt          "\<-\=0\o\+\>"
 syn match       goOctalError        "\<-\=0\o*[89]\d*\>"
+syn match       goBinaryInt         "\<-\=0[bB][01]\+\>"
+syn match       goBinaryError       "\<-\=0[bB][01]*[2-9]\+[01]*\>"
```
bhcleek: Why does this match leading and trailing 0 or 1? Would `<-\=0[bB].*[^01]` be simpler and catch all the problems?

romeovs: Your regex does not match errors like `0b2201` AFAICT.

bhcleek: It does match `0b2201`: https://play.golang.org/p/QUm4VUWnF0e

PR #2572, opened 2019-11-10
add  highlights for GoDecls + fzf

110y:
This PR enables us to change highlight color for fzf GoDecls as same as `goSameId` like below:

```vim
highlight! goDeclsFzfComment ctermbg=196 ctermfg=255 cterm=BOLD guibg=#ff0000 guifg=#eeeeee gui=BOLD
```

Even though  we can also change highlight color by modifying `Function` etc for now, I'd like to change the color just only fzf + GoDecls.

### DIFF
--- a/autoload/fzf/decls.vim
+++ b/autoload/fzf/decls.vim
@@ -121,10 +121,10 @@ function! s:source(mode,...) abort
           \ decl.col
           \)
     call add(ret_decls, printf("%s\t%s %s\t%s",
-          \ s:color(decl.ident . space, "Function"),
-          \ s:color(decl.keyword, "Keyword"),
-          \ s:color(pos, "SpecialComment"),
-          \ s:color(decl.full, "Comment"),
+          \ s:color(decl.ident . space, "goDeclsFzfFunction"),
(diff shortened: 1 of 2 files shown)


autoload/fzf/decls.vim
```diff
@@ -121,10 +121,10 @@ function! s:source(mode,...) abort
           \ decl.col
           \)
     call add(ret_decls, printf("%s\t%s %s\t%s",
-          \ s:color(decl.ident . space, "Function"),
-          \ s:color(decl.keyword, "Keyword"),
-          \ s:color(pos, "SpecialComment"),
-          \ s:color(decl.full, "Comment"),
+          \ s:color(decl.ident . space, "goDeclsFzfKeyword"),
+          \ s:color(decl.keyword, "goDeclsFzfFunction"),
```
bhcleek: `dec.ident` and `decl.keyword`'s  expected highlight groups are swapped.

110y:
@bhcleek 

Ah... Thank you for pointing it out 🙏 
I've fixed.

PR #2574, opened 2019-11-11
Fix regex for structs when expanding vars in debug

georgijd:
- Allow pointers.
- Allow hyphens, underscores and forward slashes in struct names as
they might be coming from external packages.
- Make the `{...}` bit optional as it doesn't always appear.

### DIFF
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -326,10 +326,10 @@ endfunction
 
 function! s:expand_var() abort
   " Get name from struct line.
-  let name = matchstr(getline('.'), '^[^:]\+\ze: [a-zA-Z0-9\.·]\+{\.\.\.}$')
+  let name = matchstr(getline('.'), '^[^:]\+\ze: \*\?[a-zA-Z0-9-_/\.]\+\({\.\.\.}\)\?$')
   " Anonymous struct
   if name == ''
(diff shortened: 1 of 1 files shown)


bhcleek:
Thanks for contributing again.

Hyphens aren't allowed in Go identifiers.  Would you mind removing it from the regex?

georgijd:
Hi, no problem.

I added the hyphens, because they can be a part of the package path, e.g.

```
variable: *github.com/georgijd/some-go-project/path-to-package/package.SomeStruct
```

bhcleek:
That makes sense. We can leave it as-is, and I'll test and merge soon.

bhcleek:
This works great. Thank you @georgijd. It looks like the other sections in `expand_vars` could be extended to handle package names and pointers more fully, too. If you want to contribute those, I can hold off merging this for now, or I can merge this now and those can be done later.

georgijd:
Yeah, there is some more work for sure, expanding array struct fields doesn't seem to work very well either. Can we merge this and I'll give the other sections a go after :) Also, I'll try adding some tests although my vim script knowledge is limited. Thanks!

bhcleek:
Sounds good to me. I can help with the tests either by creating one for this change or by code reviews if you just want to give it a shot before seeing an example.

PR #2584, opened 2019-11-13
syntax: Fix hex and octal int and errors highlight

k33nice:
According to language specification it is allowed to use underscore "_"
for readability in integer literals. Hence syntax matches for
goHexadecimalInt and goOctalInt must allow it. Also goOctalError regexp
mustn't match on accessing slice or array element by index.


bhcleek:
This highlights valid hexadecimals `goOctalError`.

bhcleek:
FWIW, I added some tests for verifying the syntax highlight groups of the numeric literals: https://github.com/fatih/vim-go/pull/2587.  You can rebase your branch off of master and use the tests to make sure your changes match as expected.

bhcleek:
This does not highlight `07a` as `goOctalError`.

k33nice:
Hey @bcarrell! I'm currently trying to run the tests with docker but get some errors from gopls :( 
About your question on `07a` - no, it doesn't highlight it as an error right now. I tried to explain it in the commit message:

```
    The regular expression was changed that way to match all wrong sequences
    of characters. To do not make a false matching on other language structures
    the regexp should detect only falsy octal literals based on
    the allowed symbols: [0-7], "o", "O", "_". The exclusion is "8" and "9"
    because it might be very easy to make that kind of mistake.

    The regexp breakdown is described below:

    0                  |<-- "0" digit
    \(                 |
        [0-7oO_]*      |<-- 0+ [0-7] or "o" or "O" or "_" -----+
        \(             |                                       |
            [89]\+     |<-- 1+ "8" or "9" digit ----+          |
                \|     |                            |   +---+  |   +---+
            [oO]\{2,\} |<-- 2+ "o" or "O" character |<--| 1 |  |<--| 2 |
                \|     |                            |   +---+  |   +---+
            _\{2,\}    |<-- 2+ "_" character -------+          |
        \)             |                                       |
        [0-7oO_]*      |<-- 0+ [0-7] or "o" or "O" or "_" -----+
    \)\+               |
    \S*                |<-- 0+ "non space" characters

    1. One of the rules in this group should appear exactly once.
    2. All of the rules in this group should appear once or more.
```

TBO, as a user of `vim-go` I'm not sure what kind of sequences I want to be treated as an octal error. What do you think about it?

k33nice:
I can change `[89]\+` to `[89A-Fa-f]\+` so that kind of sequences like `07a` will be highlighted as `goOctalError`. But I'm wondering is that should be `goOctalError` or `goHexadecimalError` 🤔

k33nice:
I've also added some changes to other regexp here https://github.com/fatih/vim-go/pull/2584/commits/809c94ce044b35ff1c6478c3e2a157bb30933cff related to https://github.com/fatih/vim-go/issues/2588

bhcleek:
@k33nice you can ignore the gopls errors; they're due to some kind of regression in gopls that is already being addressed by the Go tools team.

Generally speaking, the regular expression changes you're making are going to fail to highlight some patterns that are correctly highlighted on master. Here's a few examples:
```
syn match       goDecimalInt        "\<-\=\d\+\%([Ee][-+]\=\d\+\)\=\>"
```

This is now going to regress and match as a decimal instead of an octdal when the number begins with a 0.

```
syn match       goOctalError        "\<-\=0\([0-7oO_]*\([89]\+\|[oO]\{2,\}\|_\{2,\}\)[0-7oO_]*\)\+\S*\>"
```

This treats 8 and 9 specially for some reason when there's really no need to. The scope of matching invalid characters should be much wider - as it already is on master.

It also introduces an inconsistency in how goOctalError handle more than one `_` vs how goBinaryError and goHexadecimalError handle them.

I'd also strongly recommend that you rebase your branch onto master, because I added tests that will be useful for changing these highlight groups.

Try hard not to break existing behaviors. You're right that ideally `_` would be handled in all numerics and that octals should also allow `0`, `0o`, and `0O` as leading characters, but if we can't get those changes done in a way that doesn't negatively impact performance and that is relatively easy to read, then I'm ok with leaving master as-is.

k33nice:
Thanks for your clarification! A quick note on `goDecimalInt`: that change https://github.com/fatih/vim-go/pull/2587/files#diff-6bbebb95ce25cd40d0072544e2dad91eR165 was broke highlighting single zero as a number.

bhcleek:
Good catch. I'll fix that in a separate PR.

k33nice:
I'll try to rework this pr a little bit. Also just my point of view, but I'm pretty convinced that much better to receive false negatives than positives for error highlighting. Especially for the current case where we have these syntax rules are enabled by default.

bhcleek:
I agree that false negatives for errors is better than false positives. However, I see no reason to introduce false negatives where they don't exist now.

k33nice:
@bhcleek I added support of "_"  for decimal, hex, octal and binary and also some more test cases for it. I tried don't regress previous behavior. I'd be glad to get any suggestions or thoughts.

autoload/go/highlight_test.vim
```diff
@@ -104,21 +104,45 @@ function! Test_numeric_literal_highlight() abort
   let tests = {
         \ 'lone zero': {'group': 'goDecimalInt', 'value': '0'},
         \ 'integer': {'group': 'goDecimalInt', 'value': '1234567890'},
+        \ 'integerGrouped': {'group': 'goDecimalInt', 'value': '1_234_567_890'},
+        \ 'integerErrorLeadingUnderscore': {'group': 'goDecimalErrorInt', 'value': '_1234_567_890'},
+        \ 'integerErrorTrailingUnderscore': {'group': 'goDecimalErrorInt', 'value': '1_234_567890_'},
+        \ 'integerErrorDoubleUnderscore': {'group': 'goDecimalErrorInt', 'value': '1_234__567_890'},
+        \ 'integerErrorDoubleTrailingUnderscore': {'group': 'goDecimalErrorInt', 'value': '1_234__567_890__'},
```
bhcleek: this test doesn't add value over the `integerErrorDoubleUnderscore`, because this test also has the double underscores in the middle of the number.

autoload/go/highlight_test.vim
```diff
@@ -104,21 +104,45 @@ function! Test_numeric_literal_highlight() abort
...
         \ 'hexadecimal': {'group': 'goHexadecimalInt', 'value': '0x0123456789abdef'},
+        \ 'hexadecimalGrouped': {'group': 'goHexadecimalInt', 'value': '0x012_345_678_9ab_def'},
         \ 'hexadecimalErrorLeading': {'group': 'goHexadecimalError', 'value': '0xg0123456789abdef'},
         \ 'hexadecimalErrorTrailing': {'group': 'goHexadecimalError', 'value': '0x0123456789abdefg'},
+        \ 'hexadecimalErrorDoubleUnderscore': {'group': 'goHexadecimalError', 'value': '0x__0123456789abdef'},
+        \ 'hexadecimalErrorDoubleTrailingUnderscore': {'group': 'goHexadecimalError', 'value': '0x__0123456789abdef__'},
```
bhcleek: this test doesn't add value over the `hexadecimalErrorDoubleUnderscore`, because this test also has the double underscores in the middle of the number.

autoload/go/highlight_test.vim
```diff
@@ -104,21 +104,45 @@ function! Test_numeric_literal_highlight() abort
...
         \ 'heXadecimalErrorTrailing': {'group': 'goHexadecimalError', 'value': '0X0123456789abdefg'},
+        \ 'heXadecimalErrorDoublePrefix': {'group': 'goHexadecimalError', 'value': '0XX0123456789abdef'},
         \ 'octal': {'group': 'goOctalInt', 'value': '01234567'},
+        \ 'octalPrefix': {'group': 'goOctalInt', 'value': '0o1234567'},
+        \ 'octalGrouped': {'group': 'goOctalInt', 'value': '0o1_234_567'},
         \ 'octalErrorLeading': {'group': 'goOctalError', 'value': '081234567'},
         \ 'octalErrorTrailing': {'group': 'goOctalError', 'value': '012345678'},
+        \ 'octalErrorDoublePrefix': {'group': 'goOctalError', 'value': '0oo12345678'},
+        \ 'octalErrorDoubleUnderscore': {'group': 'goOctalError', 'value': '0o__12345678'},
+        \ 'octalErrorDoubleTrailingUnderscore': {'group': 'goOctalError', 'value': '0o__12345678__'},
```
bhcleek: this test doesn't add value over the `octalErrorDoubleUnderscore`, because this test also has the double underscores in the middle of the number.

autoload/go/highlight_test.vim
```diff
@@ -104,21 +104,45 @@ function! Test_numeric_literal_highlight() abort
...
+        \ 'octalErrorTrailingUnderscore': {'group': 'goOctalError', 'value': '0o_123_456_78_'},
+        \ 'OctalPrefix': {'group': 'goOctalInt', 'value': '0O1234567'},
+        \ 'OctalErrorDoublePrefix': {'group': 'goOctalError', 'value': '0OO1234567'},
         \ 'binaryInt': {'group': 'goBinaryInt', 'value': '0b0101'},
+        \ 'binaryIntGroupd': {'group': 'goBinaryInt', 'value': '0b_01_01'},
```
bhcleek: (nit)`Groupd` -> `Grouped`

syntax/go.vim
```diff
@@ -162,15 +162,17 @@ endif
...
 " Integers
-syn match       goDecimalInt        "\<-\=\(0\|[1-9]\d*\)\%([Ee][-+]\=\d\+\)\=\>"
-syn match       goHexadecimalInt    "\<-\=0[xX]\x\+\>"
-syn match       goHexadecimalError  "\<-\=0[xX]\x*[^ \t0-9A-Fa-f]\S*\>"
-syn match       goOctalInt          "\<-\=0\o\+\>"
-syn match       goOctalError        "\<-\=0[XxBb\]]\@!\o*[^ \t0-7]\S*\>"
-syn match       goBinaryInt         "\<-\=0[bB][01]\+\>"
-syn match       goBinaryError       "\<-\=0[bB][01]*[^ \t01]\S*\>"
+syn match       goDecimalInt        "\<-\=\(0\|[1-9]_\?\(\d\+_\?\)*\)\%([Ee][-+]\=\d\+\)\=\>"
+syn match       goDecimalErrorInt   "\<-\=\(_\(\d\+_*\)\+\|\(\d\+_*\)\+_\{2,\}\(\d\+_*\)\+\|\(\d\+_*\)\+_\+\)\%([Ee][-+]\=\d\+\)\=\>"
```
bhcleek: To avoid matching the octal errors, the second and third branches of this should have `[1-9]` as their starting character instead of `\d`.

syntax/go.vim
```diff
@@ -162,15 +162,17 @@ endif
...
+syn match       goHexadecimalInt    "\<-\=0[xX]_\?\(\x\+_\?\)\+\>"
+syn match       goHexadecimalError  "\<-\=0[xX]_\?\(\x\+_\?\)*\(\([^ \t0-9A-Fa-f_]\|_\{2,\}\)\S*\|_\)\>"
+syn match       goOctalInt          "\<-\=0[oO]\?_\?\(\o\+_\?\)\+\>"
+syn match       goOctalError        "\<-\=0\([0-7oO_]*\([^ \t0-7oOxX\]\}\_]\+\|[oO]\{2,\}\|_\{2,\}\)[0-7oO_]*\)\+\S*\>"
```
bhcleek: Why does this have `the xX` in `[^ \t0-7oOxX\]`? And why is there a `\}` immediately following that set?

Why is this checking for double prefixes?

Why doesn't this have the same kind of check for trailing underscores that goHexadecimalError and goBinaryError have?

syntax/go.vim
```diff
@@ -162,15 +162,17 @@ endif
...
+syn match       goBinaryInt         "\<-\=0[bB]_\?\([01]\+_\?\)\+\>"
+syn match       goBinaryError       "\<-\=0[bB]_\?\([01]\+_\?\)*\([^ \t01_]\S*\|_\{2,\}\S*\|_\)\>"
```
bhcleek: for both goBinaryInt and goBinaryError, wouldn't `[01_]\+` suffic instead of `[01]\+_\?`?

syntax/go.vim
bhcleek: I think this might end up regressing https://github.com/fatih/vim-go/pull/2587/commits/69d5dad3a343de23953b3368a79c60f2d2b606bc.

autoload/go/highlight_test.vim
k33nice: Yeah thanks, nice catch! Fixed. Technically it works as expected because we have a separate group for matching trailing underscores. But I completely agree with you, to won't get in false-positives.

syntax/go.vim
k33nice: Yeah, it makes sense. Fixed.

syntax/go.vim
k33nice: We need the `xX` to doesn't match on valid hexadecimal smth like: `0x123`.

I change that regexp a bit so it doesn't have `\}` anymore. I don't have much experience with vimsrcipt just had tried to exclude something like `[]int{0}` but understood that it doesn't match anyway.

> Why is this checking for double prefixes?

Cuz it is forbidden to use double prefixes. e.g. `0oo123` syntaxly incorrect.

>Why doesn't this have the same kind of check for trailing underscores that goHexadecimalError and goBinaryError have?

Thanks, that's was fixed.

syntax/go.vim
k33nice: `[01_]\+` matches more than one sequential underscore, and that is syntaxly incorrect.

syntax/go.vim
k33nice: I changed it a bit already. But I cannot see any sequences that the previous regexp covered and the new does not. Of course, if not count those that it should not match (e.g. `0o1`, `0_1` `0123/5` , `0123+5`and so on). Maybe I missed smth. Could you please elaborate.

k33nice:
@bhcleek 
> Why do the *DoublePrefix tests need to exist? The existing *ErrorLeading tests should cover that case fine.

Yup, *ErrorLeading should cover that. I just added them as a common typo and since we need to explicitly check it in `goOctalError`. I got rid of them though.

I really appreciate your advice! I know that the review is the hard work especially for something tedious as regexp. I just decided to help little cuz errors started to annoyed during my work. I'll try to do it better and not so dreary for reviewers next time. Again, thanks a lot for all of your work and generally for maintaining this project!

syntax/go.vim
bhcleek: > We need the xX to doesn't match on valid hexadecimal smth like: 0x123

Then shouldn't it also have `Bb` to avoid matching binary literals?

> Cuz it is forbidden to use double prefixes. e.g. 0oo123 syntaxly incorrect.

I understand that. My question is really why a check for the double prefix is needed given that it `0oo123` is an error just as `0oa123` is an error

syntax/go.vim
bhcleek: Yes, it's syntactically in incorrect, but that's what `goBinaryError` is trying to catch. Can `goBinaryError` be simplified to `\<-\=0[bB]\([01_]*[^ \t01_]\S*\|[01]*__\S*\)\>`

autoload/go/highlight_test.vim
```diff
@@ -104,18 +104,37 @@ function! Test_numeric_literal_highlight() abort
...
         \ 'heXadecimalErrorTrailing': {'group': 'goHexadecimalError', 'value': '0X0123456789abdefg'},
         \ 'octal': {'group': 'goOctalInt', 'value': '01234567'},
+        \ 'octalPrefix': {'group': 'goOctalInt', 'value': '0o1234567'},
+        \ 'octalGrouped': {'group': 'goOctalInt', 'value': '0o1_234_567'},
         \ 'octalErrorLeading': {'group': 'goOctalError', 'value': '081234567'},
         \ 'octalErrorTrailing': {'group': 'goOctalError', 'value': '012345678'},
+        \ 'octalErrorDoubleUnderscore': {'group': 'goOctalError', 'value': '0o__1234567'},
+        \ 'octalErrorDoubleTrailingUnderscore': {'group': 'goOctalError', 'value': '0o_1234567__'},
+        \ 'octalErrorTrailingUnderscore': {'group': 'goOctalError', 'value': '0o_123456_7_'},
+        \ 'octalErrorTrailingO': {'group': 'goOctalError', 'value': '0o_123456_7o'},
```
bhcleek: What value does this add over octalErrorTrailing?

syntax/go.vim
bhcleek: https://github.com/fatih/vim-go/commit/69d5dad3a343de23953b3368a79c60f2d2b606bc handles list indexes as reported in the issue that commit fixed.

bhcleek: You're right that `[]int{0} ` needs to be handled correctly, as does references to a zeroth index in a slice.

I'll add tests for each of those.

k33nice: 1. If we apart this regexp than yes, indeed. But I believed that we also account for following overrides. Otherwise, we might need to fix up a decimal and binary error as well to do not match on some floats and imaginary.
2. The previous or-brach (`[^ \t0-7oOxX_/]`) matches on `0oa123` but not on `0oo123`, so we need to check that case specifically.

bhcleek:
Thanks for sticking with this. I know it's tedious to get through this review; you managed to pick up a piece of work for your first contribution that complex, high visibility, affects performance quite a bit, _and_ that will be noticed by the community very quickly if it introduces false positives.

syntax/go.vim
k33nice: Gotcha. Technically no, cuz it doesn't match on `0b_01__01` or `0b_01_01_` but I'll try to simplify it.

autoload/go/highlight_test.vim
k33nice: For checking a specific subgroup that matches on "oO". We need this group cuz we allow 'oO' in this subgroup `[^ \t0-7oOxX_/]`. And just realized that we need to expand that specific subgroup to [oOxX] to match on `0123x`.

syntax/go.vim
k33nice: Sigh. You're right, thanks! I've checked it on `[0]` and `[0]` but forgot about `[0][0]` and [][]int{{0},{0}}. I'll expand it.

k33nice:
> Why are there tests for double trailing underscores? Shouldn't those be tested already by virtue of the *TrailingUnderscore tests?

Agree. Them seems to be useless. Added them cuz was keeping in mind that case all the way.

Yup, can't disagree. That might be the best solution. I got the same thoughts when they started to falsy emerged during working. Handling by regexp these grouped integers is way too tricky. Furthermore, AFAIK

syntax/go.vim
```diff
@@ -162,17 +162,17 @@ endif
...
 " Integers
-syn match       goDecimalInt        "\<-\=\(0\|[1-9]_\?\(\d\+_\?\)*\)\%([Ee][-+]\=\d\+\)\=\>"
-syn match       goDecimalErrorInt   "\<-\=\(_\(\d\+_*\)\+\|\([1-9]\d*_*\)\+_\{2,\}\(\d\+_*\)\+\|\([1-9]\d*_*\)\+_\+\)\%([Ee][-+]\=\d\+\)\=\>"
+syn match       goDecimalInt        "\<-\=\(0\|[1-9]_\?\(\d\|\d\+_\?\d\+\)*\)\%([Ee][-+]\=\d\+\)\=\>"
+syn match       goDecimalError      "\<-\=\(_\(\d\+_*\)\+\|\([1-9]\d*_*\)\+__\(\d\+_*\)\+\|\([1-9]\d*_*\)\+_\+\)\%([Ee][-+]\=\d\+\)\=\>"
 syn match       goHexadecimalInt    "\<-\=0[xX]_\?\(\x\+_\?\)\+\>"
-syn match       goHexadecimalError  "\<-\=0[xX]_\?\(\x\+_\?\)*\(\([^ \t0-9A-Fa-f_]\|_\{2,\}\)\S*\|_\)\>"
+syn match       goHexadecimalError  "\<-\=0[xX]_\?\(\x\+_\?\)*\(\([^ \t0-9A-Fa-f_]\|__\)\S*\|_\)\>"
 syn match       goOctalInt          "\<-\=0[oO]\?_\?\(\o\+_\?\)\+\>"
 syn match       goOctalError        "\<-\=0[0-7oO_]*\(\([^ \t0-7oOxX_/\]\}]\|[oO]\{2,\}\|__\)\S*\|_\|[oOxX]\)\>"
```
bhcleek: Does `goOctalError` need to include `bB` in the classes that also check for `oOxX`?

k33nice: As I mentioned above, it depends on how we want to make it works. Technically, it works correctly without `bB`. Binary rules placed below so overrides octal errors if needed. But to make it fully independent should be expanded to a couple of other symbols like `.`, `i` that currently overrides by floats and imaginary rules. What's your point of view on it?

bhcleek: Fair point. 

Let's just add a test to cover the case in case these ever get over around

Sorry for asking again: it can be difficult to keep track of everything in these long running PRs

k33nice: Yeah, will do! That's fine, I also started to confuse in addressed comments.

bhcleek:
Is that the last of the changes?

k33nice:
> Is that the last of the changes?

Yup, I think that's it. Not sure that you saw all of my responses but it seems like all your reviews were address.  One more thing that I've found out that we have the same `goDecimalInt`, `goHexadecimalInt` and others in `syntax/gotexttmpl.vim`. I'm not sure that we want to change them here as well. It might be better to do in separate PR. What do you think?

bhcleek:
Let's leave the gotexttmpl for another PR.

I'll merge this later today.

PR #2585, opened 2019-11-14
Add simple use of interfunc tool

andrewstuart:
I got tired of hand-typing interface function implementations, a pattern I use often, so I wrote a quick and simple tool, and then vim-go hook to insert a known interface func signature at the cursor.

### DIFF
--- a/autoload/go/inter.vim
+++ b/autoload/go/inter.vim
@@ -1,0 +1,19 @@
+let s:cpo_save = &cpo
+set cpo&vim
+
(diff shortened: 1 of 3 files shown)


bhcleek:
If I'm understanding interfunc correctly, vim-go already has a solution for this. `:GoImpl` wraps https://github.com/josharian/impl.

andrewstuart:
This is different than impl, which I also use on a regular basis, but I'll have to figure something out; had to take down my implementation there temporarily.

Issue #2608, opened 2019-12-03
GoReferrers not working


mnarrell:
### What did you do? (required: The issue will be **closed** when not provided)

Ran the GoReferrers command on a function.

### What did you expect to happen?

Show list of invocations.

### What happened instead?

Error detected while processing function go#referrers#Referrers:
line   12:
E700: Unknown function: s:parse_guru_output
Error detected while processing function go#referrers#Referrers[12]..go#lsp#Referrers:
line    8:
E129: Function name required
E475: Invalid argument: 0
Press ENTER or type command to continue

### Configuration (**MUST** fill this out):

scriptencoding utf-8

" run :GoBuild or :GoTestCompile based on the go file
function! s:build_go_files() abort
  let l:file = expand('%')
  if l:file =~# '^\f\+_test\.go$'
    call go#test#Test(0, 1)
  elseif l:file =~# '^\f\+\.go$'
    GoBuild -i
  endif
endfunction

augroup vimgo
  au!
  au Filetype go command! -bang A call go#alternate#Switch(<bang>0, 'edit')
  au Filetype go command! -bang AV call go#alternate#Switch(<bang>0, 'vsplit')
  au Filetype go command! -bang AS call go#alternate#Switch(<bang>0, 'split')
  au Filetype go command! -bang AT call go#alternate#Switch(<bang>0, 'tabe')

  au FileType go imap <C-g> <ESC>:<C-u>GoDeclsDir<CR>
  au FileType go nmap <C-g> :GoDeclsDir<CR>

  " Option-T
  au FileType go nmap †  <Plug>(go-test-func)
  au FileType go nmap db <Plug>(go-doc-browser)
  au FileType go nmap de <Plug>(go-def-vertical)
  au FileType go nmap gb :<C-u>call <SID>build_go_files()<CR>
  au FileType go nmap gc <Plug>(go-callers)
  au FileType go nmap gf <Plug>(go-referrers)
  au FileType go nmap gi <Plug>(go-implements)
  au FileType go nmap gr <Plug>(go-rename)
  au FileType go nmap tc <Plug>(go-coverage-toggle)
augroup END

#### vim-go version:

master

####  `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):
<details><summary>vimrc</summary><br><pre>
Same as above
</pre></details>

#### Vim version (first three lines from `:version`):

NVIM v0.4.3
Build type: Release
[...]

PR #2609, opened 2019-12-03
referrers: fix handling of the gopls case

bhcleek:


bhcleek:
fixes #2608

mnarrell:
Working perfectly.  Thanks for the quick response.

PR #2613, opened 2019-12-05
fmt: respect buffer-level go_fmt_options

hdonnay:
This commit enables looking at the buffer-level fmt_options variable.
An example use is provided in the documentation.

Signed-off-by: Hank Donnay <hdonnay@gmail.com>

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -351,7 +351,7 @@ function! go#config#FmtCommand() abort
 endfunction
 
 function! go#config#FmtOptions() abort
-  return get(g:, "go_fmt_options", {})
+  return get(b:, "go_fmt_options", get(g:, "go_fmt_options", {}))
 endfunction
 
 function! go#config#FmtFailSilently() abort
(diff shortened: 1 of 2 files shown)


bhcleek:
Thank you for contributing @hdonnay .

Can you tell me more about the use case for this and what the user experience would be like for setting the buffer level variable?

hdonnay:
My use case is the one I added to the documentation, where I'd like to have the goimports `local` flag set correctly even if I have files in different modules open.

bhcleek:
So you'd expect people to set this with an autocmd in their vimrc?

bhcleek:
Can you update the documentation to provide a clearer example than setting the buffer level variable for all go files?

hdonnay:
Yeah, I imagine it'd be most useful with an autocmd.

I'm willing to write a better example, but I'm unsure what that looks like. The example I included is how I currently have it configured.

bhcleek:
Why are you configuring a buffer level variable for all go files instead of using the global?

hdonnay:
Yeah, the autocmd provided shells out to `go list` for every go buffer open. I'm not sure if there's a better way to go about it.

bhcleek:
I'll merge this soon. Do you mind if I push to your branch to tweak the docs a bit?

hdonnay:
Feel free

bhcleek:
Thank you for contributing.

PR #2614, opened 2019-12-05
Remove docs for defunct g:go_metalinter_disabled

ian-howell:
The `g:go_metalinter_disabled` option was removed in #2375. This removes it from the documentation


bhcleek:
Thank you

LGTM

PR #2615, opened 2019-12-06
Add build tags to GoVet, surface vet invocation errors

hoenn:
:wave: 
It doesn't look like values set in `:GoBuildTags` were being used with `:GoVet`. This should fix that. 

Since it didn't seem to surface elsewhere, I also added an echo for `vet` output if it returns a non-zero exit code but no code related diagnostics. 

This should help users if they both set `:GoBuildTags` and call `:GoVet -tags sometag`
```
vim-go: calling vet...
vim-go: go vet: tags flag may be set only once
```

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -126,9 +126,13 @@ function! go#lint#Vet(bang, ...) abort
   endif
 
   if a:0 == 0
-    let [l:out, l:err] = go#util#Exec(['go', 'vet', go#package#ImportPath()])
+    let [l:out, l:err] = go#util#Exec(['go', 'vet', '-tags', go#config#BuildTags(), go#package#ImportPath()])
   else
-    let [l:out, l:err] = go#util#Exec(['go', 'vet'] + a:000 + [go#package#ImportPath()])
+    if empty(go#config#BuildTags())
+      let [l:out, l:err] = go#util#Exec(['go', 'vet'] + a:000 + [go#package#ImportPath()])
+    else
+      let [l:out, l:err] = go#util#Exec(['go', 'vet', '-tags', go#config#BuildTags()] + a:000 + [go#package#ImportPath()])
+    endif
   endif
 
   let l:listtype = go#list#Type("GoVet")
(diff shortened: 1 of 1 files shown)


autoload/go/lint.vim
```diff
@@ -137,6 +141,12 @@ function! go#lint#Vet(bang, ...) abort
     let l:errorformat = "%-Gexit status %\\d%\\+," . &errorformat
     call go#list#ParseFormat(l:listtype, l:errorformat, out, "GoVet")
     let l:errors = go#list#Get(l:listtype)
+
+    if empty(l:errors)
+      call go#util#EchoError(l:out)
+      return
+    endif
```
bhcleek: I might be mistaken, but I think this isn't needed. Did you add for a specific reason?

autoload/go/lint.vim
```diff
@@ -126,9 +126,13 @@ function! go#lint#Vet(bang, ...) abort
...
   if a:0 == 0
-    let [l:out, l:err] = go#util#Exec(['go', 'vet', go#package#ImportPath()])
+    let [l:out, l:err] = go#util#Exec(['go', 'vet', '-tags', go#config#BuildTags(), go#package#ImportPath()])
   else
-    let [l:out, l:err] = go#util#Exec(['go', 'vet'] + a:000 + [go#package#ImportPath()])
+    if empty(go#config#BuildTags())
+      let [l:out, l:err] = go#util#Exec(['go', 'vet'] + a:000 + [go#package#ImportPath()])
+    else
+      let [l:out, l:err] = go#util#Exec(['go', 'vet', '-tags', go#config#BuildTags()] + a:000 + [go#package#ImportPath()])
+    endif
```
bhcleek: Can you simplify this if/else to be more similar to the pattern in https://github.com/fatih/vim-go/blob/master/autoload/go/def.vim#L31-L34 ?

autoload/go/lint.vim
hoenn: Hmm I believe invocations errors that occur in `:GoVet` will silently fail without this block.

Try running `:GoVet -tags sometag -tags sometag`. `vim-go` will just report 
`vim-go: calling vet...` and not report the error that `vet` would (`go vet: tags flag may be set only once`)

`go#lint#ErrCheck` does something similar to report invocation errors.

autoload/go/lint.vim
hoenn: Will, do! Thanks for the link 👍

autoload/go/lint.vim
```diff
@@ -125,18 +125,33 @@ function! go#lint#Vet(bang, ...) abort
...
+  let buildtags = go#config#BuildTags()
+  if buildtags isnot ''
+    let cmd += ['-tags', buildtags]
   endif
 
+  if a:0 != 0
+    call extend(cmd, a:000)
+  endif
+
+  let cmd += [go#package#ImportPath()]
```
hoenn: This seems to work but I'm not sure if I need `let l:cmd` here.

bhcleek: You got it right

PR #2639, opened 2019-12-31
Louis/vim8 term mode

ldelossa:
This PR enables term mode on vim8.
The implementation is more or less ported from the nvim implementation.

### DIFF
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -118,9 +118,8 @@ endfunction
 " Run runs the current file (and their dependencies if any) in a new terminal.
 function! go#cmd#RunTerm(bang, mode, files) abort
   let cmd = ["go", "run"]
-  let tags = go#config#BuildTags()
-  if go#config#BuildTags() > 0
-    call extend(cmd, ["-tags", tags])
+  if len(go#config#BuildTags()) > 0
+    call extend(cmd, ["-tags", go#config#BuildTags()])
   endif
 
   if empty(a:files)
--- a/autoload/go/cmd.vim
+++ b/autoload/go/cmd.vim
@@ -117,16 +117,16 @@ endfunction
 
 " Run runs the current file (and their dependencies if any) in a new terminal.
 function! go#cmd#RunTerm(bang, mode, files) abort
-  let cmd = "go run "
+  let cmd = ["go", "run"]
   let tags = go#config#BuildTags()
-  if len(tags) > 0
-    let cmd .= "-tags " . go#util#Shellescape(tags) . " "
+  if go#config#BuildTags() > 0
+    call extend(cmd, ["-tags", tags])
   endif
 
   if empty(a:files)
-    let cmd .= go#util#Shelljoin(go#tool#Files())
+    call extend(cmd, go#tool#Files())
   else
-    let cmd .= go#util#Shelljoin(map(copy(a:files), "expand(v:val)"), 1)
+    call extend(cmd, map(copy(a:files), "expand(v:val)"))
   endif
   call go#term#newmode(a:bang, cmd, s:runerrorformat(), a:mode)
 endfunction
@@ -136,7 +136,7 @@ endfunction
 " suitable for long running apps, because vim is blocking by default and
 " calling long running apps will block the whole UI.
 function! go#cmd#Run(bang, ...) abort
-  if has('nvim')
+  if go#config#TermEnabled()
     call go#cmd#RunTerm(a:bang, '', a:000)
     return
   endif
--- a/autoload/go/term.vim
+++ b/autoload/go/term.vim
@@ -74,8 +74,8 @@ function! go#term#newmode(bang, cmd, errorformat, mode) abort
           \ 'exit_cb' : function('s:exit_cb', [], state),
         \ }
 
-    if !(l:mode =~ "split") || !(l:mode =~ "new")
-      let l:term["vertical"] = l:mode
+    if l:mode =~ "vertical" || l:mode =~ "vsplit" || l:mode =~ "vnew"
+          let l:term["vertical"] = l:mode
     endif
 
     let l:id = term_start(a:cmd, l:term)
--- a/autoload/go/term.vim
+++ b/autoload/go/term.vim
@@ -30,47 +30,69 @@ function! go#term#newmode(bang, cmd, errorformat, mode) abort
 
   execute l:cd . fnameescape(expand("%:p:h"))
 
-  execute l:mode . ' __go_term__'
-
-  setlocal filetype=goterm
-  setlocal bufhidden=delete
-  setlocal winfixheight
-  setlocal noswapfile
-  setlocal nobuflisted
-
-  " explicitly bind callbacks to state so that within them, self will always
-  " refer to state. See :help Partial for more information.
-  "
-  " Don't set an on_stderr, because it will be passed the same data as
-  " on_stdout. See https://github.com/neovim/neovim/issues/2836
-  let l:job = {
-        \ 'on_stdout': function('s:on_stdout', [], state),
-        \ 'on_exit' : function('s:on_exit', [], state),
-      \ }
-
-  let l:state.id = termopen(a:cmd, l:job)
-  let l:state.termwinid = win_getid(winnr())
-
-  execute l:cd . fnameescape(l:dir)
-
-  " resize new term if needed.
   let l:height = go#config#TermHeight()
   let l:width = go#config#TermWidth()
 
-  " Adjust the window width or height depending on whether it's a vertical or
-  " horizontal split.
-  if l:mode =~ "vertical" || l:mode =~ "vsplit" || l:mode =~ "vnew"
-    exe 'vertical resize ' . l:width
-  elseif mode =~ "split" || mode =~ "new"
-    exe 'resize ' . l:height
+  " setup job for nvim
+  if has('nvim')
+    execute l:mode . ' __go_term__'
+    setlocal filetype=goterm
+    setlocal bufhidden=delete
+    setlocal winfixheight
+    setlocal noswapfile
+    setlocal nobuflisted
+
+    " explicitly bind callbacks to state so that within them, self will always
+    " refer to state. See :help Partial for more information.
+    "
+    " Don't set an on_stderr, because it will be passed the same data as
+    " on_stdout. See https://github.com/neovim/neovim/issues/2836
+    let l:job = {
+          \ 'on_stdout': function('s:on_stdout', [], state),
+          \ 'on_exit' : function('s:on_exit', [], state),
+        \ }
+    let l:state.id = termopen(a:cmd, l:job)
+    let l:state.termwinid = win_getid(winnr())
+    execute l:cd . fnameescape(l:dir)
+
+    " Adjust the window width or height depending on whether it's a vertical or
+    " horizontal split.
+    if l:mode =~ "vertical" || l:mode =~ "vsplit" || l:mode =~ "vnew"
+      exe 'vertical resize ' . l:width
+    elseif mode =~ "split" || mode =~ "new"
+      exe 'resize ' . l:height
+    endif
+    " we also need to resize the pty, so there you go...
+    call jobresize(l:state.id, l:width, l:height)
+
+  " setup term for vim8
+  elseif has('terminal')
+    let l:term = {
+          \ 'term_rows' : l:height,
+          \ 'term_cols' : l:width,
+          \ 'out_cb': function('s:out_cb', [], state),
+          \ 'exit_cb' : function('s:exit_cb', [], state),
+        \ }
+
+    if l:mode =~ "vertical" || l:mode =~ "vsplit" || l:mode =~ "vnew"
+      let l:term["vertical"] = 1
+    endif
+
+    let l:id = term_start(a:cmd, l:term)
+    let l:state.id = l:id
+    let l:state.termwinid = win_getid(bufwinnr(l:id))
+    execute l:cd . fnameescape(l:dir)
   endif
 
-  " we also need to resize the pty, so there you go...
-  call jobresize(l:state.id, l:width, l:height)
-
   call win_gotoid(l:state.winid)
-
   return l:state.id
+endfunction
+
+" out_cb continually concat's the self.stdout_buf on recv of stdout
+" and sets self.stdout to the new-lined split content in self.stdout_buf
+func! s:out_cb(channel, msg) dict abort
+  let self.stdout_buf = self.stdout_buf . a:msg
+  let self.stdout = split(self.stdout_buf, '\n')
 endfunction
 
 function! s:on_stdout(job_id, data, event) dict abort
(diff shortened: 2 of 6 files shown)


bhcleek:
Thank you for the PR. I'll review it soon.

autoload/go/cmd.vim
```diff
@@ -136,7 +136,7 @@ endfunction
 " suitable for long running apps, because vim is blocking by default and
 " calling long running apps will block the whole UI.
 function! go#cmd#Run(bang, ...) abort
-  if has('nvim')
+  if has('nvim') || has('terminal')
```
ldelossa: @bhcleek 
Believe this can go away in favor of just checking:
```
if go#config#TermEnabled()
```

autoload/go/term.vim
```diff
@@ -30,49 +30,73 @@ function! go#term#newmode(bang, cmd, errorformat, mode) abort
...
 
   call win_gotoid(l:state.winid)
-
   return l:state.id
 endfunction
 
+" out_cb continually concat's the self.stdout_buf on recv of stdout
+" and sets self.stdout to the new-lined split content in self.stdout_buf
+func! s:out_cb(channel, msg) dict abort
+  let self.stdout_buf = self.stdout_buf . a:msg
```
ldelossa: @bhcleek 
I am not sure if the msg here is NL buffered or not...

Also this may not be necessary if the `exit_cb` callback can simply grab the contents of the terminal buffer id instead of accumulating a string buffer... my thinking is we are duplicating the data now.

autoload/go/cmd.vim
bhcleek: I agree

autoload/go/term.vim
```diff
@@ -30,49 +30,73 @@ function! go#term#newmode(bang, cmd, errorformat, mode) abort
...
+          \ 'exit_cb' : function('s:exit_cb', [], state),
+        \ }
+
+    if l:mode =~ "vertical" || l:mode =~ "vsplit" || l:mode =~ "vnew"
+      let l:term["vertical"] = 1
+    endif
 
-  " we also need to resize the pty, so there you go...
-  call jobresize(l:state.id, l:width, l:height)
+    let l:id = term_start(["/bin/sh", "-c", a:cmd], l:term)
```
bhcleek: This won't work on Windows. Can you use `a:cmd` directly and make modify the call to `go#term#newmode` in `cmd.vim` to pass a list instead of a string?

autoload/go/term.vim
```diff
@@ -30,49 +30,73 @@ function! go#term#newmode(bang, cmd, errorformat, mode) abort
...
+  elseif has('terminal')
+    let l:term = {
+          \ 'term_rows' : l:height,
+          \ 'term_cols' : l:width,
+          \ 'out_cb': function('s:out_cb', [], state),
+          \ 'exit_cb' : function('s:exit_cb', [], state),
+        \ }
+
+    if l:mode =~ "vertical" || l:mode =~ "vsplit" || l:mode =~ "vnew"
+      let l:term["vertical"] = 1
```
bhcleek: When the user has used one of the modifiers (e.g. `belowright`), `l:term["vertical"] = 1` won't honor that modifier correctly. The docs say that modifiers are supported by `vertical`. Can you refactor to make sure this supports them correctly?

autoload/go/term.vim
```diff
@@ -30,49 +30,73 @@ function! go#term#newmode(bang, cmd, errorformat, mode) abort
...
+    endif
 
-  " we also need to resize the pty, so there you go...
-  call jobresize(l:state.id, l:width, l:height)
+    let l:id = term_start(["/bin/sh", "-c", a:cmd], l:term)
+    let l:state.id = l:id
+    let l:state.termwinid = win_getid(bufwinnr(l:id))
+    execute l:cd . fnameescape(l:dir)
+  else 
+    return -1
```
bhcleek: Given that `go#config#TermMode` is checked in every instance before calling this function and that `go#config#TermMode` checks that terminal mode is supported, is this else needed?

autoload/go/test.vim
```diff
@@ -6,7 +6,10 @@ set cpo&vim
 " compile the tests instead of running them (useful to catch errors in the
 " test files). Any other argument is appended to the final `go test` command.
 function! go#test#Test(bang, compile, ...) abort
-  let args = ["test", '-tags', go#config#BuildTags()]
+  let args = ["test"]
+  if go#config#BuildTags() > 0
+    call extend(args, ["-tags", go#util#Shellescape(tags)])
+  endif
```
bhcleek: This shouldn't change. Please revert this; if you'll make the change to pass a list instead of a string as the terminal command, I think you'll find that this change is unnecessary.

autoload/go/test.vim
```diff
@@ -31,7 +34,8 @@ function! go#test#Test(bang, compile, ...) abort
   endif
 
   if go#config#TermEnabled()
-    call go#term#new(a:bang, ["go"] + args, s:errorformat())
+    let l:cmd = "go " . join(args)
+    call go#term#new(a:bang, l:cmd, s:errorformat())
```
bhcleek: A list is the right thing to pass. This should also be reverted.

autoload/go/term.vim
bhcleek: That seems like something that should be figured out before merging.

autoload/go/term.vim
ldelossa: You are correct. Will remove.

autoload/go/test.vim
ldelossa: @bhcleek 

go#cmd#RunTerm passes a string
```
  let cmd = "go run "
  let tags = go#config#BuildTags()
  if len(tags) > 0
    let cmd .= "-tags " . go#util#Shellescape(tags) . " "
  endif

  if empty(a:files)
    let cmd .= go#util#Shelljoin(go#tool#Files())
  else
    let cmd .= go#util#Shelljoin(map(copy(a:files), "expand(v:val)"), 1)
  endif
  call go#term#newmode(a:bang, cmd, s:runerrorformat(), a:mode)
```

Is the intention to have GoRun pass a string and GoTest pass an array to #newmode ?

autoload/go/term.vim
ldelossa: @bhcleek a few issues here...

vim's api will not spawn a shell "automatically" for you, instead if you pass a string, it splits it by white space, and hands it directly to execvp. When the string is split, the final argument is double quoted. The following error is encountered with GoRun

```
package '/home/louis/git/go/claircore/cmd/libindexhttp/main.go': malformed module path "'/home/louis/git/go/claircore/cmd/libindexhttp/main.go'": invalid char '\''
```
When go test is attempted and passed as a list the command fails due to no "empty" string after -tags

```
vim-go: [test -run TestSecDBParse$ -v -race -tags ] FAIL
```
Vim docs say to explicitly call your shell 

```
    The command is executed directly, not through a shell, the
                'shell' option is not used.  To use the shell: 
        let job = job_start(["/bin/sh", "-c", "echo hello"])
               Or: 
        let job = job_start('/bin/sh -c "echo hello"')
```
Should we check for unix/windows and split the logic ?

autoload/go/test.vim
bhcleek: I believe that picking one argument type is the correct thing and that the one argument type should be a list, not a string. I realize that `go#cmd#RunTerm` passes a string - let's make it a list (`go#test#Test` already uses the list form).

ldelossa: sounds good. Will do this.

autoload/go/term.vim
bhcleek: Yes, it's executed directly, not through a shell, but the docs don't say that you _must_ use a shell. The implication from the docs seem to be that if you want a shell, then it needs to be explicitly use. Maybe I'm missing something, though?

As for `go test`. The reason the tags aren't there is because they're empty. In that case, they `-tags` shouldn't be passed at all: https://github.com/fatih/vim-go/blob/master/autoload/go/test.vim#L9 should be changed to be more like https://github.com/fatih/vim-go/blob/master/autoload/go/def.vim#L31-L34

ldelossa: That is what I sought to accomplish here:
https://github.com/fatih/vim-go/pull/2639/files/e6546c8cb8947432dacec111fc2bc2c4678e0ff8#diff-c223bf93d0004ffef86b493d16e3e953R9

But you requested revert on that. Maybe my solution is not adequate here? 

Re Shell:
You are correct, shell is not required, however it passively fixed the double quoting issue which occurs here:
https://github.com/ldelossa/vim-go/blob/782297eed08ed47cedbba3b64626e532365cde61/autoload/go/cmd.vim#L126-L131

Everything works if the `go#util#Shelljoin` calls are changed to just "join" calls (vim builtins).

The issue is, the shelljoin calls are producing a string "go run 'some/path'", term_start then splits that and creates a string for each one [ "go", "run", "'some/path'"] and the double quoting of the path elements results in Go tool barking 

Let me address turning #RunTerm's cmd argument to a list first and see if it naturally solves this issue.

bhcleek: Sounds good. Let me know if you need a hand.

I think all of this is a consequence of trying to use a string argument instead of a list. Once it's converted to use a list instead of a string, the quoting that Vim is generating before passing to `execvp` will stop and everything should Just Work™.

autoload/go/test.vim
ldelossa: @bhcleek this look okay now ?

autoload/go/cmd.vim
```diff
@@ -117,16 +117,16 @@ endfunction
...
   let tags = go#config#BuildTags()
-  if len(tags) > 0
-    let cmd .= "-tags " . go#util#Shellescape(tags) . " "
+  if go#config#BuildTags() > 0
+    call extend(cmd, ["-tags", go#util#Shellescape(tags)])
   endif
 
   if empty(a:files)
-    let cmd .= go#util#Shelljoin(go#tool#Files())
+    call add(cmd, go#util#Shelljoin(go#tool#Files()))
```
ldelossa: Okay @bhcleek 
This Shelljoin call is still breaking the term_start function. the cmd var looks like this after this method runs:
```
['go', 'run', '''/home/louis/git/go/claircore/cmd/libindexhttp/main.go''']
```
When I swap the Shelljoin method with just "join" I get a correct arg list

```
['go', 'run', '/home/louis/git/go/claircore/cmd/libindexhttp/main.go']
```

may need a little help here.

bhcleek: You're on the right track. Using `join` instead of `go#util#Shelljoin` would be the right thing if the list of files were a single argument, but each file is an argument. I think what's needed here is
```
call exend(cmd, go#tool#Files())
```

autoload/go/test.vim
bhcleek: It's close. This does not nee to use`go#util#Shelljoin` now, because there's no need to shell escape it since it won't be processed by a shell. `go#config#BuildTags()` returns the space separated list of tags, so using its return value unaltered should be fine.

```
 let args = ["test"]
  if go#config#BuildTags() > 0
    call extend(args, ["-tags", go#config#BuildTags()])
  endif
```

autoload/go/cmd.vim
```diff
@@ -117,16 +117,16 @@ endfunction
 
 " Run runs the current file (and their dependencies if any) in a new terminal.
 function! go#cmd#RunTerm(bang, mode, files) abort
-  let cmd = "go run "
+  let cmd = ["go", "run"]
   let tags = go#config#BuildTags()
-  if len(tags) > 0
-    let cmd .= "-tags " . go#util#Shellescape(tags) . " "
+  if go#config#BuildTags() > 0
+    call extend(cmd, ["-tags", go#util#Shellescape(tags)])
```
bhcleek: Remove the `go#util#Shellescape` call; `tags` alone should be fine since `cmd` won't be joined into a string anymore.

autoload/go/cmd.vim
```diff
@@ -117,16 +117,16 @@ endfunction
...
   endif
 
   if empty(a:files)
-    let cmd .= go#util#Shelljoin(go#tool#Files())
+    call add(cmd, go#util#Shelljoin(go#tool#Files()))
   else
-    let cmd .= go#util#Shelljoin(map(copy(a:files), "expand(v:val)"), 1)
+    call add(cmd, go#util#Shelljoin(map(copy(a:files), "expand(v:val)"), 1))
```
bhcleek: Similarly, I think this should be
```
call extend(cmd, a:files)
```

autoload/go/cmd.vim
ldelossa: worked :+1:

ldelossa:
squashed the changes and re-pushed one commit. will handle vertical modifier next.

bhcleek:
Thank you for sticking with it! This is looking good. After you're done, I'll probably add some tests before merging the PR. Do you mind if I push to your branch once you're done with your changes?

ldelossa:
@bhcleek do not mind.

autoload/go/cmd.vim
```diff
@@ -117,16 +117,15 @@ endfunction
 
 " Run runs the current file (and their dependencies if any) in a new terminal.
 function! go#cmd#RunTerm(bang, mode, files) abort
-  let cmd = "go run "
-  let tags = go#config#BuildTags()
-  if len(tags) > 0
-    let cmd .= "-tags " . go#util#Shellescape(tags) . " "
+  let cmd = ["go", "run"]
```
ldelossa: @bhcleek just fixed this at work today - build tags weren't working correctly. looks ok? needed to len the resulting string. The results of #BuildTag() will always be a string correct?

bhcleek: yeah, this looks right now.

autoload/go/term.vim
ldelossa: @bhcleek 

Not sure I follow this correctly. That "vertical" is coming from the vim-go config file. Would a user ever have a chance to specify `belowright` when spawning a terminal via vim-go commands?

bhcleek: The docs for `term_start`'s `vertical` parameter read
> split the window vertically; note that other window position can be defined with command modifiers, such as |:belowright|.

`g:go_term_mode` is the command that's used to open the terminal window. I suspect that `l:term["vertical"]`'s value should be `l:mode` so that `g:go_term_mode` is honored in Vim8 as well as it's honored in Neovim.

autoload/go/term.vim
```diff
@@ -74,8 +74,8 @@ function! go#term#newmode(bang, cmd, errorformat, mode) abort
           \ 'exit_cb' : function('s:exit_cb', [], state),
         \ }
 
-    if l:mode =~ "vertical" || l:mode =~ "vsplit" || l:mode =~ "vnew"
-      let l:term["vertical"] = 1
+    if !(l:mode =~ "split") || !(l:mode =~ "new")
```
ldelossa: @bhcleek 

Does this accomplish the goal? If we do not see l:mode set to the normal split or new modifiers, we then populate l:term["vertical"] with which ever value the user placed in their config. Assuming "split" an "new" are not modifiers... but maybe I'm wrong.

bhcleek: The option could be `belowright`. I think you want to keep the original condition and this change's assignment to `l:term`.

ldelossa: ahh I see now, I haven't used these modifiers before. 
So the correct thing to do would be 

```
if l:mode =~ "vertical" || l:mode =~ "vsplit" || l:mode =~ "vnew"
      let l:term["vertical"] = l:mode
endif
```

and leave it at that?

bhcleek: Yes

ldelossa:
@bhcleek think this PR is ready for those tests you mentioned.

bhcleek:
Thanks @ldelossa !

I'll add some tests soon and will probably do a release of vim-go before merging this.

bhcleek:
@ldelossa I'm going to be writing those tests soon. Do you mind if I rebase your branch onto master before I start?

ldelossa:
Not at all.

bhcleek:
Thanks for contributing!

ldelossa:
Thanks for all the help on that one. Was a fun one.

PR #2650, opened 2020-01-09
factorize import in hello_world templates

skuzzymiglet:
Most programs will need a factorized `import`, so here it's the default.

### DIFF
--- a/templates/hello_world.go
+++ b/templates/hello_world.go
@@ -1,6 +1,8 @@
 package main
 
-import "fmt"
+import (
+	"fmt"
+)
 
 func main() {
 	fmt.Println("vim-go")
--- a/templates/hello_world_test.go
+++ b/templates/hello_world_test.go
@@ -1,6 +1,8 @@
 package main
 
-import "testing"
+import (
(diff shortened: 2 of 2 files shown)


skuzzymiglet:
What does the place you mentioned do? It seems to accept `import (`

bhcleek:
It tries to ignore headers in when asserting file contents. I'm not 100% certain that's what needs to be resolved to fix the tests that this PR broke, but I think it is.

bhcleek:
@skuzzymiglet I'm not sure this is necessary at all.  Multiple import statements are valid, and `goimports` will consolidate them when `g:go_fmt_command` is set to `'goimports'`. 

However, if you want to fix the test, I'll merge it.

I dug a little deeper into the failing tests. The problem with the broken test is not in https://github.com/fatih/vim-go/blob/master/autoload/gotest.vim#L90-L100. The problem is that `Test_TemplateCreate` truly doesn't pass. The failing test is in https://github.com/fatih/vim-go/blob/master/autoload/go/template_test.vim

PR #2655, opened 2020-01-14
lint: improve golangci-lint interoperability

pierreN:
Add a new option go_metalinter_use_local_config which disables linters+options for golangci-lint so users can set them on a per project basis via a config file.

golanci-lint errorformat : handle cases when no column number, since when adding some new available linters don't provide a column number

The idea of the PR is to provide the less possible amount of options to golanci-lint so vim-go users can do whatever they want in their yaml config file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ IMPROVEMENTS:
   [[GH-2696]](https://github.com/fatih/vim-go/pull/2696)
 * Add support for `gomodifytags --skip-unexported`
   [[GH-2660]](https://github.com/fatih/vim-go/pull/2660)
+* Show problems that prevent golangci-lint from running linters.
+  [[GH-2706]](https://github.com/fatih/vim-go/pull/2706)
 
 BUG FIXES:
 * Use the discovered full path for gopls when renaming.


autoload/go/config.vim
```diff
@@ -273,11 +273,19 @@ function! go#config#MetalinterCommand() abort
...
 function! go#config#MetalinterAutosaveEnabled() abort
   let l:default_enabled = ["vet", "golint"]
 
   if go#config#MetalinterCommand() == "golangci-lint"
-    let l:default_enabled = ["govet", "golint"]
+    if go#config#MetalinterUseLocalConfig() == 0
+      let l:default_enabled = ["govet", "golint"]
+    else
+      let l:default_enabled = []
+    endif
```
bhcleek: Modify this and other similar changes to be more in line with usual go standards that you may already be familiar with:
```
let l:default_enabled = ["govet", "golint"]
if go#config#MetalinterUserLocalConfig()
  let l:default_enabled = []
endif
```

pierreN:
> Instead of using an option to specify to use the config file, how about not setting --disable-all when the enabled linters is empty?

Yes this is indeed better, I was just not sure it was OK to change this behavior. Please have a look at the new diff. Thanks!

bhcleek:
This looks great. Thank you for simplifying!

I'll merge this after the v1.22 release.

bhcleek:
@pierreN when trying to resolve some conflicts with master, I inadvertently pushed the wrong branch to your repo. I've created #2715, which is all of your changes rebased onto master. Sorry for the trouble!

pierreN:
My bad, I should have checked/referenced https://github.com/fatih/vim-go/issues/2707 in this PR.

Thanks for the merge and for maintaining the awesome vim-go !

bhcleek:
That was a new PR created since this one. The conflicts were introduced when it was merged; you did everything right.

Thanks again for contributing.

PR #2660, opened 2020-01-18
Add new g:go_addtags_skip_private option

igrmk:
Tool gomodifytags now has -skip-private option for adding tags only to public fields

This PR mirrors this option in vim config

For example you may have following struct

    type Struct struct {
        Public  int
        private int
    }

Currently GoAddTags command would add these tags

    type Struct struct {
        Public  int `json:"public"`
        private int `json:"private"`
    }

With new option it would add

    type Struct struct {
        Public  int `json:"public"`
        private int
    }

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -244,6 +244,10 @@ function! go#config#AddtagsTransform() abort
   return get(g:, 'go_addtags_transform', "snakecase")
 endfunction
 
+function! go#config#AddtagsSkipPrivate() abort
+  return get(g:, 'go_addtags_skip_private', 0)
+endfunction
+
 function! go#config#TemplateAutocreate() abort
   return get(g:, "go_template_autocreate", 1)
 endfunction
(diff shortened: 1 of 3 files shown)


bhcleek:
Given that `:GoAddTags` can take a range of line to operate on, what's the value in adding this configuration option that will apply to every invocation?

igrmk:
I personally never need tags for private fields. So it is just much simpler for me. I don't need to select a range with this option

bhcleek:
The flag name for gomodifytags has changed. Would you be willing to update this PR similarly?

igrmk:
PR is updated according to new flag name

bhcleek:
Thanks @igrmk 

Do you mind if I rebase and push a fix for a couple of nitpicks to your branch before I merge?

igrmk:
Please go ahead! Thanks @bhcleek

bhcleek:
Thanks for contributing @igrmk !

PR #2684, opened 2020-01-30
plugin: pin gopls to v0.2.2

bhcleek:
Pin gopls to v0.2.2 to protect the upcoming vim-go release from any
breaking changes in the upcoming gopls release.

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -51,7 +51,7 @@ let s:packages = {
       \ 'gogetdoc':      ['github.com/zmb3/gogetdoc@master'],
       \ 'goimports':     ['golang.org/x/tools/cmd/goimports@master'],
       \ 'golint':        ['golang.org/x/lint/golint@master'],
-      \ 'gopls':         ['golang.org/x/tools/gopls@latest', {}, {'after': function('go#lsp#Restart', [])}],
+      \ 'gopls':         ['golang.org/x/tools/gopls@v0.2.2', {}, {'after': function('go#lsp#Restart', [])}],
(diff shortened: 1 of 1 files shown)


stamblerre:
Just out of curiousity - when will vim-go users be able to get `gopls/v0.3.0`? It should be a lot more stable than `v0.2.2`.

bhcleek:
I wanted the vim-go release to use a known version of gopls, but I intend to flip vim-go master back using gopls@latest in a few days.

There's already a PR to flip it back: https://github.com/fatih/vim-go/pull/2685

stamblerre:
Perfect, thanks for letting me know. Will you do another `vim-go` release when you flip it?

bhcleek:
I don't intend to do a release when I flip it to gopls@latest again. Though, I suspect I'll need to do a release shortly after Go 1.14 is out.

I'm vacillating on whether a given release of vim-go should use a known version of gopls or whether latest is ok.

stamblerre:
Excellent - thanks! My preference is that we stick to `@latest` since that is basically our team's assertion that we feel it is the most stable release we have. I know @myitcv prefers to pin a specific version and can maybe speak to the reasoning there.

myitcv:
It's very brief reasoning. The `govim` integration tests represent our (the `govim` folks) confidence in what `govim` can/can't do. Therefore it only makes sense (for us) to pin to a specific version of `gopls` because that's all we can test against. i.e. we can't test against a release/commits that haven't happened yet (obviously!) and it doesn't make much sense to test against older versions. Given `govim` takes care of managing the `gopls` version, installation etc there is no overhead. Indeed there is even less overhead because people don't have to play the game of ensuring the right version is in their PATH.

PR #2686, opened 2020-01-30
complete: drop support for gocode

bhcleek:


QuLogic:
Remove from top-level `README.md` too?

PR #2696, opened 2020-02-05
lsp: use gopls defaults by default

bhcleek:
Refactor gopls configurations to use gopls' defaults by default so that
users can take advantage of any changes to an updated gopls without
modifying vim-go's configuration.


stamblerre:
Thanks!

PR #2717, opened 2020-02-11
fix: set errorformat based on the command name instead of the command line

### LINKED ISSUE #2707: File names in GoMetalinter output are not recognized by vim when g:go_metalinter_command is a full commandline

flw-cn:
This should close #2707

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -349,12 +349,12 @@ function! s:metalinterautosavecomplete(metalinter, filepath, job, exit_code, mes
 endfunction
 
 function! s:errorformat(metalinter) abort
-  if a:metalinter == 'golangci-lint'
+  if a:metalinter =~ '^golangci-lint\>'
     " Golangci-lint can output the following:
     "   <file>:<line>:<column>: <message> (<linter>)
     " This can be defined by the following errorformat:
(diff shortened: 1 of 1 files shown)


bhcleek:
This would not catch many use cases and is insufficient to robustly solve #2707. As explained in the documentation for `g:go_metalinter_command`, the supported values are `gopls` and `golangci-lint`.

flw-cn:
So, how to fix my problem?
I want custom golangci-lint command line parameters such as `--print-issued-lines=false` and ` --exclude-use-default=true`, and any more parameters.

bhcleek:
Set `g:go_metalinter_enabled` to an empty list and configure your golangci-lint file to do what you want.

PR #2720, opened 2020-02-12
lint: handle more golangci-lint output

bhcleek:
Handle problems reported by golangci-lint that don't have a file
location at all.

Handle error reported by golangci-lint that don't have a column number.


bhcleek:
@Integralist this at least stop hiding some of the information that was being overwritten for you in some cases. It's not perfect, but it's an imnprovement.

Issue #2712, opened 2020-02-10
go#auto#template_autocreate fails on Windows if shell not cmd.exe


drichardson:
### What did you do? (required: The issue will be **closed** when not provided)

1. On Windows, change .vimrc to `set shell=pwsh`
2. vim example.go

### What did you expect to happen?

Opens with new file template.

### What happened instead?

```
"example.go" [New File]
Error detected while processing function go#auto#template_autocreate[6]..go#template#create[4]..go#tool#PackageName[1]..go#util#ExecInDir[9]..go#util#Exec[17]..<SNR>11_exec[7]..<SNR>11_system:
line   11:
E484: Can't open file C:/Users/doug/AppData/Local/Temp/VIo794.tmp
Error detected while processing function go#auto#template_autocreate[6]..go#template#create[4]..go#tool#PackageName[1]..go#util#ExecInDir[9]..go#util#Exec[17]..<SNR>11_exec[7]..<SNR>11_system:
line   11:
E484: Can't open file C:/Users/doug/AppData/Local/Temp/VIo794.tmp
Press ENTER or type command to continue
```

#### vim-go version:

d698de4a29eae6bf0ce608865d5c7631037b0f0c

####  `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):
<details><summary>vimrc</summary><br><pre>
set nocompatible
filetype off

set rtp+=~/.vim/bundle/Vundle.vim
call vundle#begin()
Plugin 'VundleVim/Vundle.vim'
Plugin 'fatih/vim-go', { 'do': ':GoUpdateBinaries' }
call vundle#end()

filetype plugin indent on

set shell=pwsh
</pre></details>

#### Vim version (first three lines from `:version`):
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Feb  4 2020 23:03:05)
MS-Windows 64-bit console version
Included patches: 1-210

#### Go version (`go version`):
go version go1.13.7 windows/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>
set GO111MODULE=
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\doug\AppData\Local\go-build
set GOENV=C:\Users\doug\AppData\Roaming\go\env
set GOEXE=.exe
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GONOPROXY=
[...]

PR #2724, opened 2020-02-13
Use COMSPEC to find cmd.exe.

drichardson:
PR #2712 constructed a path to cmd.exe, but there is a standard environment variable that points to cmd.exe, which is COMSPEC. This fix uses COMSPEC.


bhcleek:
Thanks for contributing again

Issue #2635, opened 2019-12-26
GoFmt formats the file but also (and undesirably) saves the file


adkein:
(FYI I wrote a working fix for this issue branched off master, but I get a 403 when trying to push the branch to this repository.)

### What did you do? (required: The issue will be **closed** when not provided)

I edited a .go file and I ran the GoFmt command.

### What did you expect to happen?

I expected the code file to be formatted and for the file's changes to not be saved.

### What happened instead?

The code file was formatted but also the file's changes were saved.

#### vim-go version:

v.1.19

####  `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):

I won't link to my >2,000 lines of vimrc spread across multiple files, by request. :) Nothing in all of it occurs to me as relevant to the present issue. Happy to provide any specific detail on request.

#### Vim version (first three lines from `:version`):

VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Dec  9 2019 11:02:16)
Included patches: 1-127
Compiled by pizza@middlepart
Huge version without GUI.  Features included (+) or not (-):

#### Go version (`go version`):

go version go1.10.3 linux/amd64

#### Go environment

GOARCH="amd64"
GOBIN="/home/pizza/work/bin"
GOCACHE="/home/pizza/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/pizza/work"
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build165265883=/tmp/go-build -gno-record-gcc-switches"

bhcleek:
Feel free to fork vim-go, push your branch to the fork, and create the PR in this repo.

bhcleek:
[...]

Issue #2525, opened 2019-09-30
folds get closed on write (unfocused split)


salarkhan:
### What did you do? (required: The issue will be **closed** when not provided)
installed vim-go
set `let g:go_fmt_experimental = 1`
folded some code
vertically split the screen (editing different locations of the same file)
wrote to the file 

### What did you expect to happen?
folds are maintained. related to https://github.com/fatih/vim-go/issues/502

### What happened instead?
the folds on the split i was not focused on disappeared

####  `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):
<details><summary>vimrc</summary><br><pre>
" use vim settings over vi
set nocompatible

" PLUGIN CONFIG
" ---------------------------------
" set dir for plugins
call plug#begin('~/.vim/plugged')

" plugins
Plug 'fatih/vim-go', { 'do': ':GoUpdateBinaries' }

" initialize plugin system
call plug#end()

" GENERAL CONFIG
" ---------------------------------
" syntax highlighting
syntax on

" colorscheme
set background=dark
let g:gruvbox_contrast_dark='hard'
colorscheme gruvbox

" colorize matched searches
set hlsearch

" allow me to change buffers without saving
set hidden

" do not create backup files
set nobackup

" do not create swap files
set noswapfile

" exclude files/dirs we don't care about
set wildignore+=*/tmp/*,dist,node_modules,*.pyc

" workaround to ensure :E maps to Explore
let g:loaded_logipat=1

" DISPLAY CONFIG
" ---------------------------------
" open splits to right/bottom
set splitbelow
set splitright

" always use vertical diffs
set diffopt+=vertical

" remember my folds pls
set foldmethod=syntax
set viewoptions=cursor,folds,slash,unix
" LANGUAGE CONFIG
" ---------------------------------
" let g:go_fmt_command = "goimports"
let g:go_fmt_experimental = 1
let g:go_highlight_functions = 1
let g:go_highlight_function_parameters= 1
let g:go_highlight_extra_types = 1
let g:go_highlight_function_calls = 1
</pre></details>

#### Vim version (first three lines from `:version`):

NVIM v0.3.8
Build type: Release
LuaJIT 2.0.5

#### Go version (`go version`):

go version go1.13 darwin/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GO111MODULE=""
GOARCH="amd64"
GOBIN=""
[...]

PR #2729, opened 2020-02-17
fmt: format with gopls

bhcleek:
##### fmt: add an option to format with gopls

Add an option to format with gopls.


bhcleek:
Fixes #2483 
Fixes #2635
Fixes #2525

PR #2732, opened 2020-02-18
make qf open with rightbelow as default

Shane-XB-Qian:
made qf open with righbelow looks would be more convenience. 
or maybe it can be added a option for it .. 
but for now, i think default as 'rightbelow' perhaps better, since many user may like set a side bar/split at the right side (e.g project tree or tagbar or sth) ...

### DIFF
--- a/autoload/go/list.vim
+++ b/autoload/go/list.vim
@@ -31,7 +31,7 @@ function! go#list#Window(listtype, ...) abort
   if a:listtype == "locationlist"
     exe 'lopen ' . height
   else
-    exe 'copen ' . height
+    exe 'rightbelow copen ' . height
   endif
 endfunction
 


bhcleek:
What practical effect does this have? `copen` by default opens a new horizontal split, and `belowright` doesn't seem to ever cause it to open on the right.

Shane-XB-Qian:
for example, you have a 2 gofile horizontal splits and 1 sth other split at the far right side.
the current qf would be opened at down next to that 'sth' split.
well, but `belowright` would be nice which just next down to your gofile.

bhcleek:
The quickfix list opens at the bottom of the window by default. I'm not following what you mean by a 'sth' window or why the quickfix window wouldn't be at the bottom.

Shane-XB-Qian:
'sth' means sth not a window which opened with a gofile, user may like the opened qf near the gofile such as `rightbelow` (-vs- always 'bottom') would be better.
the option perhaps is similiar like `g:go_debug_windows`, perhaps can be added as a option, but for now, i think to add `rightbelow` by default should be better.

bhcleek:
Have you tried this change to see that it has the effect you're expecting?

Shane-XB-Qian:
yes, for now, just added a `rightbelow` ahead `copen`.

bhcleek:
I don't see this adding enough value. What's done for debugging has surprising edge cases that make it difficult to justify the effort of maintaining what you're proposing here.

Shane-XB-Qian:
what is it? but yes, just a proposal, i can hack to add this at my local, but just wish it can be a common option or default. if it's not now, wish there would be (a option) at someday. :)

bhcleek:
You can use something like `:autocmd FileType qf wincmd J` in your vimrc so that you don't have to hack vim-go.

Shane-XB-Qian:
i can try, but no, i did not expect qf located at bottom (always). 
besides 'location list' opened & located at rightbelow was its default (if i did not recall wrong).
i wish there would  be an option/setting to adjust qf 's location - perhaps like `g:go_debug_windows`.
so far, looks such 'option' should be the easiest way to make it - anyway, this is just a proposal, up to you. :)

PR #2734, opened 2020-02-19
Fix g:go_gopls_matcher variable name in docs

orlangure:
#2728 introduced a change in configuration, but the example in the docs was not updated. This PR fixes it.


bhcleek:
Thank you for contributing.

PR #2756, opened 2020-03-03
lsp: only set tempModFile when g:go_gopls_temp_modfile is set

bhcleek:


bhcleek:
`modTempFile` also isn't a configuration name. Prior to this PR, vim-go was sending `modTempFile` unconditionally and `modTempfile` only when it should.

Issue #2765, opened 2020-03-13
GoCoverage appends build tags even when empty interfering with manually providing it


asutoshpalai:
### What did you do? (required: The issue will be **closed** when not provided)

```
:GoCoverage . -tags abc 
```

### What did you expect to happen?
For the normal coverage workflow

### What happened instead?
```
vim-go: run "go help test" or "go help testflag" for more information
```

With `shell-command` logging, I got the following output:

```
vim-go: job command: ['go', 'test', '-tags', '', '-coverprofile', '/var/folders/wb/hj0lprc55kl472yvy0v059yc0000gq/T/vVByTpj/2', '.', '-tags', 'abc']
```

Basically even though my build tags are empty, it's still adding it. The relevant code I believe is at https://github.com/fatih/vim-go/blob/52b66c0b125e095bc5a1d54d9810c5a0c8141906/autoload/go/coverage.vim#L53 

Locally removing the tags fixed it.

#### vim-go version:
364a8cfe33b21151dacb8886a0c4eb0855759d57

####  `vimrc` you used to reproduce (use a *minimal* vimrc with other plugins disabled; do not link to a 2,000 line vimrc):
<details><summary>vimrc</summary><br><pre>
call plug#begin()
Plug 'fatih/vim-go', { 'do': ':GoUpdateBinaries'  }
call plug#end()

let g:go_debug = ['shell-commands']

</pre></details>

#### Vim version (first three lines from `:version`):

VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Feb  3 2020 01:40:25)
macOS version
Included patches: 1-200

#### Go version (`go version`):

go version go1.13.7 darwin/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="redacted"
GOENV="redacted"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="redacted"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.13.7/libexec"
GOSUMDB="off"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.13.7/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
[...]

PR #2766, opened 2020-03-13
Fixed build tags for coverage when empty #2765

asutoshpalai:
Fixes for cases where build tags are empty and we need to provide tags with the command options.

This also fixes `:GoCoverageBrowser` not accepting additional arguments


bhcleek:
Thank you for contributing.

`:GoBuildTags` exists so that you can set the build tags; trying to support multiple ways of setting them will be difficult to support. For example, if this PR were to be merged, then we'd have people reporting that the tags set on the command line and those set on with `:GoBuildTags` didn't use the right set of tags.

Issue #2771, opened 2020-03-14
Suggestion: Go Format on Window Close


eL0ck:
I find that doing a formatting on every save with `let g:go_fmt_autosave = 1` is a little much for me.   I'm happy enough if a format is done when i'm done with the file.

Suggest a parameter such as `let g:go_fmt_autoclose = 1` that defaults to `0` (off).

PR to follow

bhcleek:
If that's your desire, then I'd suggest turning off the fmt in save and write a simple autocmd to call `:GoFmt` on `BufWinLeave` .

I'm unlikely to merge a PR that does what you describe, because the feature isn't generally needed, and adding an autocmd to your vimrc is a simple fix for your preferred workflow.

eL0ck:
Good point.  Got a bit trigger happy there.

bhcleek:
No worries. Your willingness to contribute is much appreciated! Please don't let this discourage you from making future contributions.

PR #2772, opened 2020-03-14
Add option for GoFmt on close

eL0ck:
Ref Issue #2771 .

setting the following will disable auto formatting on save but enable it on closing the buffer.

```
let g:go_fmt_autosave = 0 
let g:go_fmt_autoclose = 1
```


bhcleek:
Thank you for contributing.

`autocmd BufWinLeave <buffer> GoFmt` in a `after/ftplugin/go.vim` file can solve this for you without a PR.

PR #2797, opened 2020-04-01
adjust imports on save

bhcleek:
Introduce a new option, g:go_imports_autosave, to allow imports to be
adjusted even when g:go_fmt_command is not set to goimports.

Closes #2795

PR #2800, opened 2020-04-01
adjust imports on save

### LINKED ISSUE #2795: Organizing imports on save with gopls
Is it possible to organize imports on save using `go_imports_mode='gopls'`? I currently use the following configuration, which explicitly says that `go_fmt_command` is `goimports`, and it runs on save:

```
  let g:go_fmt_autosave=1
  let g:go_fmt_command='goimports'
  let g:go_imports_mode='gopls'
```

If I change `go_fmt_command` to `gopls`, the code is formatted on save as expected, but the imports are no longer organized. Is there a way to both use `gopls` for `fmt` and `imports`, and run it on save?

I use the recent master, with all binaries updated to their master as well.

Thank you!

bhcleek:
Introduce a new option, g:go_imports_autosave, to allow imports to be
adjusted even when g:go_fmt_command is not set to goimports.

Closes #2795

### DIFF
--- a/autoload/go/auto.vim
+++ b/autoload/go/auto.vim
@@ -133,11 +133,28 @@ function! s:handler(timer_id)
 endfunction
 
 function! go#auto#fmt_autosave()
-  if !(go#config#FmtAutosave() && isdirectory(expand('%:p:h')) && expand('<afile>:p') == expand('%:p'))
+  if !(isdirectory(expand('%:p:h')) && expand('<afile>:p') == expand('%:p'))
     return
   endif
 
-  " Go code formatting on save
+  if !(go#config#FmtAutosave() || go#config#ImportsAutosave())
+    return
+  endif
+
+  if go#config#ImportsAutosave() && !(go#config#FmtAutosave() && go#config#FmtCommand() == 'goimports')
+    call go#fmt#Format(1)
+
+    " return early when the imports mode is goimports, because there's no need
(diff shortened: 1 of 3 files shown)


bhcleek:
@orlangure I think the recent changes on master have taken care of the issues that you saw on #2797. Feel free to give this a run through if you want.

orlangure:
Found two issues (and one probably related to `gopls`):
1. When creating a new file and using both `fmt` and `imports` on save (enabled by default), the first save causes error since the file does not exist, and the next save works.
2. When using the same options, every save leads to `Press ENTER or type command to continue`.

The last issue is inconsistency, I think it should be reported to `gopls`, but I'll post here anyway:
```go
package main

import (
	"time"
	"github.com/orlangure/gnomock"
)

func meh() {
	_ = 0
	_ = gnomock.DefaultPort
	_ = time.Now()
}
```
Saving this file changes the order of imports to
```go
import (
	"github.com/orlangure/gnomock"
	"time"
)
```
And saving this file changes it to
```go
import (
	"time"

	"github.com/orlangure/gnomock"
)
```

bhcleek:
The first item, the error on the first save, is an issue on master that I'll address in a separate PR.

I can't duplicate the second item.

I can duplicate the third item; based on the debug output (`let g:go_debug=["lsp"]`), it looks like  a `gopls` issue.

PR #2815, opened 2020-04-07
quick fix for error message format; do not preserve errors when autos…

bhcleek:
…ave = 1


bhcleek:
Created the PR to capture the change referenced in #2815 to protect against a deletion of the fork . Now that it's captured, I'm closing this, because the changes here need more work and may not be mergeable as-is:
1. removing the call s:preserveerrors should not be merged
2. the errorformat string changes may cause other golangci-lint build output to not be matched; we likely need additional prepend warning and error errorformat entries here to the existing string so that they will be preferred, but leave the current errorformat entries in place, too, so that those variant can also be caught.

PR #2859, opened 2020-04-24
syntax: highlight package name if requested

atsman:
Hello,

I am building custom color scheme for vim and was trying to highlight package name like:
```
package *main* // <-- I want main to be highlighted
```
Because package is keyword and has highest priority I wan't able to customise it from my local vim config, so I prepared this PR. If there is a way to do it differently please let me know.

Anyway I think it is nice addition for those who need the same functionality.

Thank you!

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -406,6 +406,10 @@ function! go#config#SearchBinPathFirst() abort
   return get(g:, 'go_search_bin_path_first', 1)
 endfunction
 
+function! go#config#HighlightPackageName() abort
+  return get(g:, 'go_highlight_package_name', 0)
+endfunction
+
 function! go#config#HighlightArrayWhitespaceError() abort
   return get(g:, 'go_highlight_array_whitespace_error', 0)
 endfunction
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -2038,6 +2038,12 @@ Disable everything (same as not setting 'foldmethod' to `syntax`):
 >
   let g:go_fold_enable = []
 <
+                                     *'g:go_highlight_package_name'*
+
+Highlight package name.
+
+  let g:go_highlight_package_name = 0
+
                                      *'g:go_highlight_array_whitespace_error'*
 
 Highlight white space after `[]`. >
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -11,7 +11,13 @@ endif
 
 syn case match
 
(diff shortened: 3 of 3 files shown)


bhcleek:
Thank you for contributing.

Yes, that's the way to do it.

atsman:
@bhcleek please let me if I need to change anything in order to merge it!

bhcleek:
You don't need to do anything else. I'm considering whether to merge this or not: syntax highlighting is already relatively complex and impacts performance; I'm not sure I want to add anything else to syntax highlighting.

arp242:
> Because package is keyword and has highest priority I wan't able to customise it from my local vim config

You can override it by using `syn clear` first:

```
fun! s:go()
    syn clear goPackage
    syn keyword goPackage     package nextgroup=goPackageName skipwhite
    syn match   goPackageName /\w\+/  contained skipwhite
    hi goPackageName guibg=red
endfun
autocmd Syntax go call s:go()
```

Looks like using `syn clear goPackage | syn keyword [..]` doesn't work, as it doesn't seem to recognize the `|` here (bug? feature?)

bhcleek:
I didn't catch the detail that Martin identified regarding changing within your own vimrc. Since there is a path forward for you without introducing more complexity to vim-go, I'm going to close this.

Note that you _might_ need to put your override in a `after/syntax/go.vim` file.

atsman:
@arp242 amazing thank you!

arp242:
Yeah, the after/ directory is the same; I just prefer `autocmd` because I like to keep everything in my vimrc file, but it's just a matter of personal taste.

> syntax highlighting is already relatively complex and impacts performance

As a little thread hijack: I did quite a lot of work on improving the performance in gopher.vim; for example these 2 are identical in behaviour:

    syn match       goSingleDecl      /\(import\|var\|const\) [^(]\@=/
    syn match       goSingleDecl      /^\s*\(import\|var\|const\) [^(]\@=/

But the second one is faster as the `^\s*` means Vim can stop matching quite a lot faster in most cases.

I made a bunch of small changes like this; with the default settings it's more than twice as fast (and even faster than Vim's runtime go.syntax).

I don't really have the time to compare and upstream it all since there are quite a few differences, but if you're looking to improve performance then you can probably do a few things, FYI. Feel free to reach out if you've got any specific questions about it.

bhcleek:
Thanks @arp242 I may do that after the next release; it'd be nice to tidy up the syntax file and improve performance to boot.

PR #2863, opened 2020-04-27
fixes errormessage when detecting a neovim version < 0.4.0

paukul:
currently, vim-go states that neovim version has to be >= 0.3.2 while it actually checks for a version >= 0.4.0

This commit changes the errormessage


bhcleek:
Thank you!

PR #2867, opened 2020-04-28
Add bar option to relevant commands

piersy:
This PR adds the bar option to many of the commands defined by vim go, so that users can set up bindings such as.

```vim
nnoremap <buffer> <leader>b :<C-u>GoImports \| GoBuild<CR>
```
I've attempted to apply the `-bar` option to every command where it cant conflict with the command arguments (I.E the arguments list cannot contain a `|`)


bhcleek:
Thank you for contributing.

I wonder whether these changes are necessary, though: do the mappings (`:help go-mappings`) work with the bar? If so, then I'm not sure this is needed.

For the use case you described, there's no need to do `:GoImports` before `:GoBuild`, because `:GoBuild` saves the buffer before building and imports can be adjusted automatically on save.

piersy:
Hi @bhcleek thanks for your response, the plug mappings seem to work :+1: 

For reference I'm able to use the following mapping
```
nmap <buffer> <leader>b <Plug>(go-imports) \| :call <SID>build_go_files()<CR>
``` 
Where `build_go_files` is defined in the [vim-go tutorial](https://github.com/fatih/vim-go-tutorial#vimrc-improvements-1)

My use case is actually a bit more complicated. I use the following autosave options.
```
let g:go_fmt_autosave = 1
let g:go_fmt_options = {
  \ 'gofmt': '-s',
  \ }
```

So I was looking for another way to incorporate `:GoImports`.

Ideally I would be able to run both `gofmt -s` and `goimports` automatically on save, but I don't think that is possible.

bhcleek:
Try using `gopls` for fmting and for imports on save with
```
let g:go_fmt_command='gopls'
let g:go_imports_mode='gopls'
let g:go_imports_autosave=1
```

PR #2883, opened 2020-05-09
snippets: add fmt.Errorf error wrapping snippets

swithek:
Add UltiSnips snippets for error wrapping and extending with fmt.Errorf.


bhcleek:
Thank you for contributing!

PR #2884, opened 2020-05-12
use pkg.go.dev instead of godoc.org

Code-Hex:
Because it is recommended to use pkg.go.dev in 2020.

See: https://blog.golang.org/pkg.go.dev-2020


bhcleek:
Thank you for contributing. I'll merge this soon after the next release.

PR #2887, opened 2020-05-19
Improve formatting experience

sheerun:
This PR fixes annoying bug in golang formatting with goimport: after saving the errors are in wrong lines and each package added by go fmt. You can reproduce this by configuring vim-go to use go-import and formatting following file:

```golang
package main

import (
	"github.com/gocql/gocql"
)

func main() {
	ip := "ok"

	fmt.Println(gocql.NewCluster(ip))
}
```

After saving this the `fmt` is added to imports, but you'll get an error on line `ip := "ok"` with following message: `undeclared name: fmt`. The error happens because file is formatted with goimport after saving, not before.

To to fix this this PR formats current buffer directly, instead of using temporary files. This results in much faster formatting, no errors with "undeclared names" like above. Also the undo history is preserved by default (no need for experimental mode).

### DIFF
--- a/autoload/go/fmt.vim
+++ b/autoload/go/fmt.vim
@@ -9,15 +9,28 @@
 let s:cpo_save = &cpo
 set cpo&vim
 
-"  we have those problems :
-"  http://stackoverflow.com/questions/12741977/prevent-vim-from-updating-its-undo-tree
-"  http://stackoverflow.com/questions/18532692/golang-formatter-and-vim-how-to-destroy-history-record?rq=1
-"
-"  The below function is an improved version that aims to fix all problems.
-"  it doesn't undo changes and break undo history.  If you are here reading
-"  this and have VimL experience, please look at the function for
-"  improvements, patches are welcome :)
 function! go#fmt#Format(withGoimport) abort
+  let l:lines = getbufline(bufnr('%'), 1, '$')
+  let l:filepath = expand("%")
+
+  let l:out = go#fmt#run(a:withGoimport, l:lines, l:filepath)
+
+  if empty(l:out)
+    return
+  endif
+
+  if l:out == l:lines
+    return
+  endif
+
+  " replace buffer with formatted lines
+  call go#fmt#replace(l:out)
+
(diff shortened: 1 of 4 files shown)


sheerun:
I have no idea why tests are failing, they seem unrelated

bhcleek:
Can you fill out a bug report? I can't duplicate the problem you've described using the information provided.

We can't remove the experimental support; as the code comments document, the experimental support exists for good reasons, mostly to do with code folding and undo. I understand that your PR aims to address the undo concern, but I'd need to test thoroughly to make sure it doesn't mess with folding.

As an aside, you might consider using these options in your vimrc:
```
let g:go_fmt_command='gopls'
let g:go_imports_mode='gopls'
let g:go_imports_autosave=1
```

sheerun:
I've rollbacked experimental support. gopls doesn't work well in my project, I prefer pure goimports. I'll try to file proper bug report, but regardless this PR makes formatting much faster

bhcleek:
Thank you for contributing.

This PR might make formatting more performant in some cases, but will cause problems performance regressions in others. The approach used by formatting when using `gofmt` and `goimports` was arrived at after many iterations early in vim-go's life  and balances a number of concerns. Therefore, I am not eager to merge this change, because it would likely cause regressions of issues that were solved long ago.

I am curious why `gopls` doesn't work in your project and would be happy to help you figure that out if you want.

Please do file an issue; I've never seen the problem you've described and can't duplicate based on the sparse description here despite multiple attempts.

sheerun:
@bhcleek I tried my best to retain all features that were already present. The code comes in good part from vim-prettier plugin so it should be stable enough. I filled #2888 to show the issue I'm experiencing. It looks like it happens when combined with coc.vim

bhcleek:
The performance problems that come with using `setline` and `append` are explained in https://github.com/fatih/vim-go/issues/502#issuecomment-126243590. Given that history, the ability to use `gopls` for formatting, which will be much faster, and the test failures here, I'm don't expect to merge this.

bhcleek:
More explanations of the issues with `append` and `setline`: https://github.com/fatih/vim-go/issues/459

PR #2900, opened 2020-05-28
chore: godoc popup border style

zhengkai:
godoc popup cames from `g:go_doc_popup_window`

what I changed:

1. beautify popup border style if utf-8 supported
2. add variable `g:go_doc_popup_border` to support custome border style

default style (original):

default style

default style if utf-8 supported:

utf-8 style

bounded style with `let g:go_doc_popup_border = ['─', '│', '─', '│', '╭', '╮', '╯', '╰']` :

rounded style

### DIFF
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -69,7 +69,7 @@ function! s:GodocView(newposition, position, content) abort
       endif
       call popup_atcursor(split(a:content, '\n'), {
             \ 'padding': [1, 1, 1, 1],
-            \ 'borderchars': get(g:, 'go_doc_popup_border', borderchars),
+            \ 'borderchars': borderchars,
             \ 'border': [1, 1, 1, 1],
             \ })
     elseif has('nvim') && exists('*nvim_open_win')
--- a/autoload/go/doc.vim
+++ b/autoload/go/doc.vim
@@ -6,6 +6,8 @@
 let s:cpo_save = &cpo
 set cpo&vim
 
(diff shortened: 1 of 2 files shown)


zhengkai:
Sorry, I don't know how to change it in other way ? I saw the original style is hard coded https://github.com/fatih/vim-go/blob/9fa8d1edb27d9b1c88aadab7d31f8854438faf73/autoload/go/doc.vim#L66

Maybe someone like rounded or other style ? so he/she can changed it simply by new option ?

bhcleek:
I think that we can do with out `g:go_doc_popup_border` and just use `borderchars` as you've defined it.

i.e.
```
let borderchars = ['-', '|', '-', '|', '+', '+', '+', '+']
if &encoding == "utf-8"
   let borderchars = ['─', '│', '─', '│', '┌', '┐', '┘', '└']
endif
```

is sufficient; I don't see a reason to support a custom border style with `g:go_doc_popup_border`.

zhengkai:
From my perspective, I like as many options as possible. They don’t waste any memory and CPU. The only drawback is that they need to be written. But I have already done it.

Not everyone needs it, but someone needs it.

Maybe we can look at the opinions of others. If all of you feel unnecessary, you can remove it ( or by me ? )

bhcleek:
Yes, please remove the new option. More options is unnecessary overhead and should only be added when it adds clear value, because every option adds complexity. Constraints are useful, and I only have so much time to debug, triage issues, and maintain vim-go.

PR #2908, opened 2020-06-06
debug: fix RPC params of method CreateBreakpoint

MarcoQin:
When to clear breakpoint after debug process started, the debugger will exit with status 0. Correct the RPC method's param type can fix this.

### DIFF
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -1063,7 +1063,7 @@ function! go#debug#Breakpoint(...) abort
     if type(l:found) == v:t_dict && !empty(l:found)
       exe 'sign unplace '. l:found.id .' file=' . l:found.file
       if s:isActive()
-        let res = s:call_jsonrpc('RPCServer.ClearBreakpoint', {'id': l:found.id})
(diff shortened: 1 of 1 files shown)


bhcleek:
I'm not sure what this fixes. Can you provide a description the problem that this solves?

MarcoQin:
I encountered a problem when I did the following actions:

1. :GoDebugStart
2. Use :GoDebugBreakpoint to add a breakpoint on a line
3. :GoDebugContinue
4. Use :GoDebugBreakpoint again to clear the breakpoint I set previously
5. The status bar shows: vim-go [debug] SUCCESS, and I noticed the dlv process exit (maybe killed by the job, because the jsonRPC did not correct responsed)

The RPCServer.ClearBreakpoint method receives an int `id` and a string `name`, but the type of l:found.id is string. So I just cast it to int, and the error gone.

bhcleek:
Thank you for contributing @MarcoQin.

The underlying issue here would be ideally resolved by executing a function that provides the id as  a number, `sign_getplaced()`, but that function is not available until Vim 8.1.0614 and vim-go currently supports > Vim 8.0.1453 and therefore must rely on normal mode commands, which return a string, so this is indeed the right fix.

PR #2910, opened 2020-06-06
Add support for running a single inner test.

elliotcourant:
This allows GoTestFunc to run a single t.Run() inner test rather than
running all of the inner tests within a single test function. It will
only run an inner test if the user's cursor is inside the test
declaration. If the cursor is on the test definition itself then it will
run the entire test. If the cursor is within the definition it will run
the first inner test that is on a line before or equal to the current
cursor line. If there are no inner tests then the entire test method
will be run normally.

### DIFF
--- a/autoload/go/test.vim
+++ b/autoload/go/test.vim
@@ -114,16 +114,34 @@ function! go#test#Func(bang, ...) abort
   "
   " for the full list
   " :help search
-  let test = search('func \(Test\|Example\)', "bcnW")
+  " Find the last line that has a full test declaration.
+  let outerTest = search('func \(Test\|Example\)', "bcnW")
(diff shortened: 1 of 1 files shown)


bhcleek:
I hope you'll consider making future contributions to vim-go. Thank you again 🙇

elliotcourant:
Would you be open to having this option behind a setting that defaults to disabled? If not that's totally fine!

bhcleek:
I'm inclined to say no, because hiding it behind a variable wouldn't address the concerns I mentioned earlier: vim-go's behavior should be predictable and consistent regardless of the particulars of the code being edited.

I'll give it some thought, though. I've wanted a feature like this for a while, too, but haven't been able to identify a way to implement it with a UX that I believe is needed.

PR #2917, opened 2020-06-11
support rename all file for go.mod project

lvht:
gopls rename require to switch to the root dir to rename all files.

### DIFF
--- a/autoload/go/rename.vim
+++ b/autoload/go/rename.vim
@@ -54,7 +54,7 @@ function! go#rename#Rename(bang, ...) abort
 
   let l:wd = go#util#ModuleRoot()
   if l:wd == -1
-    let l:wd = getcwd()
+    let l:wd = expand("%:p:h")
   endif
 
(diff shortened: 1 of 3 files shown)


autoload/go/rename.vim
```diff
@@ -52,7 +52,12 @@ function! go#rename#Rename(bang, ...) abort
     return
   endif
 
-  let [l:out, l:err] = go#util#ExecInDir(l:cmd)
+  let l:wd = go#util#ModuleRoot()
+  if l:wd == -1
+    let l:wd = getcwd()
```
bhcleek, line 57: `l:wd` will need to be set to the directory of the current buffer here for parity with current behavior.

lvht:
@bhcleek

lvht:
> are you ok if I push to your branch
feel free to make any change @bhcleek

bhcleek:
Apologies for force pushing to your branch. I've only force pushed over the new commits that I've added, though, so your local branch shouldn't be affected. I'll merge as soon as tests pass.

PR #2936, opened 2020-06-25
Look for packages in $GOPATH/pkg/mod/ for completion.

WhoIsSethDaniel:
Strip out version information placed in the path.

### DIFF
--- a/autoload/go/package.vim
+++ b/autoload/go/package.vim
@@ -236,6 +236,7 @@ function! go#package#Complete(ArgLead, CmdLine, CursorPos) abort
   for dir in dirs
     " this may expand to multiple lines
     let root = split(expand(dir . '/pkg/' . s:goos . '_' . s:goarch), "\n")
+    let root = add(root, expand(dir . '/pkg/mod'))
     let root = add(root, expand(dir . '/src'), )
     let root = extend(root, vendordirs)
     let root = add(root, module)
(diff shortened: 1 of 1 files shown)


WhoIsSethDaniel:
Everything I 'go get' is placed in $GOPATH/pkg/mod. This location is not currently looked at for package name completion. Without this change (or some similar change) I cannot use package name completion for third-party packages.

stamblerre:
I'm not sure if this is relevant to the PR (not sure I fully understand it), but the location of the module cache will be determined by the `GOMODCACHE` env variable in Go 1.15: https://github.com/golang/go/issues/34527.

bhcleek:
Thanks, Rebecca! That is relevant for this PR.

WhoIsSethDaniel:
It's not clear to me what I am supposed to be doing, if anything. I look at the failing test and I don't see how they relate to the change I made. Test_goimports and the routines it calls don't intersect with completion at all AFAICT.  If there is something I can do wrt making this more compatible with Go 1.15 I will do that.  Is that required for this to be accepted?

bhcleek:
I'm willing to merge this without the GOMODCACHE change: I can add that in a separate PR after merging.

I'll take a closer look at the tests this weekend and will merge if the failures aren't relevant to this PR.

WhoIsSethDaniel:
@bhcleek thanks. FWIW I cannot duplicate the test failure.

bhcleek:
Thank you for contributing @WhoIsSethDaniel !

PR #2959, opened 2020-07-16
Ensure Neovim version requirements match

mattr-:
This will keep folks from seeing 0.4.0 in one spot and 0.3.2 in another.

I ran into this while using Neovim 0.3.2 on a debian server. :smiley:


bhcleek:
Thank you for contributing!

PR #2961, opened 2020-07-21
Add omnifunc icase override setting

peterstace:
This change allows a plugin user to set the `icase` key in `omnifunc` results. By setting this to 1, the user is able to filter omnifunc results in a non-case sensitive manner.

For example:

1. Enter insert mode.
2. Type `fmt.`
3. Trigger omnifunc completion using `<C-X><C-O>`
4. Deselect the first item in the list (`Formatter`) by pressing `<C-P>`.
5. If `g:go_code_completion_icase` is set to 0 (default), then anything you type next will filter the completion list but in a case sensitive manner. This is also the behaviour when `icase` is not present at all. But if `g:go_code_completion_icase` set to 1, then the filtering is case-insensitive.

---

My own personal workaround to get this behaviour has been to define my own `omnifunc` for Go files that calls `go#complete#Complete` and then modifies the result on the way back through.

```
function MyOmniFunc(findstart, base)
    let results = go#complete#Complete(a:findstart, a:base)
    if !a:findstart
        for result in results
            let result.icase = 1
        endfor
    endif
    return results
endfunction
autocmd FileType go set omnifunc=MyOmniFunc
```

(so if you don't want to accept this PR, there aren't any dramas on my part since I have an functional workaround).

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -494,6 +494,10 @@ function! go#config#CodeCompletionEnabled() abort
   return get(g:, "go_code_completion_enabled", 1)
 endfunction
 
+function! go#config#CodeCompletionIcase() abort
+  return get(g:, "go_code_completion_icase", 0)
+endfunction
+
 function! go#config#Updatetime() abort
   let go_updatetime = get(g:, 'go_updatetime', 800)
   return go_updatetime == 0 ? &updatetime : go_updatetime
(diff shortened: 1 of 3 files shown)


bhcleek:
Thanks for contributing.

LGTM

Shane-XB-Qian:
very good, i like this !
lots langs' omni compl supported case insensitive, or lang itself did not care func name must be Upper case or not.
Unlucky, so far, golang & this plugin limited had to input 1st Upper case letter to filter omni compl items, in its most cases ...

PR #2974, opened 2020-08-07
Documentation: installation with vim-plug to get the latest release

etnnth:
After installing the plugin with vim-plug i realized that it got the master branch and not the latest release.
I changed the documentation so that the recommended way to install vim-go with vim-plug install the latest release.
I don't really know what could be the solution for Pathogen, Vim 8, I guess we could use :
```
git clone -b v1.23 https://github.com/fatih/vim-go
```
but this will requires to be change for each release.
I have no idea how to do the same with Vundle

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ for popular package managers:
 * [Pathogen](https://github.com/tpope/vim-pathogen)
   * `git clone https://github.com/fatih/vim-go.git ~/.vim/bundle/vim-go`
 * [vim-plug](https://github.com/junegunn/vim-plug)
-  * `Plug 'fatih/vim-go', { 'do': ':GoUpdateBinaries' }`
+  * `Plug 'fatih/vim-go', { 'tag': '*', 'do': ':GoUpdateBinaries' }`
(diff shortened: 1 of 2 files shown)


bhcleek:
How does plug's `tag` option work? How does it determine which tag to use?

etnnth:
Documentation on vim-plug says that { 'tag': '*' } option will point to the latest tagged release: https://github.com/junegunn/vim-plug/releases/tag/0.9.0 and line 169-170 of https://github.com/junegunn/vim-plug/blob/master/doc/plug.txt

bhcleek:
Thanks for contributing!

In general, I try to keep master working well. I use it daily, though there can sometimes be issues.

teleivo:
@etnnth thank you for this suggestion! Just in case someone else is also experiencing this: when I do a `:PlugUpdate` plug will always run the `Post-update` hook and thus reinstall all vim-go related binaries like gopls, guru, ... Even though I am on the latest release of vim-go already. That behavior is unexpected to me. I would expect plug not to do anything if I am already on the latest release. Like it does when I am just relying on master.

PR #2994, opened 2020-09-01
lsp: support gofumpt and unknown gopls settings

### LINKED ISSUE #2993: Allow gofumports as goimports alternative
https://github.com/mvdan/gofumpt was added to golangci-lint, and it looks reasonable.
Please allow to set `let g:go_fmt_command = 'gofumports'`.
Also it may be a good idea to include it into list of installed/updated binaries.

bhcleek:
Closes #2993

### DIFF
--- a/autoload/go/lsp/message.vim
+++ b/autoload/go/lsp/message.vim
@@ -269,6 +269,8 @@ function! go#lsp#message#ConfigurationResult(items) abort
     let l:tempModfile = go#config#GoplsTempModfile()
     let l:analyses = go#config#GoplsAnalyses()
     let l:local = go#config#GoplsLocal()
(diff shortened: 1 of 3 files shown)


autoload/go/lsp/message.vim
```diff
@@ -322,6 +324,14 @@ function! go#lsp#message#ConfigurationResult(items) abort
       let l:config.local = l:local
     endif
 
+    if l:gofumpt isnot v:null
+      let l:config.gofumpt = l:local
```
powerman, line 328: Isn't this a typo? I suppose it should be `= l:gofumpt`.

bhcleek: Yeah, that's a typo. PR coming up.

PR #3001, opened 2020-09-05
doc: fix GoAddTags for fixed value

Thanonchai:
The help section of GoAddTags seems wrong.  When executing :GoAddTags key=value, vim-go actually inserts key=value:"_tagname_"

However, executing :GoAddTags key:value seems to be correct... not sure if the doc should be fixed or the code should be fixed though, but I went with the easier, but safer route by fixing the documentation for now :-).


bhcleek:
Thank you for contributing!

PR #3017, opened 2020-09-20
lsp: fix g:go_gopls_gofumpt to allow more liberal truthy values

bhcleek:


autoload/go/lsp/message.vim
```diff
@@ -325,7 +325,11 @@ function! go#lsp#message#ConfigurationResult(items) abort
     endif
 
     if l:gofumpt isnot v:null
-      let l:config.gofumpt = l:gofumpt
+      if l:config.gofumpt
```
NicholasAsimov, line 328: Shouldn't this check `l:gofumpt`?

Getting this error at the moment when setting `g:go_gopls_gofumpt`: `E716: Key not present in Dictionary: gofumpt`

bhcleek: 🤦 PR is to fix it is up now: #3022

Issue #1788, opened 2018-04-08
Debugger mappings should be configurable


arp242:
Now it will always just remap `F5`, `F10`, etc. overriding any previous mappings that the user may have set.

See: https://github.com/fatih/vim-go/pull/1696#issuecomment-374813174

bhcleek:
Closed by #3035

PR #3035, opened 2020-10-01
debug: configurable key mappings

### LINKED ISSUE #3012: Key mappings only active during debugging
Would you be interested in a PR where a debug editing mode is simulated. In which a user can define a key mapping which is active only when a debug session is in place and the user's mappings are restored afterwards. This will allow you to make mappings that are more natural like 'c' for continue without losing the builtin vim 'c' change semantics.

I wrote an extension a while back that does this. Would love to get your feedback on it https://github.com/kernel-panic96/vim-go-extensions

EDIT: sorry for some reason I cannot add a feature request label

kernel-panic96:
Closes #3012 #1788


autoload/go/debug.vim
```diff
@@ -25,6 +25,16 @@ if !exists('s:start_args')
   let s:start_args = []
 endif
 
+if !exists('s:default_debug_mappings')
+  let s:default_debug_mappings = [
+    \["nmap", "<F5>",  "<Plug>(go-debug-continue)"],
+    \["nmap", "<F6>",  "<Plug>(go-debug-print)"],
+    \["nmap", "<F9>",  "<Plug>(go-debug-breakpoint)"],
+    \["nmap", "<F10>", "<Plug>(go-debug-next)"],
+    \["nmap", "<F11>", "<Plug>(go-debug-step)"],
```
kernel-panic96, line 34: This slightly differs from the default. I tested it with <buffer> but when I step into some file the buffer changes and the mappings are no longer active, thus I decided to leave it without the buffer attribute

kernel-panic96: Ideally the behavior should be that these mappings will be only active in a certain window or tab. E.g. I have the n key mapped to step over in debug mode and I usually have the n key mapped to next search result, I would like the semantics of n to be the default in other tabs. WDYT?

autoload/go/debug_mode.vim
```diff
@@ -0,0 +1,136 @@
...
+      let command = join([l:merged_lhs_to_map_cmd[lhs], lhs, rhs])
+      execute command
+    endfor
+  endfor
+endfunction
+
+function! go#debug_mode#Restore(...) abort
+  for [lhs, save] in items(s:mappings_save)
+    let command = s:restore_mapping(lhs, save)
+    silent! execute command
```
bhcleek, line 71: use `execute()` here instead of normal mode commands.

kernel-panic96:
The CI fails with:

> --- FAIL Test_GoTermReuse (0.155905s)
vim-go: [go run '/tmp/vim-go-test/testrun/0cd3f7c6e22668318f531ac9655e91b8e94221ce94fb1355844e31d6066c7407/src/term/term.go'] FAIL
        function Test_GoTermReuse line 20: Expected 3 but got 4
        function Test_GoTermReuse line 26: Expected 3 but got 4
FAIL term_test.vim              0.481729s / 3 tests

At a first glance this doesn't seem to me related to the changes in the PR

bhcleek (commented):
Making the mappings local to the buffer is critical.

Thanks for contributing @kernel-panic96  🙇 

Can you put the debug_mode.vim contents into `debug.vim` and make the public functions script-scoped functions instead?

Please don't change the location of where the mappings are applied and cleared; debugging is surprisingly finicky, and the placement of those operations has been a deliberate choice to make sure that mappings are available at the right times in the debugging lifecycle. I suspect this means that the mappings list will need converted to a dictionary where each operations' name is used as a key (e.g. `{'go-debug-continue': ["nmap", "<F5>"]}`, though I think that the value may be able to just be the keybindings; I can't think of a reason to use anything other than nmap).

The problems that you had with using buffer-local mappings is because the `doautocmd` is no longer applied. If you'll make the changes I requested in the previous paragraph, we should be able to make the changes buffer-local when the user's preferred keys are processed.

Also, please do not remove the autogroup. Instead, leverage it as it was leveraged before, just with with customizations.

Please use `execute()` instead of `:execute` in VimL.

If you'd prefer to be done with this PR, I can take over and make the changes I've requested.

bhcleek:
> At a first glance this doesn't seem related to the changes in the PR
> EDIT: Is there a way to restart the CI? It passed the first time.

Those failures seem unrelated to your changes so far. That failing test has been flaky recently. I thought I'd resolved it in a recent PR, though; perhaps you're work is based on an older commit? If so, that's ok, there's no need to rebase or merge master into your branch.

kernel-panic96:
I'm fine with you finishing it up. I tried and ran into some race conditions where the backing of user mappings happens twice and when you try to restore them they get restored to the debugger mappings which is not what you would want.

Currently the code is structured with the assumption that the setup & teardown would be executed only once. Making <buffer> work by executing it continually requires a violation of this assumption. I tried adding some guards to backup the user mappings if they weren't backed up before, but I ran in some race conditions. I'll try to test it before the merge

bhcleek:
> I'm fine with you finishing it up.

Ok, I'll take over from here and make some changes. Are you ok if I push to your branch?

kernel-panic96:
> > I'm fine with you finishing it up.
> 
> Ok, I'll take over from here and make some changes. Are you ok if I push to your branch?

Yes.

kernel-panic96:
I'm willing to give it another try. Do you have unpushed code?

bhcleek:
I just finished up the major refactor last night. I need to do some testing locally, and then I'll push changes; I'll likely do that today or tomorrow.

Thank you for offering to test; that will be very helpful.

bhcleek:
@kernel-panic96 I've created a pull request against your branch. Please don't merge it to your branch regardless of your testing results. After you test, I can push directly to your branch instead and then merge this if your testing is successful.

bhcleek:
@kernel-panic96 I'm going to rebase this onto master and push to your branch before I merge.

bhcleek:
Thank you @kernel-panic96 !

kernel-panic96:
Thank you for the collaboration @bhcleek . It was awesome 👍 .

PR #3053, opened 2020-10-17
lsp: support gopls local per workspace

### LINKED ISSUE #3023: How to get buffer-local g:go_gopls_local effect
Is there a way to get same effect as this (from vim-go docs):
```vim
  autocmd FileType go let b:go_fmt_options = {
    \ 'goimports': '-local ' .
      \ trim(system('{cd '. shellescape(expand('%:h')) .' && go list -m;}')),
    \ }
```
when using `let g:go_imports_mode = 'gopls'`?

This works, but it isn't buffer-local, so when I've opened files from several projects at once it doesn't work correctly:
```vim
let g:go_gopls_local = trim(system('{cd '. shellescape(expand('%:h')) .' && go list -m;}'))
```

I've also tried to move this into `autocmd BufEnter *.go`… which is probably works but in general it feels wrong to change global vars this way and also it result in huge delay on first save in different buffer (I suppose it's because gopls re-initialize for new project specific to current buffer).

bhcleek:
Closes #3023

### DIFF
--- a/autoload/go/lsp.vim
+++ b/autoload/go/lsp.vim
@@ -1101,6 +1101,9 @@ function! go#lsp#AddWorkspaceDirectory(...) abort
   let l:workspaces = []
   for l:dir in a:000
     let l:dir = fnamemodify(l:dir, ':p')
+    if len(l:dir) > 1 && l:dir[-1:] == '/'
+      let l:dir = l:dir[:-2]
+    endif
     if !isdirectory(l:dir)
       continue
     endif
--- a/autoload/go/lsp/message.vim
+++ b/autoload/go/lsp/message.vim
@@ -342,7 +342,12 @@ function! go#lsp#message#ConfigurationResult(items) abort
     endif
 
     if l:local isnot v:null
-      let l:config.local = l:local
+      if type(l:local) is v:t_string
+        let l:config.local = l:local
+      elseif type(l:local) is v:t_dict
+        let l:workspace = go#path#FromURI(l:item.scopeUri)
+        let l:config.local = get(l:local, l:workspace, v:null)
+      endif
     endif
 
     if l:gofumpt isnot v:null
--- a/autoload/go/lsp/message.vim
+++ b/autoload/go/lsp/message.vim
@@ -273,6 +273,7 @@ function! go#lsp#message#ConfigurationResult(items) abort
 
   " results must be in the same order as the items
   for l:item in a:items
+    let l:workspace = go#path#FromURI(l:item.scopeUri)
     let l:config = {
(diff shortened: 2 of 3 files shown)


Shane-XB-Qian:
not sure if you noticed there was a regression flaw: 
for example: there some `a.go` `b.go` `c.go` gofiles, now `vi c.go`, why now it would loop to read each of them when startup?
though only 'c.go' left at the end, and looks 'a.go', 'b.go' was 'bd' or 'bw' at the end.
but that's annoyed, unless disabled the 'gopls' when vim startup (let 'g:go_gopls_enabled=0'). but how to enable it then/later?

Shane-XB-Qian:
@bhcleek actually i am not very sure it caused by this pr, but looks more or less ..

bhcleek:
The behavior you're seeing is because of this change 
image
in https://github.com/fatih/vim-go/pull/3052/files#diff-44b441a3a765b097802fe5a14def2b19f77ebc7c9727405849248396a128676dR251.

What makes you refer this as a regression? Are you seeing some kind of problem or are you just seeing something unfamiliar?

> but how to enable it then/later?

Can you clarify what you mean by this?

Shane-XB-Qian:
i did not dig those code detail, but why it needed loop to read all gofiles under current dir to a / cur vim window ?


> > but how to enable it then/later?
> 
> Can you clarify what you mean by this?

if no solution/workaround from you, perhaps i needed to disable gopls at startup by `let 'g:go_gopls_enabled=0'`, but if so, so far i do not know how to start it later when vimenter done.

Shane-XB-Qian:
and `let g:go_diagnostics_level=0` and/or `let g:go_diagnostics_enabled=0` did not help this, still annoyed loop to read into vim window. :(

bhcleek:
Give #3056 a try; I think you'll get the behavior you want.

Shane-XB-Qian:
yes, i tried, i works after that pr.
but needed to set `g:go_diagnostics_level=0` too, ('g:go_diagnostics_level' replaced 'g:go_diagnostics_enabled')
looks current default value is `2`, perhaps should/better be default with `0` ?

Shane-XB-Qian:
since `g:go_diagnostics_enabled` default was `0`, if so, user did not need to change something, otherwise, looks meaningless to debug 'surprise' by them ... :)

bhcleek:
Ah yes, I did not mean to change the default. PR coming up.

bhcleek:
@Shane-XB-Qian let's move conversation to #3057

Shane-XB-Qian:
thanks. wish #3056 #3057 merged soon.

PR #3054, opened 2020-10-17
l:efm must be defined with a value before it can be used

WhoIsSethDaniel:


WhoIsSethDaniel:
Hi @bhcleek. I started getting an error about l:efm after I pulled recent master.  It looks like the line I have added was removed in a recent change. I've simply added it back.

bhcleek (approved):
Thanks!

PR #3057, opened 2020-10-18
config: disable diagnostics by default

bhcleek:
Disable diagnostics by default so that the default behavior is preserved
to be as it was before g:go_diagnostics_level was introduced.


Shane-XB-Qian:
i check the change a bit, the #3056 and doc you modified, so far looks ok. thx.

PR #3097, opened 2020-11-17
Delete a param that is removed in Python 3.9

delphinus:
https://docs.python.org/3/library/json.html#json.loads

`encoding` option in `json.loads` has been removed in Python 3.9. This param has been being deprecated from long time ago (since 3.1), so it should not affect users.


bhcleek:
Thank you!

PR #3125, opened 2020-12-31
debug: add explicit preserve layout setting

### LINKED ISSUE #3020: Debugger caused the quitting of some other plugins
### What did you do?
I installed Vim-go and other plugins, such as Nerdtree for my vim.
I entered my vim with all of plugins executed as expected, such as Nerdtree.
Then I typed `:GoDebugStart .` to try to debug with Vim-go.

Before typing `:GoDebugStart .`, one window for the file `main.go` and one window for the plugin Nerdtree are opened. The cursor is in the `main.go`.
image

After executing `:GoDebugStart .`, the window of Nerdtree disappeared and the cursor is still in the `main.go`.
image

Then when I execute `:GoDebugStop`, the window of Nerdtree won't come back:
image

When I type `:GoDebugStart .`, the window of Nerdtree shouldn't be closed. Or, when I type `:GoDebugStop` to quit the debugger of Vim-go, I was expecting that the window of Nerdtree could come back.

### What happened instead?
When I type `:GoDebugStart .`, the window of Nerdtree is closed. When I type `:GoDebugStop` to quit the debugger of Vim-go, the window of Nerdtree won't come back.

#### vim-go version:
commit ef23d90b84231759a992adf79b0dcd72844c1f84
####  `vimrc` you used to reproduce:

<details><summary>vimrc</summary><br><pre>
    
    set rtp+=~/.vim/bundle/Vundle.vim
    call vundle#begin()

    " let Vundle manage Vundle, required
    Plugin 'VundleVim/Vundle.vim'

    " options
    "Plugin 'Valloric/YouCompleteMe'
    Plugin 'scrooloose/nerdtree'
    Plugin 'fatih/vim-go'
    Plugin 'ctrlpvim/ctrlp.vim'
    Plugin 'scrooloose/nerdcommenter'
    Plugin 'majutsushi/tagbar'
    Plugin 'vim-airline/vim-airline'
    Plugin 'vim-airline/vim-airline-themes'
    Plugin 'rstacruz/sparkup', {'rtp': 'vim/'}
    Plugin 'Yggdroot/indentLine'
    Plugin 'jiangmiao/auto-pairs'
    Plugin 'tpope/vim-repeat'
    Plugin 'YvesZHI/vim-code-dark'
    Plugin 'YvesZHI/killersheep'
    call vundle#end()

</pre></details>

#### Vim version (first three lines from `:version`):

VIM - Vi IMproved 8.2
#### Go version (`go version`):

go version go1.14.6 linux/amd64
#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GO111MODULE="auto"
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/zyh/.cache/go-build"
GOENV="/home/zyh/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/zyh/go"
GOPRIVATE=""
GOPROXY="https://goproxy.cn,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build308368298=/tmp/go-build -gno-record-gcc-switches"

</pre></details>

#### gopls version
<details><summary><code>gopls version</code> Output:</summary><br><pre>

golang.org/x/tools/gopls 0.4.3
</pre></details>

patapra:
Adds a new setting, `g:go_debug_preserve_layout`, which, when explicitly set, will preserve the current window layout, irrespective of the value of `g:go_debug_windows`. Default value keeps current behavior (does not preserve). Fixes #3020.

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -212,7 +212,7 @@ function! go#config#DebugWindows() abort
 endfunction
 
 function! go#config#DebugPreserveLayout() abort
-  return get(g:, 'go_debug_preserve_layout', -1)
+  return get(g:, 'go_debug_preserve_layout', 0)
 endfunction
 
 function! go#config#DebugAddress() abort
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -211,6 +211,10 @@ function! go#config#DebugWindows() abort
 
 endfunction
 
+function! go#config#DebugPreserveLayout() abort
+  return get(g:, 'go_debug_preserve_layout', -1)
+endfunction
+
 function! go#config#DebugAddress() abort
   return get(g:, 'go_debug_address', '127.0.0.1:8181')
 endfunction
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -459,10 +459,7 @@ function! s:start_cb() abort
   let l:debugwindows = go#config#DebugWindows()
   let l:debugpreservelayout = go#config#DebugPreserveLayout()
 
-  " If g:go_debug_preserve_layout is not explicitly set, for backwards
-  " compatibility defer to g:go_debug_windows for :only behavior (ie if
-  " empty, preserve layout).
-  if (l:debugpreservelayout == -1 && !empty(l:debugwindows)) || l:debugpreservelayout == 0
+  if !(empty(l:debugwindows) || l:debugpreservelayout)
     silent! only!
   endif
 
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -305,10 +305,26 @@ function! go#debug#Stop() abort
   else
     wincmd p
   endif
-  silent! exe bufwinnr(bufnr('__GODEBUG_STACKTRACE__')) 'wincmd c'
-  silent! exe bufwinnr(bufnr('__GODEBUG_VARIABLES__')) 'wincmd c'
-  silent! exe bufwinnr(bufnr('__GODEBUG_OUTPUT__')) 'wincmd c'
-  silent! exe bufwinnr(bufnr('__GODEBUG_GOROUTINES__')) 'wincmd c'
+
+  let stackbufnr = bufnr('__GODEBUG_STACKTRACE__')
+  if stackbufnr != -1
+    silent! exe bufwinnr(stackbufnr) 'wincmd c'
+  endif
+
+  let varbufnr = bufnr('__GODEBUG_VARIABLES__')
+  if varbufnr != -1
+    silent! exe bufwinnr(varbufnr) 'wincmd c'
+  endif
+
+  let outbufnr = bufnr('__GODEBUG_OUTPUT__')
+  if outbufnr != -1
+    silent! exe bufwinnr(outbufnr) 'wincmd c'
+  endif
+
+  let gorobufnr = bufnr('__GODEBUG_GOROUTINES__')
+  if gorobufnr != -1
+    silent! exe bufwinnr(gorobufnr) 'wincmd c'
+  endif
 
   if has('balloon_eval')
     let &ballooneval=s:ballooneval
@@ -441,7 +457,8 @@ endfunction
 function! s:start_cb() abort
   let l:winid = win_getid()
   let l:debugwindows = go#config#DebugWindows()
-  if !empty(l:debugwindows)
+  let l:debugpreservelayout = go#config#DebugPreserveLayout()
+  if (empty(l:debugwindows) && l:debugpreservelayout == -1) || l:debugpreservelayout != 1
     silent! only!
   endif
 
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -458,7 +458,11 @@ function! s:start_cb() abort
   let l:winid = win_getid()
   let l:debugwindows = go#config#DebugWindows()
   let l:debugpreservelayout = go#config#DebugPreserveLayout()
-  if (empty(l:debugwindows) && l:debugpreservelayout == -1) || l:debugpreservelayout != 1
+
+  " If g:go_debug_preserve_layout is not explicitly set, for backwards
+  " compatibility defer to g:go_debug_windows for :only behavior (ie if
+  " empty, preserve layout).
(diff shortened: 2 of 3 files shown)


autoload/go/debug.vim
```diff
@@ -305,10 +305,26 @@ function! go#debug#Stop() abort
   else
     wincmd p
   endif
-  silent! exe bufwinnr(bufnr('__GODEBUG_STACKTRACE__')) 'wincmd c'
-  silent! exe bufwinnr(bufnr('__GODEBUG_VARIABLES__')) 'wincmd c'
-  silent! exe bufwinnr(bufnr('__GODEBUG_OUTPUT__')) 'wincmd c'
-  silent! exe bufwinnr(bufnr('__GODEBUG_GOROUTINES__')) 'wincmd c'
+
+  let stackbufnr = bufnr('__GODEBUG_STACKTRACE__')
```
patapra, line 309: without this change, plugin windows were getting closed

autoload/go/config.vim
```diff
@@ -211,6 +211,10 @@ function! go#config#DebugWindows() abort
 
 endfunction
 
+function! go#config#DebugPreserveLayout() abort
+  return get(g:, 'go_debug_preserve_layout', -1)
```
bhcleek, line 215: Please make the default value `0`, as the documentation you added indicates it should be.

autoload/go/debug.vim
```diff
@@ -441,7 +457,8 @@ endfunction
 function! s:start_cb() abort
   let l:winid = win_getid()
   let l:debugwindows = go#config#DebugWindows()
-  if !empty(l:debugwindows)
+  let l:debugpreservelayout = go#config#DebugPreserveLayout()
+  if (empty(l:debugwindows) && l:debugpreservelayout == -1) || l:debugpreservelayout != 1
```
bhcleek, line 461: `if empty(l:debugwindows) || l:debugpreservelayout`

doc/vim-go.txt
```diff
@@ -2490,7 +2490,15 @@ Show only variables on the right-hand side: >
   let g:go_debug_windows = {
         \ 'vars':  'rightbelow 60vnew',
   \ }
+<
+
+                                                *'g:go_debug_preserve_layout'*
 
+Preserve window layout in debugging mode. If explicitly set, it will override
+the preserving behavior when |'g:go_debug_windows'| is empty.
+>
```
bhcleek, line 2499: As worded, this sounds like the inverse of what it does. I believe it should indicate that it preserves window layout in debugging mode regardless of the value of g:go_debug_windows.

Also, there should be no need for a user to explicitly set it. Making the default value `0` should be sufficient and results in a user experience consistent with how vim-go works.

bhcleek (changes requested):

doc/vim-go.txt
patapra: I agree that the wording can be improved. However, a default of 0 would conflict with setting `g:go_debug_windows = {}` and break backwards compatibility.

The problem is this new setting controls the same window option (`:only`) as `g:go_debug_windows`, so there needs to be a way to determine precedence when their values indicate conflicting behavior. The default value of -1 (unset) makes this new setting backwards compatible because -1 means defer to `g:go_debug_windows` for `:only` behavior ie if `g:go_debug_windows` is empty, preserve layout. If not -1, the user explicitly set the value and it should take precedence.

This also allows it to work in conjunction with `g:go_debug_windows` - a user can set `g:go_debug_windows = {}` and `g:go_debug_preserve_layout = 0`, which would not display any debug windows and also not preserve their layout (I'm not sure why someone would want this, but the settings behavior would be consistent nonetheless).

g:go_debug_windows | g:go_debug_preserve_layout | only
---- | ---- | ----
{} | -1 | noonly
{} | 0 | only
{} | 1 | noonly
non-empty | -1 | only
non-empty | 0 | only
non-empty | 1 | noonly

autoload/go/debug.vim
patapra: this is the condition for calling `only` ie not preserving the layout, so it needs to be flipped. i left off the `!` on `empty(` and fixed that tho

autoload/go/config.vim
patapra: see comment below

doc/vim-go.txt
bhcleek: I think there was a typo in my original request. The logic should be `if !(empty(l:debugwindows) || l:debugpreservelayout)`. With that, there's no need to determine whether `g:go_debug_preserve_layout` was explicitly set.

When `go#config#DebugWindows()` returns an empty dict, then the current behavior on master is correct and should be preserved: `only` is not used. What you've asked for is a way to also skip `only` when `l:debugwindows` is not empty; the option you've introduced is useful for that purpose, but there's no need to determine whether the user explicitly set it; it only needs to be considered when `l:debugwindows` is not empty.

g:go_debug_windows | g:go_debug_preserve_layout | only
---- | ---- | ----
{} | 0 | noonly
{} | 1 | noonly
non-empty | 0 | only
non-empty | 1 | noonly

i.e. `g:go_debug_preserve_layout` should only be considered when `l:debugwindows` is not empty.

patapra:
I don't understand the test failures. Looks like maybe a temporary test runner problem?

bhcleek:
That one failure looks like it was a problem with a flaky test. I've re-run them all.

Thanks for contributing to vim-go!

bhcleek (approved):

bhcleek (approved):

PR #3136, opened 2021-01-15
Fix GoDef E474: Invalid argument

tacahiroy:
This fixes an error when you execute `:GoDef` in Vim 8.2.2324 or newer.
Actually, the error is due to some changes in Vim itself here: https://github.com/vim/vim/commit/6f02b00bb0
The commit makes `lnum` and `col` for `cursor()` function has to be an integer if it's not a list; it was ok to pass string values for that.
You can confirm this behaviour easily like this:

```vim
" this doesn't work Vim 8.2.2324 or newer, but works with Vim 8.2.2323 or older
:cal cursor('1', '1')

" Interestingly this works, it could be a bug may be?
:cal cursor(['1', '1'])

" {lnum} and {col} have to be an integer in >= Vim 8.2.234
:cal cursor(1, 1)
```

```
vim-go: [definition] SUCCESS
"/usr/local/bin/go1.15.5/src/flag/flag.go" [readonly] 1045L, 34608B
Error detected while processing function 271[13]..273[37]..function 271[13]..273[26]..<SNR>104_definitionHandler[8]..<SNR>179_jump_to_declaration_cb[5]..go#def#jump_to_declaration:
line   92:
E474: Invalid argument
Error detected while processing function 271:
line   13:
E171: Missing :endif
```

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -138,10 +138,10 @@ function! go#def#jump_to_declaration(out, mode, bin_name) abort
   let ident = 0
   let filename = parts[0]
   if len(parts) > 1
-    let line = parts[1]
+    let line = str2nr(parts[1])
   endif
   if len(parts) > 2
-    let col = parts[2]
+    let col = str2nr(parts[2])
   endif
   if len(parts) > 3
     let ident = parts[3]


bhcleek:
Thank you. I'd like to keep an eye on Vim before merging this; assuming your description is accurate, Vim's new behavior sounds like a regression that I'd expect to be remedied soon.

If it's not remedied, then I'll need to take a deeper dive through vim-go code to address any other areas that have this same problem.

tacahiroy:
I created an issue about this behaviour.
The issue was closed as it was the same behaviour as before, but I have confirmed some incompatible changes were made in recent version of Vim. Let's see how it goes.
https://github.com/vim/vim/issues/7686#issuecomment-761591031

bhcleek:
Vim 8.2..2363 purports to fix the problem. Can you confirm so we can close this without merging?

tacahiroy:
I have confirmed that `GoDef` works with d68fc4d version of vim-go with Vim below:
```console
$ vim --version | head -2
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jan 17 2021 23:24:27)
Included patches: 1-2368
```

So this patch needn't to be merged any more. closing.

justgoodman:
I can't update vim to version more than 8.2.2363 on MacOS using brew. @bhcleek thank you for you commit it is really help 👍

PR #3141, opened 2021-01-22
Fix <Plug>(go-callers)

Felixoid:
Currently, it's broken:

`E117: Unknown function: go#guru#Callers`


bhcleek:
Thank you

Felixoid:
Wow, that was quick. Thank you :-j

PR #3143, opened 2021-01-23
Make go-debug-stepout available for debug mapping

csch0:
Now go-debug-stepout could be used like this:

`let g:go_debug_mappings = { '(go-debug-stepout)': {'key': '<F12>'} }`


bhcleek:
Thank you for contributing!

PR #3146, opened 2021-01-26
Fix: Ensure filetype is set for Go templates 

PaulDugdale:
*.tmpl is already detected and a filetype is already set here:
https://github.com/vim/vim/blob/e0e3917554327f2524066f89fbbef9c83c1535da/runtime/filetype.vim#L744

Using `setfiletype` prevents this from being overridden, and the filetype is left as "template" and not set to "gohtmltmpl" as intended.

Screenshot 2021-01-26 at 15 07 18

### DIFF
--- a/ftdetect/gofiletype.vim
+++ b/ftdetect/gofiletype.vim
@@ -7,7 +7,7 @@ set cpo&vim
 " Note: should not use augroup in ftdetect (see :help ftdetect)
 au BufRead,BufNewFile *.go setfiletype go
 au BufRead,BufNewFile *.s setfiletype asm
-au BufRead,BufNewFile *.tmpl setfiletype gohtmltmpl
+au BufRead,BufNewFile *.tmpl set filetype=gohtmltmpl
 au BufRead,BufNewFile go.sum set filetype=gosum
 
 " remove the autocommands for modsim3, and lprolog files so that their


bhcleek:
I'm not sure we should override what Vim is doing with the filetype. 🤔

PaulDugdale:
Thanks! Here's some more info...

With the vim-go plugin installed, I expect the syntax and indentation to work correctly on Go text/html template files.
This isn't the case, the standard Vim html syntax/indentation will be used.

The problem is that there is already a [template](https://github.com/vim/vim/blob/e0e3917554327f2524066f89fbbef9c83c1535da/runtime/syntax/template.vim) filetype in vim, that is a simple wrapper around the html filetype.

Because of this, and the fact that`setfiletype` is used, the `gotexttmpl` and `gohtmltmpl` filetypes will never be used unless they are forced in your .vimrc

This used to work until [this change](https://github.com/fatih/vim-go/commit/1ed9b521929606a282edc72c7627addb8c3914b9#diff-72fbf82c2e95f9df47c529c364f1aaad9b987d4c52e0a1d4132f5a4e30b87305R10)

bhcleek:
My concern is that there may be a lot of people that do edit html template files and that use `.tmpl` as an extension for those files.  Since vim-go sets the template filetype based on a wildcard pattern (as opposed to a specific file name like `go.sum` uses), I'm reluctant to unconditionally override.

I think I'd prefer a solution that changes vim-go to use `setfiletype` with a different extension (e.g. `*.go.tmpl`, `*.gotmpl`,  etc.) or that users that really want this put the `au BufRead,BufNewFile *.tmpl set filetype gohtmltmpl` in their own vimrc file.

PaulDugdale:
I understand the concerns, but I respectfully disagree...

The `.tmpl` has been given "official" blessing in both the Golang examples, and the Golang team maintained VSCode extension, so I think it's better to use `.tmpl` than an alternate extension.
I also think that leaving it in it's current state will break syntax/indentation for more users than changing it would. It's a very common extension inside of Go, and a very unusual one outside of it.

bhcleek:
Since the default filetype for .tmpl files is template and since it seems to be a subset of what vim-go uses for the the .tmpl, this change is appears to be purely additive and reasonably safe. I'm not super keen on doing this unconditionally, but let's see if anyone complains.

PR #3157, opened 2021-02-14
fix shell commands debug message for GoRun

mudhoney:
If I set ` let g:go_debug=['shell-commands']` for debugging shell commands, 
`:GoRun` fails like below

```
Error detected while processing function go#cmd#Run:
line   78:
E730: using List as a String
E116: Invalid arguments for function go#util#EchoInfo
```
this PR hopefully tries to fix it


bhcleek:
Thank you

PR #3172, opened 2021-03-10
Add GoCyclo command

EVODelavega:
Just a crude implementation. Added a quick `:GoCyclo` command to list the cyclomatic complexity for a given file to the plugin. I'm happy to work on this some more, adding things like `-top N` and `-over N` support, perhaps being able to get the complexity for the entire package, or instead of displaying just the quickfix buffer, perhaps add some type of bookmarks to the worst functions etc..

I just wanted to get some eyes on this before I started investing too much time on this stuff. I get that cyclomatic complexity is somewhat of a grey area, especially in golang, but I figured if enough people thought that this would be a worthwhile addition, I might as well go ahead and implement the command.

### DIFF
--- a/autoload/go/cyclo.vim
+++ b/autoload/go/cyclo.vim
@@ -1,0 +1,32 @@
+" don't spam the user when Vim is started in Vi compatibility mode
+let s:cpo_save = &cpo
+set cpo&vim
+
+function! go#cyclo#Cyclo() abort
+  let l:cmd = ['gocyclo', bufname('')]
+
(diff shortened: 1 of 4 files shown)


bhcleek:
Thank you for contributing. I'm going to decline, though, for all the reasons you might imagine 😁

EVODelavega:
Fair enough, looking back at this now, I'm not quite sure where my mind was at when I opened this PR :stuck_out_tongue_closed_eyes:

PR #3173, opened 2021-03-10
Add basic GoDebugConnect command

EVODelavega:
Just noticed this issue, and seeing as I'll probably need to connect to a container when debugging soon, I went ahead and added the suggested `GoDebugConnect` command. It's untested ATM, but thought I'd push it to see if this works (yay, testing in production).

### DIFF
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -760,6 +760,8 @@ function! go#debug#Start(mode, ...) abort
     elseif a:mode is 'attach'
       let l:cmd = add(l:cmd, a:1)
       let s:state['kill_on_detach'] = v:false
+    elseif a:mode is 'connect'
+      let l:cmd = add(l:cmd, a:1)
     else
       call go#util#EchoError('Unknown dlv command')
     endif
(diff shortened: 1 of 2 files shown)


bhcleek:
Thanks!

I think this won't quite work, but it's close. The listen address and maybe `--headless` are probably not relevant to connect.

I'll try to test in the next few days

EVODelavega:
@bhcleek I know this is most likely not the complete solution. I'm a bit swamped at my dayjob, but I expect to be needing some dlv debugging with containers in the next 1~2 weeks. I'm planning to use the opportunity to implement this properly, or at least document/update `:GoDebugAttach` in a way that would address the problems reported in the linked issue.

I've been using vim-go for years, so I'm all too happy to contribute.

bhcleek:
@EVODelavega I've been playing with this a bit and there's quite a bit more to do to make this work. I'm close to having a working solution, but I wanted to point out that if you're using Linux and the container you need to debug is on your same machine, then you can use `:GoDebugAttach` to attach to the process.

You can also use `dlv connect` at the commandline and use the CLI interface to debug. It's really not terribly onerous to do vs updating vim-go if this is a rare operation for you.

bhcleek:
@EVODelavega I have working implementation locally. Would you like for me to push to your branch, or open a separate PR for it?

bhcleek:
@EVODelavega do you want to give https://github.com/fatih/vim-go/pull/3179 a try? It has your commit here, plus the required additional changes. I'm happy to close that out and push the changes to your branch if you'd prefer.

Issue #2769, opened 2020-03-13
Go/golang auto-indentation not working as expected: left aligns in some cases


IAXES:
### Pre-amble:
I'm not 100% certain if this is an issue within `vim-go` or `vim` in general, given somewhat similar posts I've read on here (can't find the exact ones at the moment) that imply an issue in `vim` specifically. Presumably `vim-go` overrides `vim`'s default behavior for Go/golang buffers, so I'll start here.

### What did you do?

I setup a Docker test image (instructions plus files will be included at the bottom of this submission), and am using it to write/debug Go/golang code. If I, for example, write in code that represents a function call, and I manually insert a newline between the parentheses (i.e. **`(`, `)`** ) characters, I would normally expect the brackets to align with the beginning of the function name, or otherwise honor the `cindent` plus `cinoptions` configuration in my `.vimrc`, which appear to be honored for other file types (i.e. `.c` and `.h` for C, `.hpp` and `.cpp` for C++, `.py` for Python 2/3, etc.).

For example, here's a simple hello-world snippet I'm working on.
![image](https://user-images.githubusercontent.com/8325672/76652148-370bc380-6523-11ea-864e-7020fe3aa0ca.png)

If I were to insert a comma plus newline literal after `world"` and before the trailing `)`, I normally expect (due to my `cinoptions`) for the `)` character to align with the opening `(` from the line above. If `cinoptions` isn't specified, I'd expect it to at least align with the first non-tab character of the line above.

Instead, this is what happens:

![image](https://user-images.githubusercontent.com/8325672/76652279-7cc88c00-6523-11ea-8cd1-247de96c7c69.png)

There are lots of other examples I could provide, but it seems the recurring them is that auto-indentation, either when typical code normally, or re-indenting a file (i.e. `gg=G`) results in auto-indent not being invoked (at least, no as expected), and `cinoptions` likely also being ignored.

### What did you expect to happen?

`cinoptions` to be honored, or for "prettier" auto-indentation to be generated.

### What happened instead?

Error case noted above: auto-indentation does not appear to be working.

**Note:** I've repeated this test with/without other plugins (i.e. Ale, YCM), and with/without `cinoptions` being set in my `.vimrc`, and in all cases, I encountered the exact same error case described at the beginning of this issue/ticket.

#### vim-go version:

**HEAD/master** (installed via `vundle`):

```sh
[...]

PR #3178, opened 2021-03-18
Fixed indent for functions with parameters on one line

macmv:
I was having the same issue as #2769, which was bugging me. I do not use `gofmt`, and I would prefer it if live indents worked as they should. This function makes sure that indents always work as expected, no matter what comments/arguments exist on the previous/current line. It's not the most performant function, but I couldn't find a better way with regex, so iterating through the whole line seemed like the best option to me.

I understand that I should use `gofmt`, but I couldn't be bothered to deal with tabs in vim. I know this issue is partially fixed when using `gofmt`, but that only applies when you save. This PR fixes the indents all the time, and makes the code you're in the middle of writing much cleaner.

### DIFF
--- a/indent/go.vim
+++ b/indent/go.vim
@@ -53,7 +53,10 @@ function! GoIndent(lnum) abort
     endif
   endfor
 
-  if prevl =~ '[({]\s*$'
+  let num_parenthases = s:CountIndent(prevl, '(', ')')
+  let num_brackets = s:CountIndent(prevl, '{', '}')
+
+  if num_parenthases > 0 || num_brackets > 0
     " previous line opened a block
     let ind += shiftwidth()
   endif
@@ -63,7 +66,10 @@ function! GoIndent(lnum) abort
   endif
   " TODO: handle if the previous line is a label.
 
-  if thisl =~ '^\s*[)}]'
+  let num_parenthases = s:CountIndent(thisl, '(', ')')
+  let num_brackets = s:CountIndent(thisl, '{', '}')
+
+  if num_parenthases < 0 || num_brackets < 0
     " this line closed a block
     let ind -= shiftwidth()
   endif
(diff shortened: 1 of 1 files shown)


bhcleek:
Thank you for contributing!

As explained in #2769, trying to do this will result in wrong indentation within raw strings that span multiple lines. Given the general correctness of vim-go's current implementation and the usefulness of `gofmt`, I'm disinclined to merge this. `gofmt`'s style is the community standard for Go code.

If you'd like to work on an change to format with gopls as one types, I would happily review that PR and help you get it over the finish line.

macmv:
`gopls` looks like it would be difficult to format while typing, without being very slow. Other than building a custom binary that can accept commands over stdin, and return indent level over stdout (or something along those lines), I can't think of any way to make `gopls` fast. I'm not really up to making that, so I'll just leave those closed. Thank you for your reply!

macmv:
Actually, I was just writing code with `vim-go`, and I didn't notice any indent in the middle of multi-line strings. The only indent I noticed was the fact that hitting return added extra spacing that was the same as the previous line. It didn't add any indent when I put an open brace in a multi-line string. Is that the indented effect, or were you expecting the indent to always go the start of the line every time you hit enter? Because I prefer it the way it is. I don't see how gopls would make this any better.

bhcleek:
If you do something like this with a raw string, I think you'll see the concern that I have:

```
var s = `thi is a string
with a second line and then a parenthetical (a parenthetical
is a real thing).`
```

I think the performance with gopls would be pretty fast. You'd use autocmds to send the text on a line to gopls, send the format request, and process the response.

macmv:
With this change, there is no indent there. For some reason I thought I would need to spawn a new gopls process for every line, but I now realize that is not how gopls works. I will start work on that as a different PR.

bhcleek:
There are several examples in `autoload/go/lsp.vim` of waiting for the response to be received and processed. Look for `await()` (e.g. https://github.com/fatih/vim-go/blob/master/autoload/go/lsp.vim#L1499-L1501)

PR #3185, opened 2021-03-26
snippets: add a colon after case

kit494way:


bhcleek:
Thank you for contributing!

None of the test failures are due to this change.

PR #3188, opened 2021-04-01
Don't treat `[[...]]` as template syntax region

alker0:
[The official template doc](https://pkg.go.dev/text/template)
> The input text for a template is UTF-8-encoded text in any format. "Actions"--data evaluations or control structures--are delimited by "{{" and "}}"; all text outside actions is copied to the output unchanged. Except for raw strings, actions may not span newlines, although comments can.

Double angle brackets `[[` is not syntax for starting go template region.
This PR removes the highlight for it.


bhcleek:
Thank you

PR #3189, opened 2021-04-01
Fix Highlight for template's dot-identifier

alker0:
Currently, Dot-identifiers `{{ .foo.bar }}` are not highlighted.
https://github.com/fatih/vim-go/blob/bd89f4734555c60f53acaf21731bd43617e6e526/syntax/gotexttmpl.vim#L69
In character range `[a-z]` of vim pattern, it is necessary to use `[:blank:]` as spaces instead of `\s`.
Also, `+` needs escape for working as 'one or more' in vim magic pattern (default mode).
This PR fixes it.


bhcleek:
Thank you for contributing

The test failures are due to a golangci-lint change that I'm working on fixing and can be ignored for this PR.

PR #3191, opened 2021-04-02
highlight progress message with ModeMsg

mcpeng:


bhcleek:
Thank you for contributing. I don't see a compelling reason to change this, though.

PR #3194, opened 2021-04-03
Include sub-packages in golangci-lint command

orlangure:
With current vim-go version, executing metalinter using `golangci-lint` runs the following command:
```
vim-go: [golangci-lint] dispatched
vim-go: job command: ['/Users/yury/go/bin/golangci-lint', 'run', 
    '--print-issued-lines=false', '--build-tags', '', '--exclude-use-default=false', 
    '--disable-all', '--enable=vet', '--enable=golint', '--enable=errcheck', 
    '/Users/yury/go/src/github.com/orlangure/gnomock']
```

The last argument is the path of currently opened file (folder). When golangci-lint is executed with folder path, it only runs the linter on this folder, and not its subfolders. So this PR adds `/...` in the end so that all subpackages are checked as well.

golangci-lint version: v1.39.1

### DIFF
--- a/autoload/go/lint.vim
+++ b/autoload/go/lint.vim
@@ -6,7 +6,12 @@ function! go#lint#Gometa(bang, autosave, ...) abort
   let l:metalinter = go#config#MetalinterCommand()
 
   if a:0 == 0
-    let l:goargs = [expand('%:p:h')]
+    let l:filepath = expand('%:p:h')
+    if l:metalinter == 'golangci-lint'
+      let l:filepath = printf('%s/...', l:filepath)
+    endif
+
+    let l:goargs = [l:filepath]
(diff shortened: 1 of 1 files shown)


bhcleek:
Thank you for contributing. This would introduce inconsistencies with the other metalinters and with other commands. Vim-go tries hard to be consistent in its approach and to be faithful to Vim. Here are some thing to keep in mind:
* Vim is not an IDE. It's designed to work with files, not whole directory trees
* The other metalinters (e.g staticcheck) and tools (e.g. `errcheck`, `golint`, etc.) are also passed only the current package
* The vast majority of commands in vim-go operate either on the current file or the current package.

If this is the behavior you desire, then you can pass arguments to `:GoMetaLinter` to get what you want (e.g. `:GoMetaLinter ./...`)

PR #3199, opened 2021-04-11
Update hello_world.go

ujukiSakuta:
Added parenthesis to import block to simplify package addition


bhcleek:
@ujukiSakuta do you want to fix those tests that failed or should we close this?

PR #3210, opened 2021-04-24
Add a new snippet for t.Run

qeubar:
Adding a new snippet so that developers can easily start a `t.Run("",  func(t *testing.T){})` unit test.


bhcleek:
Thanks for contributing!

PR #3216, opened 2021-05-05
Fix error highlights for hex and binary constants

bluekeyes:
Add `)` as a terminating character to avoid marking sequences like:

```go
int64(a&0xFF00)>>8
```

as invalid constants. This matches the behavior for octal constants.

vim-go-hex-error


bhcleek:
Thank you for contributing!

PR #3218, opened 2021-05-07
Add `shadow` tool on goVet command

drgarcia1986:
Hi everyone, I want to check for shadowed variables on `:GoVet` command.
I don't know if it's the better way to do that (ok, I need to say: _"it's working on my machine"_), so I really appreciate if someone helps me.


bhcleek:
First, thank you for contributing.

There are a lot of valid cases for shadowing; it's not something we'd want to always use in vim-go.

You can get what you're looking for by installing the shadow tool and then using `:GoVet` already and passing the options you want (e.g. `:GoVet -vettool=shadow`).

PR #3234, opened 2021-05-23
Add commentstring to gohtml template files

jvcarli:
Currently gohtmltmpl files use html commentstrings . I've changed this option so [`{{/* comment */}}`](https://golang.org/pkg/text/template/#hdr-Actions) notation is used instead.

### DIFF
--- a/ftplugin/gohtmltmpl.vim
+++ b/ftplugin/gohtmltmpl.vim
@@ -4,4 +4,6 @@ endif
 
 runtime! ftplugin/html.vim
 
+setlocal commentstring={{/*%s*/}}
+
 " vim: sw=2 ts=2 et


bhcleek:
Thank you for contributing. What are you trying to accomplish with this change; what problem are you trying to solve?

This would fully replace the `commentstring` value and would abandon the html comments, but the html comments are also applicable, because a lot of Go templates are html templates.

`commentstring` only affects folding, so this change is relatively inconsequential and using `set foldmethod=syntax` may be a better option to allow you to fold templates if that's what you're after.

If you're trying to control how comments are indented in the middle of a Go template comment, then using a different value for the `comments` option may be what you're after.

Even if we don't make a change to vim-go, you can still get the behavior you're looking for by adding a `after/ftplugin/go.vim` file with the settings you want.

PR #3248, opened 2021-06-10
lint: use revive instead of golint

bhcleek:
##### lint: replace golint with revive in golangci-lint linters

Use revive with golangci-lint instead of golint; golint is officially
deprecated.

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -287,7 +287,7 @@ function! go#config#MetalinterEnabled() abort
 endfunction
 
 function! go#config#GolintBin() abort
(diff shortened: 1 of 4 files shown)


Rican7:
Any reason that you went with revive over staticcheck?

The golint repo even recommends staticcheck: https://github.com/golang/lint/commit/6edffad5e6160f5949cdefc81710b2706fbcd4f6

bhcleek:
revive is a tool "such as", vim-go already has support for staticcheck as a metalinter, and arguably staticcheck is more of a metalinter than revive is.

Rican7:
Ah, fair. Thanks!

PR #3249, opened 2021-06-13
Neovim packages instructions

mdvthu:
Plugin is compatible with neovim but installation instructions are slightly different


bhcleek:
Thank you for contributing.

PR #3268, opened 2021-08-01
Make README tutorial link more specific

ehershey:
There used to be a separate wiki link but I believe it got lost amid other edits. This makes the blurb about the tutorial link directly to it and link to the wiki itself in a separate reference.


bhcleek (approved):

bhcleek:
thank you

Issue #3272, opened 2021-08-06
Accepting a completion logs a weird message to statusline


daturkel:
### What did you do? (required: The issue will be **closed** when not provided)

In a go file, trigger a completion and then accept it. For example:
```go
package main

import "time"

func main () {
    time.S
}
```
Hitting `tab` after the `time.S` will pop up a suggestion for `time.Sleep`, and selecting it with enter inserts it.

### What did you expect to happen?

I expect the completion to be inserted and the statusline to remain unchanged.

### What happened instead?

The completion is inserted but an odd message gets written to the statusline:

```
vim-go: {"cid":1628215692,"source":"languageserver.golang","index":0}
```

Is there a way to disable this message?

####  `vimrc` you used to reproduce:

<details><summary>vimrc</summary>

```vim

```
</details>

#### Vim version (first three lines from `:version`):

NVIM v0.5.0
Build type: Release
LuaJIT 2.1.0-beta3

#### Go version (`go version`):

go version go1.16.6 darwin/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/danturkel/Library/Caches/go-build"
GOENV="/Users/danturkel/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/danturkel/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/danturkel/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.16.6/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.16.6/libexec/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.16.6"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/dev/null"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
[...]

PR #3274, opened 2021-08-09
return early from CompleteDone handler

bhcleek:
Return early from the CompleteDone handler when omnifunc is not set to
go#complete#Complete.

This should help interactions with other plugins like coc.nvim that some
users leverage to do completion.

Relates to #3272


daturkel:
I tested this out and confirmed it works as desired with coc in eliminating the issue in #3272 ! Thank you for the quick investigation and fix!

Issue #3284, opened 2021-08-19
Integrating other language servers is difficult due to structured hover result


aeneev:
I've had moderate success using non-gopls language server with vim-go, but am struggling with handling the structured hover responses.

[LSP spec](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_hover) defines hover response as plain text or markdown, but gopls supports structured response and vim-go is making use of that when [showing info](https://github.com/fatih/vim-go/blob/c34c73a4269857e694cda38431601ab753fcbc3f/autoload/go/lsp.vim#L1148), [docs](https://github.com/fatih/vim-go/blob/c34c73a4269857e694cda38431601ab753fcbc3f/autoload/go/lsp.vim#L987), etc.

This in turn makes vim-go misbehave when another language server is used.

Are there any plans on reducing dependencies on non-standard LSP features?
It appears that [textDocument/signatureHelp](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_signatureHelp) method might be able to cover some of the features encoded in structured hover responses.

bhcleek:
I don't plan to support every possible language server. When other language servers are desired, there are already great plugins facilitate using those with Vim, and trying to support others will inevitably result in an onslaught of issues to support various means of communication and idiosyncracies of other LSP servers. Even adding `gopls` support was met with dissent when it was originally considered.

aeneev:
I am not at all versed in the ecosystem of LSPs and go development tooling, so forgive my ignorance.

I understand that attempting to support every LSP and every extension employed by every LSP wouldn't be feasible. But isn't there some common denominator?

I'll look into the other plugins for making use of LSPs, but as my journey with Go started with vim-go, I've got used to it and would love to be able to continue using it.

I've had success with the PR #3285, and am glad that I can make use of our LSP and vim-go by changing path to the gopls binary and command line options, and implementing fallbacks when textDocument/hover response isn't json.

bhcleek:
If you'd like to use a different LSP implementation, I'd recommend using one of the LSP-specific plugins for Vim.

PR #3285, opened 2021-08-19
Support non-gopls LSP servers

aeneev:
This change adds g:go_gopls_binary option that specifies path to the LSP
binary.

Gopls uses custom format for hover response [1], and other LSPs do
not, this change adds bare-bones support for plain text hover responses
(no markdown support).  :GoDocBrowser is driven by fields in structured
responses and is not supported at all, and results in "could not find
path for doc URL" error.

Because non-gopls LSP server support is pretty much experimetal, the
options are not documented, and the option name uses "gopls" (as opposed
to "lsp") as to not create an impression that an arbitrary LSP server
would work out of the box.

Tested:
  Manually in my environment with our internal LSP server and confirmed that
  :GoDoc, :GoInfo, :GoDef work. Admittedly, :GoDoc and :GoInfo are pretty
  much the same, except the latter shows results in the command prompt,
  and the former opens up a doc window.

  `make` reports:
  All (vim-8.0) tests PASSED
  All (vim-8.2) tests PASSED
  All (nvim) tests PASSED

Issues:

  :GoInfo doesn't produce any output when the result has several lines.

Related: https://github.com/fatih/vim-go/issues/3284

[1] https://github.com/golang/tools/blob/bf6c7f26e9f0fea5b256f9ef6968435dddc5be25/internal/lsp/source/hover.go#L562

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -592,6 +592,10 @@ function! go#config#GoplsOptions() abort
   return get(g:, 'go_gopls_options', ['-remote=auto'])
 endfunction
 
+function! go#config#GoplsBinary() abort
+  return get(g:, 'go_gopls_binary', 'gopls')
+endfunction
+
 function! go#config#FillStructMode() abort
   return get(g:, 'go_fillstruct_mode', 'fillstruct')
 endfunction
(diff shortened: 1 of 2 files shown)


bhcleek:
Thank you for contributing.

Vim-go is purpose built for Go and will continue to rely on gopls. Unfortunately, I don't have the time to support arbitrary language servers and the different methods they may have to communicate with them, their varying options for starting launching them, or the different idiosyncrasies they may have.

aeneev:
No problem, thanks for reviewing and working on vim-go.

PR #3292, opened 2021-09-06
Feature: syntax highlight for //go:build directives

oz123:
The new style replacement for // +build directives is
now highlighted.

This is probably still missing, but it's a start. 

image

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -454,6 +454,10 @@ function! go#config#HighlightTypes() abort
   return get(g:, 'go_highlight_types', 0)
 endfunction
 
+function! go#config#HighlightBuildConstraintsNewFormat() abort
+  return get(g:, 'go_highlight_build_constraints_new_format', 0)
+endfunction
+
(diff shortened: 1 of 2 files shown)


bhcleek:
Thank you for contributing.

We don't need to create a new setting for this; we can lump it in with the existing `// +build` constraint highlighting and settings. Would you be willing to make those changes?

oz123:
Honestly, my vim script skills are non-existing. I don't know how to do it without guiding or a big time investment.
Can you guide me how to?

bhcleek:
It should be a matter of updating the regular expression in https://github.com/fatih/vim-go/pull/3292/files#diff-6c12071ba887bcb758b2bb5627e5bd19705a4d45bacff48d5b26e056869aa883R370 instead of doing your original batch of changes.

If this isn't something you can do, that's ok; I'll get to it soon.

oz123:
Thanks! I am fine with you taking over :-)

oz123:
Thank you for fixing this. Eventually, just extending the Regex did the job!

PR #3299, opened 2021-10-04
Revert to using lcd when changing directory.

ixru:
Don't know if there was some other reason for the change. Changing it back to what it was originally seems to work.


bhcleek (changes requested):
Thanks for the PR. Can you change this to fix the issue by using `execute printf('%s %s', cd, fnameescape(a:dir))` instead of reverting?

bhcleek:
Thank you

Issue #3307, opened 2021-10-18
fzf does not wrap in GoDecls


co60ca:
### What did you do? (required: The issue will be **closed** when not provided)

1. Opened a .go file
2. Used GoDecls to view the functions

(Images below)

### What did you expect to happen?

GoDecls would use fzf with a wrap to show long function lines.

### What happened instead?

fzf shows the long function lines truncated.

####  `vimrc` you used to reproduce:

<details><summary>vimrc</summary>

```vim
" Go - required for vim-go
filetype plugin indent on

" Go syntax highlighting
let g:go_highlight_fields = 1
let g:go_highlight_functions = 1
let g:go_highlight_function_calls = 1
let g:go_highlight_extra_types = 1
let g:go_highlight_operators = 1

" Auto formatting and importing
let g:go_fmt_autosave = 1
let g:go_fmt_command = "goimports"

" Status line types/signatures
let g:go_auto_type_info = 1

" Go Autocomplete
autocmd filetype go set listchars=tab:\ \ 
"                                     ctrl+space
autocmd filetype go inoremap <buffer> <C-@> <C-x><C-o>
autocmd filetype go set completeopt=menu,preview
autocmd filetype go set et!

" Cheatsheet

autocmd BufNewFile,BufRead *.go setlocal noexpandtab tabstop=4 shiftwidth=4
let g:go_auto_type_info = 1
let g:go_auto_sameids = 

let g:go_fmt_command="gopls"
let g:go_gopls_gofumpt=1

```
</details>

#### Vim version (first three lines from `:version`):

VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Aug 30 2021 06:51:40)
macOS version
Included patches: 1-2029

#### Go version (`go version`):

go version go1.16 darwin/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/<user>/Library/Caches/go-build"
GOENV="/Users/<user>/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/<user>/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/<user>/go"
GOPRIVATE=""
[...]

PR #3310, opened 2021-10-19
Added preview-window flag to GoDecls fzf for text wrapping

co60ca:
As reported in: https://github.com/fatih/vim-go/issues/3307

Found that, at least in my version, fzf doesn't automatically text wrap and you need a preview-window setting.

Concerns (as per https://github.com/fatih/vim-go/issues/3307) is that it may break existing users (assuming they have a version older than when fzf added `--preview-wrap`. I'm not sure if thats a real concern.

Ideal solution would be parameterization but I don't have any skills in vimscript so I'd need handholding.


bhcleek:
Thank you. I'll merge this soon.

In the meantime, you can always use `DEFAULT_FZF_OPTS` to set `--preview-window wrap`, too.

PR #3312, opened 2021-10-20
The time to wait for the connection to be made in milliseconds.

c77cc:
The time to wait for the connection to be made in
milliseconds.  A negative number waits forever.

The default is zero, don't wait, which is useful if a local
server is supposed to be running already.  On Unix Vim
actually uses a 1 msec timeout, that is required on many
systems.  Use a larger value for a remote server, e.g.  5000 msec.

### DIFF
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -609,7 +609,7 @@ function! s:connect(addr) abort
       return
     endif
   else
-    let l:ch = ch_open(a:addr, {'mode': 'raw', 'timeout': 20000, 'callback': l:state.on_data})
+    let l:ch = ch_open(a:addr, {'mode': 'raw', 'waittime': 5000, 'timeout': 20000, 'callback': l:state.on_data})
     if ch_status(l:ch) !=# 'open'
       call go#util#EchoError("could not connect to debugger")
       if has_key(s:state, 'job')


bhcleek:
The server should already be running by the time `ch_open` is called, and I've never seen a need to set `waittime`. Does this fix some bug that you've seen?

c77cc:
> The server should already be running by the time `ch_open` is called, and I've never seen a need to set `waittime`. Does this fix some bug that you've seen?

Yes, when I debug the remote process, the local vim will quickly return the connection failure error(could not connect to debugger); 
when I fix the waittime of ch_open, the connection failure problem disappears.

bhcleek:
What remote process are you debugging? I'd normally expect the dlv process that vim-go is communicating with to be local.

c77cc:
> What remote process are you debugging? I'd normally expect the dlv process that vim-go is communicating with to be local.

The code of the remote process (running on another physical machine) is:

```go
package main

import (
	"fmt"
	"time"
)

func main() {
	msg := `test_msg`

	fmt.Println(msg)

	time.Sleep(1 * time.Hour)
}
```

c77cc:
I did not set g:go_debug_address.
I did it in vim like this: GoDebugConnect {REAL_IP}:7001

Issue #3319, opened 2021-11-03
fzf wrap for GoImplements


kerma:
### What did you do?
I'm trying to quickly navigate in a larger project which has a lot of interfaces and many implementations to them. Usual workflow goes as:

* Use `CTRL-]` until I end up on an interface definition
* Trigger :GoImplements
* Navigate to locationlist, usually `CTRL+w j`
* Navigate to implementation line and open file in new buffer, something like `2j<CR>`

### What did you expect to happen?
GoImplements opens up a fzf popup window, like GoDecls

### What happened instead?
GoImplements can only open location or quickfix window

zouchao:
Hope `:GoCallers` will support it too

PR #3320, opened 2021-11-05
POC for fzf wrap for GoImplements

kerma:
This is a draft PR for #3319 

Code is largely based on GoDecls implementation.

It's not fully tested, needs some code cleanup and has a problem where fzf popup window is out of focus after GoImplements is triggered. Something to do with `go#lsp#Implements` async behavior I'm guessing.

EDIT: I haven't yet figured out the focus issue, but the insert mode is related to this. fzf calls [startinsert](https://github.com/junegunn/fzf/blob/e4c3ecc57e99f4037199f11b384a7f8758d1a0ff/plugin/fzf.vim#L915), but due to wrong focus this affects the main window.

https://user-images.githubusercontent.com/735108/140577115-26448ac3-d5fd-47a7-9a0a-24eb0373b144.mov

### DIFF
--- a/autoload/fzf/decls.vim
+++ b/autoload/fzf/decls.vim
@@ -25,37 +25,6 @@ function! s:color(str, group) abort
   return printf("\x1b[%sm%s\x1b[m", color, a:str)
 endfunction
 
-function! s:sink(str) abort
-  if len(a:str) < 2
-    return
-  endif
-  try
-    " we jump to the file directory so we can get the fullpath via fnamemodify
-    " below
-    let l:dir = go#util#Chdir(s:current_dir)
-
-    let vals = matchlist(a:str[1], '|\(.\{-}\):\(\d\+\):\(\d\+\)\s*\(.*\)|')
-
-    " i.e: main.go
-    let filename =  vals[1]
-    let line =  vals[2]
-    let col =  vals[3]
-
-    " i.e: /Users/fatih/vim-go/main.go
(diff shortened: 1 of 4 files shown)


bhcleek:

I'd like to find a more general solution and tie this to the `g:go_list_type` and `g:go_list_type_commands` instead of to the mode; the mode is for what's used to get the information, while `g:go_list_type` and `g:go_list_type_commands` are for how the results are displayed.

kerma:
I'm not sure whether my initial approach could work at all. Calling `fzf#run()` via `jobstart` creates the focus issue and I couldn't find an easy maintainable solution to get the focus back. It would probably need changes in `fzf#run()` to add that support.

Due to that general solution via `g:go_list_type` is probably also quite big change as I don't think you can add the fzf wrap for `go#lsp` methods at all. 

I rewrote the code to use gopls via [command line](https://github.com/golang/tools/blob/master/gopls/doc/command-line.md) instead. It's now very similar to GoDecls with `motion`.

bhcleek:
You're right; your initial approach won't work with the asynchronous nature it had. 

Changing `:GoImplements` to be a synchronous call is going to have performance problems in large repositories; if we're going to make this work, then the call needs to remain asynchronous.

I believe it is possible to change this to work more generally. Thank you for your contribution. I'm not interested in doing a one-off for `:GoImplements` to support fzf, though.

PR #3322, opened 2021-11-09
syntax: add highlight for Go templates within YAML

fioncat:
This can enable highlight for Go templates within YAML. Like `helm template` files.

The syntax won't be automatically enabled, we need to enable it by `set syntax=goyamltmpl`.

Screenshot:

ss


bhcleek:

You can do this in your own vim configuration by putting this file in one of the `after/syntax` directories (e.g. `$HOME/.vim/after/syntax.vim`).

PR #2713, opened 2020-02-10
Set shell to cmd.exe with flag /C on Windows

drichardson:
If shell was set to pwsh, then vim-go would fail when creating a new file from a template.

This PR sets the shell to `cmd.exe` with the `/C` flag, which is how stock vim on Windows is configured.

Fixes #2712.

PR #3323, opened 2021-11-09
Set shellxquote properly for cmd.exe

nonsleepr:
I have custom `shellxquote` in my config and it breaks the plugin. Setting it to quote and restoring after the fact fixes the issue.

### DIFF
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -194,8 +194,8 @@ function! s:system(cmd, ...) abort
     if executable($COMSPEC)
       let &shell = $COMSPEC
       set shellcmdflag=/C
-      set shellquote=
-      set shellxquote=\"
+      set shellquote&
+      set shellxquote&
     endif
   endif
 
--- a/autoload/go/util.vim
+++ b/autoload/go/util.vim
@@ -183,6 +183,8 @@ function! s:system(cmd, ...) abort
   let l:shell = &shell
   let l:shellredir = &shellredir
   let l:shellcmdflag = &shellcmdflag
+  let l:shellquote = &shellquote
+  let l:shellxquote = &shellxquote
 
   if !go#util#IsWin() && executable('/bin/sh')
       set shell=/bin/sh shellredir=>%s\ 2>&1 shellcmdflag=-c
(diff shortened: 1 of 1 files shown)


nonsleepr:
In my case I have the value of `shellxquote` changed according to [this recommendations](https://github.com/neovim/neovim/blob/4174244d8b6718c703f8b8d3e907809bde9430bd/runtime/doc/options.txt#L5285) and that broke `vim-go`.
If I understand correctly, that custom handling is to fix #2713, it looks like the fix for just didn't set all the settings for `cmd.exe`.
An alternative is to remove the change made in #2713 and somehow make `shell=pwsh` work, but it seems for that `shellescape` in Vim/Neovim has to be changed. I might be wrong.

bhcleek:
Ah, I see. In that case, should your changes be modified from 
```
      set shellquote=
      set shellxquote=\"
```
to
```
      unset shellquote
      unset shellxquote
```

so that Vim will do the right thing, relying solely on `shell`?

nonsleepr:
You're right. That's a better idea.

Tested with the following block to reset everything `shell-powershell` recommends to change and it works alright:
```vim
  if go#util#IsWin()
    set shell&
    set shellredir&
    set shellcmdflag&
    set shellpipe&
    set shellquote&
    set shellxquote&
  endif
```

Would it make sense to do it everywhere, not only on Windows and just rely on defaults to properly escape arguments?

bhcleek:
I'm a little hesitant to change more than what we know is needed here, because the blast radius covers more systems than I can test on. For now, let's focus only on the `shellquote` and `shellxquote` needs that you've identified.

nonsleepr:
Makes sense.

Do you think we can add a test global variable which makes the plugin use all defaults?

bhcleek:
I'm not sure I understand what you mean. Can you clarify?

nonsleepr:
I mean, have a global variable, like `g:go_use_shell_defaults` which, when enabled, sets the following and calls `system` instead of performing custom handling which is done by `s:exec`:

```vim
set shell&
set shellredir&
set shellcmdflag&
set shellpipe&
set shellquote&
set shellxquote&
```

bhcleek:
Thanks for clarifying. I don't think we need to introduce a new option yet.

I'll merge these changes soon.

bhcleek:
Sorry for the long delay; I'd forgotten about this.

PR #3348, opened 2022-01-28
Fix play url and add additional curl header

shaunduncan:
Hi!

I was having some trouble getting `:GoPlay` to actuall work correctly - the share would "work" but the resulting url was a 404. I discovered two small issues after looking at the play site's source code that I have included in this patch:

1. The `go.dev` url needed an additional `/p/` item in the url path, otherwise it will 404
2. The curl needed to explicitly specify the content type as `text/plain`, otherwise the resulting play share is just url encoded text.

I also noticed that the share url was not using https, so I modified that as well.

Thanks!

### DIFF
--- a/autoload/go/play.vim
+++ b/autoload/go/play.vim
@@ -26,7 +26,7 @@ function! go#play#Share(count, line1, line2) abort
     return
   endif
 
-  let url = printf("http://go.dev/play/p/%s", snippet_id)
+  let url = printf("https://go.dev/play/p/%s", snippet_id)
 
   " copy to clipboard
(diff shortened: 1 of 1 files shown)


bhcleek:
Thank you for contributing. I'll merge this as soon as tests pass.

I suspect the play site changed again; this was definitely working shortly after play.golang.org transitioned to go.dev/play.

telemachus:
> I suspect the play site changed again; this was definitely working shortly after play.golang.org transitioned to go.dev/play.

For what it's worth, you must be right. I was using it without trouble just last week. (Not a big deal, but I figured it'd be nice to know you're not imagining things!)

PR #3356, opened 2022-02-15
Set `b:main_syntax`, not `g:`

rsrchboy:
The commit history doesn't indicate if this was intentional or not, but
I can't think of a reason why we'd want to set a _global_ variable
talking about a _buffer's_ main syntax.

### DIFF
--- a/syntax/gohtmltmpl.vim
+++ b/syntax/gohtmltmpl.vim
@@ -2,8 +2,8 @@ if exists("b:current_syntax")
   finish
 endif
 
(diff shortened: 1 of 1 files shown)


bhcleek:

bhcleek:
I've reverted this, because `main_syntax` has legitimate uses and it is used appropriately here. See the Vim-provided syntax file for html.

You can read more about it at https://stackoverflow.com/a/16171975.

Issue #3300, opened 2021-10-05
GoRun errors out with 'no packages loaded from `/path/filename.go`


mfrw:
### What did you do? (required: The issue will be **closed** when not provided)
I ran `:GoRun` in a test hello world with main.

```go
package main

import "fmt"

func main() {
        fmt.Println("vim-go")
}
```
I tried to run `:GoRun` on this program.

### What did you expect to happen?

Print `vim-go`

### What happened instead?
```vim
:!'go' 'run' '/home/mfrw/mariner-org/dashboard/cmd/test/main.go\\n'  2>&1| tee /tmp/nvim7Egq2g/6                                                                                                                                                                     
go run: no packages loaded from /home/mfrw/mariner-org/dashboard/cmd/test/main.go\n
```

####  `vimrc` you used to reproduce:

<details><summary>vimrc</summary>

```vim
call plug#begin('~/.vim/bundle')
Plug 'fatih/vim-go', { 'do': ':GoUpdateBinaries' }
call plug#end()

au FileType go nmap <leader>r <Plug>(go-run)
```
</details>

#### Vim version (first three lines from `:version`):

NVIM v0.6.0-dev
#### Go version (`go version`):

go version go1.17.1 linux/amd64
#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/mfrw/.cache/go-build"
GOENV="/home/mfrw/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/mfrw/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/mfrw/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/lib/go-1.17"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go-1.17/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.17.1"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/mfrw/mariner-org/dashboard/go.mod"
CGO_CFLAGS="-g -O2"
[...]

PR #3358, opened 2022-02-16
Use literal newline in order to properly split the string

BodeSpezial:
This should fix `go#tool#Files()` and `go#tool#Imports()` as described by my comment in #3300.

### DIFF
--- a/autoload/go/tool.vim
+++ b/autoload/go/tool.vim
@@ -36,7 +36,7 @@ function! go#tool#Files(...) abort
     if go#util#IsWin()
       let combined .= '{{range $f := .' . sf . '}}{{$.Dir}}\{{$f}}{{printf \"\n\"}}{{end}}{{range $f := .CgoFiles}}{{$.Dir}}\{{$f}}{{printf \"\n\"}}{{end}}'
     else
-      let combined .= "{{range $f := ." . sf . "}}{{$.Dir}}/{{$f}}{{printf \"\\n\"}}{{end}}{{range $f := .CgoFiles}}{{$.Dir}}/{{$f}}{{printf \"\\n\"}}{{end}}"
+      let combined .= "{{range $f := ." . sf . "}}{{$.Dir}}/{{$f}}\n{{end}}{{range $f := .CgoFiles}}{{$.Dir}}/{{$f}}\n{{end}}"
     endif
   endfor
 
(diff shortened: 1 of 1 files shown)


bhcleek:
I'd also be interested in knowing what you see for these `go list` commands on master when `g:go_debug` is set to `['shell-commands']` (e.g. `let g:go_debug=['shell-commands']` in your vimrc).

bhcleek:
FWIW, I _can_ duplicate the issue with Neovim, but not with Vim. It seems like the issue here may be how Neovim is handling escaped characters when it invokes external commands.

BodeSpezial:
Ah that's interesting.
I thought this was a Windows/Linux thing. I actually didn't try it with Vim.

I'll try it with Vim, but if you can only reproduce it with Neovim it is probably a nvim issue.

I will also look into `go list`.

Yes, it seems as if it has to do with the escaping in neovim.

BodeSpezial:
Ok, so.

I looked into it and I get the same error with vim as well (I am on master. The complete config file I load for vim as well as nvim is
```vim
let data_dir = '~/.vim'

if empty(glob(data_dir . '/autoload/plug.vim'))
  silent execute '!curl -fLo '.data_dir.'/autoload/plug.vim --create-dirs  https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
  autocmd VimEnter * PlugInstall --sync | source $MYVIMRC
endif

call plug#begin(data_dir . '/plugged')
Plug 'fatih/vim-go', { 'for': 'go', 'do': ':GoUpdateBinaries' }
call plug#end()

let g:go_debug=['shell-commands']
```

I dug around quite some time and thought I could pin it down but I couldn't settle on a definitive result.

Some findings:
- when using `split(l:out, '\\n')` the newline character is properly cut off
- when running the command logged by setting the go_debug variable:
`go list -f '{{range $f := .GoFiles}}{{$.Dir}}/{{$f}}{{end}}{{printf "\\n"}}{{range $f := .CgoFiles}}{{$.Dir}}/{{$f}}{{printf "\\n"}}{{end}}'`
 manually in a shell I get the same result, but when using `printf "\n"` (in sh) instead it works fine

---

But regarding your concerns it could break compatibility on other OS: My fix uses basically the same code as in `go#util#Deps` which works just fine. Given that this works I wouldn't expect any problems in the other functions.

Unfortunately I can't test it on any non-Linux OS right now to see how it behaves there.

Issue #3110, opened 2020-12-10
Does my :GoDecls display correctly?


hungrybirder:
Hello, @bhcleek 
I got a problem when I use :GoDecls command with fzf.vim.
On the right side, it just shows the function name, no function body. Is that right?

![image](https://user-images.githubusercontent.com/1021047/101743851-c6bbbe00-3b05-11eb-8b19-8ea7a9352a60.png)

`main.go`
```
package main

import "fmt"

// Hello just hello
func Hello() {
	fmt.Println("hello world")
}

func main() {
	fmt.Println("main done")
}
```

#### vimrc
```
Plug 'fatih/vim-go', { 'do': ':GoUpdateBinaries' }
Plug 'junegunn/fzf', { 'dir': '~/.fzf', 'do': './install --all' }
Plug 'junegunn/fzf.vim'
Plug 'stsewd/fzf-checkout.vim'
let g:fzf_layout = { 'window': { 'width': 0.8, 'height': 0.8 } }
nnoremap <C-p> :GFiles<CR>
let g:fzf_action = {
  \ 'ctrl-x': 'split',
  \ 'ctrl-]': 'vsplit' }

let g:go_test_prepend_name = 1
let g:go_fmt_fail_silently = 1
let g:go_fmt_command = "goimports"
let g:go_def_mode = "gopls"
" let g:go_implements_mode = 'gopls'
let g:go_list_type = "quickfix"
let g:go_auto_type_info = 1
let g:go_fmt_autosave = 1
let g:go_auto_sameids = 0
let g:go_doc_popup_window = 1
let g:go_null_module_warning = 0
let g:go_echo_command_info = 1
let g:go_autodetect_gopath = 1
let g:go_metalinter_autosave_enabled = ['vet', 'golint']
let g:go_metalinter_enabled = ['vet', 'golint']
let g:go_highlight_space_tab_error = 0
let g:go_highlight_array_whitespace_error = 0
let g:go_highlight_trailing_whitespace_error = 0
let g:go_highlight_extra_types = 0
let g:go_highlight_build_constraints = 1
let g:go_highlight_types = 0
let g:go_highlight_operators = 1
let g:go_highlight_format_strings = 0
let g:go_highlight_function_calls = 0
let g:go_gocode_propose_source = 1
let g:go_addtags_transform = 'camelcase'
let g:go_fold_enable = []
let g:go_debug_log_output = 'debugger'
" set vars/stack/out/goroutines windows
[...]

PR #3362, opened 2022-02-19
Show the `pos` (file info) column for GoDecls FZF view

jonaustin:
For some reason the file info column is not displayed in the FZF GoDecls view.

This PR simply adds that column to the view.

The lack of file info can be seen in the screenshot here: https://github.com/fatih/vim-go/issues/3110

The difference is simply e.g.:

```
main                    func
```

versus:

```
main                    func    |main.go:7:1| 
```

### DIFF
--- a/autoload/fzf/decls.vim
+++ b/autoload/fzf/decls.vim
@@ -143,7 +143,7 @@ function! fzf#decls#cmd(...) abort
         \)
   call fzf#run(fzf#wrap('GoDecls', {
         \ 'source': call('<sid>source', a:000),
-        \ 'options': printf('-n 1 --with-nth 1,2 --delimiter=$''\t'' --preview "echo {3}" --preview-window "wrap" --ansi --prompt "GoDecls> " --expect=ctrl-t,ctrl-v,ctrl-x%s', colors),
+        \ 'options': printf('-n 1 --with-nth 1,2,4 --delimiter=$''\t'' --preview "echo {3}" --preview-window "wrap" --ansi --prompt "GoDecls> " --expect=ctrl-t,ctrl-v,ctrl-x%s', colors),
         \ 'sink*': function('s:sink')
         \ }))
 endfunction


bhcleek:
Thank you for contributing.

Why is this needed?

jonaustin:
Because it's useful to know which file a function is from, but maybe I'm missing something?

bhcleek:
I don't see why that would be useful given that directories contain only a single package and function names are within a package a unique. You can navigate to the function by selecting it and you can see the function signature in the preview window.

jonaustin:
That's fair -- I guess I got confused because the ctrlp version of this does include the the file attributes: 
image

Apparently also had it in my head for some reason that GoDecls acted recursively over the directory tree which I see now was incorrect.

Thanks for the quick feedback, feel free to close this issue.

bhcleek:
That's an interesting point; it may make sense for the ctrlp and fzf integrations to be more consistent in their presentation.

Honestly, I'd lean more towards removing the file info from the ctrlp integration, because `:GoDecls` only operates on the current file and `:GoDeclsDir` only operates on a single directory and the file info takes up valuable screen real estate.

I'll give it some thought, but I'm interested to hear your perspective.

jonaustin:
That makes a lot of sense; happy to refactor this PR to instead remove it from the ctrlp autoload.

bravoecho:
Just a quick comment in support of providing additional context to the symbol names. 

I often have to give up `:GoDeclsDir` when there are different implementations of the same symbol, for example different types implementing the same interface cannot be told apart in the results, their methods are all, for example `String() string`, and I don't know which one to select.

Another common situation is when the same symbol is present both in the actual package (say, `mypkg`), and in the `..._test` package (`mypkg_test`). The `..._test` package coexists in the same directory as the real one, and symbols with the same name across the two cannot be identified in the `GoDeclsDir` results. This is often the case when using generate mocks.

bhcleek:
@bravoecho in those cases, the preview window that allows me to distinguish between the types that are implementing the methods. Are you not seeing the preview window on the right hand side?

bravoecho:
Thank you @bhcleek for your message 🙂 Good point! I actually disabled all previews to avoid information overload. You make me realise that perhaps this is because I'm using the feature for navigating code, rather than for searching it.

PR #3375, opened 2022-03-19
syntax: add syntax for go.work files

### LINKED ISSUE #3373: go.work syntax
Now go1.18 is out with workspace support via `go.work`. The syntax for it should be defined. Similarly there is the `go.work.sum` which is the same as `go.sum`?

bhcleek:
Closes #3373


jackieli-tes (approved):
works for me. Thanks!

PR #3376, opened 2022-03-19
update test to use Go 1.18

bhcleek:


bhcleek:
The only failing tests here are due to staticcheck, which doesn't yet provide a release that works with Go 1.18.

PR #3379, opened 2022-03-20
Code to always display the current goroutine first in the list of gor…

kkrime:
Simple change to make the current go routine running always be displayed first on the goroutine output at the bottom of the debug screen:

Screen Shot 2022-03-20 at 12 51 19


bhcleek:
Thank you

kkrime:
Some of the tests are failing, from looking at the logs, the reasons looks like they're related to the testing environment and not from my code. Can someone look into this?

PR #3406, opened 2022-04-15
remove unneeded tt := tt in go snippets

alovak:
I don't think that in table test snipped `tt := tt` is needed:

```
for _, tt := range tests {
	tt := tt
```

am I missing something?


bhcleek:
Capturing the range variable with `tt := tt` allows the correct value to be used if/when the user adds `t.Parallel()` in the subtest or otherwise leverages concurrency in the subtest, and capturing it this way be default has no meaningful penalty.

bhcleek:
Thank for contributing 🙇

alovak:
@bhcleek thanks for answering! that absolutely makes sense

PR #3437, opened 2022-07-28
Delve options

rikard-acst:
* Allowing delve to use the current working directory
Currently, I am working on a code base where executables are defined inside of a `/cmd/{exename}/` folder and share code, including a configuration file from the parent folder.  This can be accomplished with VSCode in the `launch.json` file using the property `cwd`.

* max string length, max array, and max recursion
Reading variables, especially strings, with a max length of 20 is limiting. I decided not to change the defaults here, but allow the end user to expand them.

### DIFF
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -785,6 +785,9 @@ function! go#debug#Start(mode, ...) abort
       if a:mode is 'debug' || a:mode is 'test'
         let l:cmd = extend(l:cmd, s:package(a:000))
         let l:cmd = extend(l:cmd, ['--output', tempname()])
+        if get(g:, "go_debug_use_getcwd", "false") ==? "true"
+          let l:cmd = extend(l:cmd, ['--wd', getcwd()])
+        endif
       elseif a:mode is 'attach'
         let l:cmd = add(l:cmd, a:1)
         let s:state['kill_on_detach'] = v:false
@@ -1166,7 +1169,7 @@ function! s:update_variables() abort
   " MaxStructFields is the maximum number of fields read from a struct, -1 will read all fields.
   let l:cfg = {
         \ 'scope': {'GoroutineID': s:goroutineID()},
-        \ 'cfg':   {'MaxStringLen': 20, 'MaxArrayValues': 20, 'MaxVariableRecurse': 10}
+        \ 'cfg':   {'MaxStringLen': get(g:, "go_debug_maxstringlen", 20), 'MaxArrayValues': get(g:, "go_debug_maxarrayvalues", 20), 'MaxVariableRecurse': get(g:, "go_debug_maxvariablerecur", 10)}
         \ }
 
   try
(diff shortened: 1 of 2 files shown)


bhcleek (commented):
Thank you for contributing a well-crafted and simple PR with documentation; it's a rare treat.

I'll need to think about some of this, but other things should be changed for sure. First, I'd rather not introduce specific support or delve's `--wd` flag. Vim isn't VSCode and has a very different model; vim-go strives to be consistent with Vim's practices and patterns. One of my concerns with adding support for delve's `--wd` is that it opens the flood gates for users to want similar functionality for the other tools that vim-go relies on. 🤔 

An alternative that's already available to you for debugging in the directory you prefer is to open one of the go files there and execute `:GoDebugStart` and give it the import path of the package you want to run; that will ensure the working directory is what you're after.

If we were going to offer something to set the working directory, I'd want to consider another method to set it explicitly instead of relying only on the current working directory, but I'm not convinced we should offer something like that at all. Give me a few days to consider it.

The variables window isn't intended to show everything. `:GoDebugPrint` is a good alternative to see more complete information. Can you give me some more information about the actual problem you're encountering that motivated the change to set the cfg options?

autoload/go/debug.vim
```diff
@@ -785,6 +785,9 @@ function! go#debug#Start(mode, ...) abort
       if a:mode is 'debug' || a:mode is 'test'
         let l:cmd = extend(l:cmd, s:package(a:000))
         let l:cmd = extend(l:cmd, ['--output', tempname()])
+        if get(g:, "go_debug_use_getcwd", "false") ==? "true"
```
bhcleek, line 788: If we were to do something like this, then it'd need to be
```suggestion
if get(g: "go_debug_use_getcwd", 0) is 1
```

rikard-acst:
> I'd rather not introduce specific support or delve's --wd flag. Vim isn't VSCode and has a very different model ... open one of the go files there and execute :GoDebugStart

It's a bit more work to pass in variables each time I need to debug. That being said, I know that maintaining any open source project requires commitment against scope creep, so I understand.

> :GoDebugPrint is a good alternative to see more complete information. Can you give me some more information about the actual problem you're encountering that motivated the change to set the cfg options?

So, apparently I have worked using vim-go for over a year now and completely missed `GoDebugPrint` 🤦

bhcleek:
> It's a bit more work to pass in variables each time I need to debug. That being said, I know that maintaining any open source project requires commitment against scope creep, so I understand.

The working directory that vim-go uses for delve will be the directory that contains the active file when a user executes `:GoDebugStart`. So let's say you have a file directory structure like this:
```
github.com/rikard-acst
    cmd/
        tool1/
            main.go
        tool2/
            main.go
    internal/
        config/
            config.go
```

You can debug the  command `github.com/rikard-acst/cmd/tool1` with the working directory set to `github.com/rikard-acst/internal/config` by editing `github.com/rikard-acst/internal/config/config.go` and running `:GoDebugStart github.com/rikard-acst/cmd/tool1`

PR #3438, opened 2022-08-01
Fix typo in manual

erodrigufer:
It is a very minimal fix of a typo in the manual. The word 'templates' was incorrectly spelled.


bhcleek:
Thank you

PR #3456, opened 2022-09-26
Adding more packages to "extra types"

hbourgeot:
# Added to extra types the following:

## io package now has more highlighting for some functions, and types, like below:
- Copy / CopyBuffer
- ByteReader / ByteScanner /ByteWriter

### bufio

- Reader / Scanner / Writer

### os

- Getenv
- Stdin / Stdout / Stderr

### strconv

- Atoi
- ParseFloat

### strings

- Contains
- Repeat
- Replace / ReplaceAll
- Split
- ToLower / ToUpper
- Trim

### sql

- DB
- Row/Rows
- Stmt

### fmt

- Scan / Scanf / Scanln
- Sprintf
- Errorf
- Fprint / Fprintf
- Print / Printf / Println

### http

- Handle / HandleFunc
- FileServer
- Request
- ResponseWriter
- Error
- ListenAndServe
- NotFound
- Redirect
- Serve
- StatusText
- Handler
- Header
- Servemux
- Server
- Get / Post

### template

Template

### log

- Fatal / Fatalf / Fatalln
- Print / Printf / Println
- Logger

### json

- Marshal / Unmarshal
- Decoder / Encoder

### time

- Sleep
- Duration
- Location
- Time

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -264,7 +264,7 @@ if go#config#HighlightExtraTypes()
   syn match goExtraType /\<bufio\.\%(Reader\|Scanner\|Writer\)\>/
   syn match goExtraType /\<os\.\%(Getenv\|Stdin\|Stdout\|Stderr\)\>/
   syn match goExtraType /\<strconv\.\%(Atoi\|ParseFloat\)\>/
-  syn match goExtraType /\<strings\.\%(Compare\|Contains\|Repeat\|Replace\|ReplaceAll\|Split\|ToLower\|ToUpper\|Trim\|TrimSpace\|)\>/
+  syn match goExtraType /\<strings\.\%(Contains\|Repeat\|Replace\|ReplaceAll\|Split\|ToLower\|ToUpper\|Trim\)\>/
   syn match goExtraType /\<sql\.\%(DB\|Row\|Rows\|Stmt\)\>/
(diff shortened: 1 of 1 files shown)


bhcleek (commented):
Thank you for contributing.

The difficult thing with these changes is their value is a bit subjective in the absence of clear data about which standard library identifiers are most often used, the lack of clear guidelines on vim-go's part about how to decide what to include in the extras, and every regex we add to the syntax slows down parsing.

Some of these seem useful to me, but I suspect others wouldn't be in the most common standard library identifiers (e.g. `template.Template`). Are you open to removing some of these?

I wonder if there's been any information available that's been obtained via static analysis on public Go code regarding which standard library identifiers are most commonly referenced?

hbourgeot:
> Some of these seem useful to me, but I suspect others wouldn't be in the most common standard library identifiers (e.g. `template.Template`). Are you open to removing some of these?
> 
> I wonder if there's been any information available that's been obtained via static analysis on public Go code regarding which standard library identifiers are most commonly referenced?

Yes, I'm very open to removing some of these, even, I'm open to delete the half of these. About the static analysis on which standart libray identifiers are most commonly referenced, I don't get any information about that.

I can delete the following packages, from my personal opinion, I don't use them much.

- io: CopyBuffer
- os: Stdin / Stdout / Stderr
- strings: Contains / Repeat / Replace / ReplaceAll
- sql: DB / Row / Rows / Stmt
- fmt: Scan / Scanf / Scanln
- http: Request / ResponseWriter / Serve / StatusText / Handler / Server
- log: Fatal / Fatal  / Print / Logger
- json: Marshal / Unmarshal
- time: all

Issue #2911, opened 2020-06-08
(go-def-vertical/split) doesn't open a split when jumping to a different place in same file [NeoVim]


ahouseago:
### What did you do? (required: The issue will be **closed** when not provided)
In NeoVim, I used the `(go-def-vertical)` mapping to jump to a definition of a function defined in the same file in a vertical split:
```vim
autocmd FileType go nnoremap <buffer> <localleader>dv :<C-U>call go#def#Jump("vsplit", 0)
```

The same as `<Plug>(go-def-vertical)` except without immediately invoking it, waiting for me to hit `<CR>` before jumping.

### What did you expect to happen?
The file would be opened again in a new vertical split and it would jump to the function definition, like using `:vsplit` then `:GoDef`.

### What happened instead?
A split was not opened, and instead it jumped to the definition in the current buffer.

#### vim-go version:
Latest (not sure how to check, but used `:PlugUpdate` before submitting this.
####  `vimrc` you used to reproduce:

<details><summary>vimrc</summary><br><pre>
let g:go_def_mapping_enabled = 0
augroup GoStuff
  autocmd!
  autocmd FileType go nmap <buffer> <localleader>h <Plug>(go-describe)
  autocmd FileType go nnoremap <buffer> <localleader>dd :<C-U>call go#def#Jump('', 0)
  autocmd FileType go nnoremap <buffer> <localleader>dx :<C-U>call go#def#Jump("split", 0)
  autocmd FileType go nnoremap <buffer> <localleader>dv :<C-U>call go#def#Jump("vsplit", 0)
  autocmd FileType go nnoremap <buffer> <localleader>dt :<C-U>call go#def#Jump("tab", 0)
augroup end
call plug#begin('~/.local/share/nvim/plugged')
  Plug 'fatih/vim-go', { 'do': ':GoInstallBinaries' }
call plug#end()
let g:go_fmt_command = "goimports"
let g:go_auto_sameids = 1
let g:go_updatetime = 100
</pre></details>

#### Vim version (first three lines from `:version`):

NVIM v0.4.3
Build type: Release
LuaJIT 2.0.5

#### Go version (`go version`):

go version go1.14.3 darwin/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/ahouseago/Library/Caches/go-build"
[...]

PR #3459, opened 2022-10-27
Optionally split to definition in the same buffer

ryankinderman:
My typical workflow in vim when traversing a codebase is to start at a function of interest, then `<C-w>]` my way through various call stacks from there, so I have a full view. As a result of this workflow, splitting when jumping to a definition is an important feature for me. Currently, if a definition is in the same file as the current buffer, the feature to jump to the definition does not open a split when using a command like `<C-w>]`, which typically creates a split.

These changes add a `g:go_def_split_same_buffer` option that, when enabled, will cause jumping to a definition with a split using e.g. `<C-w>]` to always create a split, even if the definition is in the same file that is currently open.

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -188,6 +188,10 @@ function! go#config#DefReuseBuffer() abort
   return get(g:, 'go_def_reuse_buffer', 0)
 endfunction
 
+function! go#config#DefSplitSameBuffer() abort
+  return get(g:, 'go_def_split_same_buffer', 0)
+endfunction
+
 function! go#config#DefMode() abort
   return get(g:, 'go_def_mode', 'gopls')
 endfunction
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -167,7 +167,7 @@ function! go#def#jump_to_declaration(out, mode, bin_name) abort
   let old_switchbuf = &switchbuf
 
   normal! m'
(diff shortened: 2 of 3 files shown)


bhcleek:
Thank for you contributing.

We don't need a new option for this. You've identified a bug in how vim-go handles jumping to a definition using the split variants. On the one hand, we have clear Vim documentation that says that `ctrl-w+]` should split the window. On the other hand vim-go docs are clear that `g:go_def_reuse_buffer` will cause the `:GoDef` related variants that cause splits to reuse an existing buffer. 

I've created #3462 to address the bug without introducing a new option. Do you want to give it a try before I merge it?

ryankinderman:
Oh! Yea, I came across [issue 2911](https://github.com/fatih/vim-go/issues/2911) in my initial search for a solution, which made it seem like not splitting on the same file was the desired default behavior. I'm new to the plugin and was a bit confused about the purpose of `g:go_def_reuse_buffer` in light of that, but figured maybe I was missing some context.

I tested out #3462 and it seems to work as expected, thanks so much for the fix.

I've been getting a lot of use out of the plugin, and really appreciate it!

bhcleek:
Thank you for researching that before you submitted a PR. In this case, you use case of relying on `<C-w>]` prompted me to take a look at Vim feature for splitting and jumping. Given that use case, it's clear to me now that the split mappings should function similarly, too.

PR #3461, opened 2022-10-28
fix go work syntax having wrong highlighting entirely

ArtAndreev:
`go.work` file has wrong syntax highlighting:

image

The reason is broken regexp that colors all file wrong:
```vim
syntax match goworkPath "\f\+"
```

I suggest to drop this regexp, because path can be treated like module name without version, and provide the same highligthing for `use` block as for `require` block in `go.mod`:

image

Also I did some housekeeping in file.

PR with the same changes is opened in `vim-polyglot`: https://github.com/sheerun/vim-polyglot/pull/819.

### DIFF
--- a/syntax/gowork.vim
+++ b/syntax/gowork.vim
@@ -12,13 +12,11 @@ syntax case match
 
 " match keywords
 syntax keyword goworkGo      go      contained
-syntax keyword goworkUse use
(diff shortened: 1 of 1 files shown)


bhcleek:
Thank you

PR #3472, opened 2022-12-10
Functionality to run single test case in test func

jrahme-cci:
# What Changed
This adds the ability to run a single case from a table of tests in a similar manner to the Go plugin for IntelliJ (GoLand). Using the line under the cursor, the test case name is extracted from a field initialization and appended to the test function name. Otherwise the test is ran in the same manner as `go#test#Func()`

For example; with a test function like:

```
func FooBar_Test (t *testing.T){
    testCases := []struct {
        name string,
        want int,
        have int,
    }{
        {
            name: "first",
            want: 1,
            have: 1,
        },
        {
            name: "second",
            have: 2,
            want: 3,
        },
    }
    for _, tc := range testCases {
        t.Run(tc.Name, func(t2 *testing.T){
            assert.Equal(t2, tc.have, tc.want)
        }
    }
}
```

If the cursor is on line 8 and `:GoTestCase` is called; the test `./FooBar_Test/first` would be passed as the package parameter to `go test`. 

# How it was tested

Manual testing was done to verify the value initialization was being correctly parsed. I did not write automated tests for this functionality as the functionality is very similar to `go#test#Func` and saw limited value in adding additional automated tests for the regex being used to extract the value. 

# Other considerations

Initially I had attempted to be clever and identify the variable used by `t.Run` to determine which symbol to look for in the testcase, and the LSP `textDocument/foldingRanges` used to determine the scope of the test case surrounding the cursor. This lead to more complexity than I thought necessary just to execute a test case, especially when vim already provides tooling to accurately move the cursor to specific lines or patterns. 

I also attempted using the `textDocument/documentSymbol` LSP functionality too. But determined that the additional complexity for the implementation was also a bit extra for this feature, and increased the error footprint.

### DIFF
--- a/autoload/go/test.vim
+++ b/autoload/go/test.vim
@@ -127,16 +127,24 @@ endfunction
 " Arguments are passed to the `go test` command.
 function! go#test#Case(bang, ...) abort
   let l:testParent = go#util#TestName()
+  if l:testParent == ""
+    call go#util#EchoWarning("[test] no test found immediate to cursor")
+    return
+  endif
 
   " Check for a struct initialization on the line, and take its value
-  let l:searchPattern = '\s*\:\s*"\zs\w*\ze'
-  let l:subTestName = matchstr(getline("."), l:searchPattern)
+  " e.g. `name: \"test name\",` would extract 'testName'
+  let l:subTestName = go#test#subtestUnderCursor(getline("."))
   if l:subTestName == ""
     call go#util#EchoError('[test] no name for test case found on cursor line')
     return
   endif
 
-  let l:args = [a:bang, 0, "-run", l:testParent . "/" . l:subTestName . "$"]
+  " the pattern after l:testParent will allow testcases that are nested
+  " arbitrarily deep to be selected to run
+  let l:testParentPattern = l:testParent . "/\([^/]\+/\)*"
+
+  let l:args = [a:bang, 0, "-run", l:testParentPattern . "/" . l:subTestName . "$"]
   if a:0
     call extend(l:args, a:000)
   else
@@ -152,6 +160,13 @@ function! s:test_job(cmd, args) abort
   call go#cmd#autowrite()
 
   call go#job#Spawn(a:cmd, a:args)
+endfunction
+
+function! go#test#subtestUnderCursor(cursorLine) abort
+  " e.g. `name: \"test name\",` would extract 'test name'
+  " e.g. `name: \"test_name\"` would extract 'test_name'
+  let l:structInitPattern =  '\:\s*\"\zs\(\w\|\s\)*\ze\"'
+  return matchstr(a:cursorLine, l:structInitPattern)
 endfunction
 
 let s:efm = ""
--- a/autoload/go/test.vim
+++ b/autoload/go/test.vim
@@ -122,6 +122,31 @@ function! go#test#Func(bang, ...) abort
   call call('go#test#Test', args)
 endfunction
 
+" Testcase runs a single subtest of a function for a subtest name on the
+" current cursor line.
+" Arguments are passed to the `go test` command.
+function! go#test#Case(bang, ...) abort
+  let l:testParent = go#util#TestName()
+
+  " Check for a struct initialization on the line, and take its value
+  let l:searchPattern = '\s*\:\s*"\zs\w*\ze'
+  let l:subTestName = matchstr(getline("."), l:searchPattern)
+  if l:subTestName == ""
+    call go#util#EchoError('[test] no name for test case found on cursor line')
+    return
+  endif
+
+  let l:args = [a:bang, 0, "-run", l:testParent . "/" . l:subTestName . "$"]
+  if a:0
+    call extend(l:args, a:000)
+  else
+    let l:timeout = go#config#TestTimeout()
+    call add(l:args, printf("-timeout=%s", l:timeout))
+  endif
+
+  call call('go#test#Test', l:args)
+endfunction
+
 function! s:test_job(cmd, args) abort
   " autowrite is not enabled for jobs
   call go#cmd#autowrite()
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -423,6 +423,19 @@ CTRL-t
 
     If [!] is not given the first error is jumped to.
 
+
+:GoTestCase[!] [expand]                                          *:GoTestCase*
+
+    Runs :GoTest, but only for a single test case within a single test
+    function immediate to your cursor using 'go test's '-run' flag.
+
+    Lookup of the parent test function is as in `:GoTestFunc`. The test case
+    name lookup is done by taking the cursor line, checking for a struct
+    varliabe initialization, and taking the value of the name. The test
+    is then identified in the format of `TestFunctionName/TestCaseName$` .
+
+    if [!] is not given the first error is jumped to.
+
                                                               *:GoTestCompile*
 :GoTestCompile[!] [expand]
 
--- a/doc/vim-go.txt
+++ b/doc/vim-go.txt
@@ -431,7 +431,7 @@ CTRL-t
 
     Lookup of the parent test function is as in `:GoTestFunc`. The test case
     name lookup is done by taking the cursor line, checking for a struct
-    varliabe initialization, and taking the value of the name. The test
+    variable initialization, and taking the value of the name. The test
     is then identified in the format of `TestFunctionName/TestCaseName$` .
 
     if [!] is not given the first error is jumped to.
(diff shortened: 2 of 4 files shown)


doc/vim-go.txt
```diff
@@ -423,6 +423,19 @@ CTRL-t
...
 
+
+:GoTestCase[!] [expand]                                          *:GoTestCase*
+
+    Runs :GoTest, but only for a single test case within a single test
+    function immediate to your cursor using 'go test's '-run' flag.
+
+    Lookup of the parent test function is as in `:GoTestFunc`. The test case
+    name lookup is done by taking the cursor line, checking for a struct
+    varliabe initialization, and taking the value of the name. The test
```
bhcleek, line 434: `varliabe` -> `variable`

autoload/go/test.vim
```diff
@@ -122,6 +122,31 @@ function! go#test#Func(bang, ...) abort
   call call('go#test#Test', args)
 endfunction
 
+" Testcase runs a single subtest of a function for a subtest name on the
+" current cursor line.
+" Arguments are passed to the `go test` command.
+function! go#test#Case(bang, ...) abort
+  let l:testParent = go#util#TestName()
```
bhcleek, line 129: Please handle the case when `l:testParent` is the empty string.

jrahme-cci:
> How often do you find that you really need this kind of functionality when `:GoTestFunc` is insufficient?

Testing a single case in a large set of them is pretty common for me. Especially when looking to test new cases when addressing bugs. It's been about the only pain point for me switching back to vim from Intelli-J for my normal workflow. I'll get the change requests wrapped up tomorrow in my AM.

autoload/go/test.vim
```diff
@@ -122,6 +122,31 @@ function! go#test#Func(bang, ...) abort
...
+
+  " Check for a struct initialization on the line, and take its value
+  let l:searchPattern = '\s*\:\s*"\zs\w*\ze'
```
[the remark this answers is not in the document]
jrahme-cci: Good call about the regex there, you're right it is a little sloppy. I don't need to use the preceding whitespace for the back anchor, I'll chop that out. I'll also some testing to cover that the pattern works as expected for names with and w/o whitespace. It looks like `go test -run` will take a subtest name and transform the spaces for me. So I'm happy to let it do that instead of more string manipulation here.

autoload/go/test.vim
```diff
@@ -122,6 +122,31 @@ function! go#test#Func(bang, ...) abort
...
+  let l:subTestName = matchstr(getline("."), l:searchPattern)
+  if l:subTestName == ""
+    call go#util#EchoError('[test] no name for test case found on cursor line')
+    return
+  endif
+
+  let l:args = [a:bang, 0, "-run", l:testParent . "/" . l:subTestName . "$"]
```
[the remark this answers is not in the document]
jrahme-cci: You're right that it'd be a corner case, I don't think that patter will make things overly unreadable either so no reason to not add that in for some extra safety.

autoload/go/test.vim
jrahme-cci: Done

autoload/go/test.vim
```diff
@@ -122,13 +122,53 @@ function! go#test#Func(bang, ...) abort
...
+  " e.g. `name: \"test name\",` would extract 'testName'
+  let l:subTestName = go#test#subtestUnderCursor(getline("."))
+  if l:subTestName == ""
+    call go#util#EchoError('[test] no name for test case found on cursor line')
+    return
+  endif
+
+  " the pattern after l:testParent will allow testcases that are nested
+  " arbitrarily deep to be selected to run
+  let l:testParentPattern = l:testParent . "/\([^/]\+/\)*"
```
bhcleek, line 145: No need to escape the `()` and `+` here. `go test` will handle the expression according to the rules at https://github.com/google/re2/wiki/Syntax. 

My apologies for leading you astray on that regexp syntax earlier. I always have to look it up to remember what I do and don't have to escape when using Go's regular expressions, because I spend way more time using Vim's regular expression engine than Go's.

autoload/go/test.vim
```diff
@@ -122,13 +122,53 @@ function! go#test#Func(bang, ...) abort
...
+
+  let l:args = [a:bang, 0, "-run", l:testParentPattern . "/" . l:subTestName . "$"]
```
bhcleek, line 147: The argument to `-run` will have two slashes in it now. I think what's needed is either pass `l:testParentPattern . l:subTestName` as the argument to `-run` or to change the value assigned to `l:testParentPattern` two lines up to `l:testParent . '(/[^/]+)*`

autoload/go/test.vim
```diff
@@ -122,13 +122,53 @@ function! go#test#Func(bang, ...) abort
...
   " autowrite is not enabled for jobs
   call go#cmd#autowrite()
 
   call go#job#Spawn(a:cmd, a:args)
 endfunction
 
+function! go#test#subtestUnderCursor(cursorLine) abort
+  " e.g. `name: \"test name\",` would extract 'test name'
+  " e.g. `name: \"test_name\"` would extract 'test_name'
+  let l:structInitPattern =  '\:\s*\"\zs\(\w\|\s\)*\ze\"'
```
bhcleek, line 168: This is still too narrow. Tests names can easily be outside the set you've defined here. Consider:
* `name: "foo-bar",`
* `name: "México",`
* `name: "⌘",`
* `name: "thing 1, thing 2, and thing 3",`
*  `name: "condition a; condition b",`
* `name: "1/2",`

I don't mean for that list to be exhaustive; it's only illustrative of some examples in the broader scope that this regular expression needs to be constructed to handle.

I also wonder if this needs to handle raw strings in addition to the usual double quoted strings..

bhcleek (changes requested):
Thank you for the changes. We're definitely closer.

I think you may be starting to see why this feature has never been implemented before; the cases it needs to handle are broader than most of us encounter regularly, and using regular expressions requires careful thought about what both false positives and false negatives.

bhcleek:
You may find `g:go_debug=['shell-commands']` useful when you're testing your changes. With `g:go_debug=['shell-commands']` set, vim-go will output the shell command to the the commandline and you can see the output again with `:messages` if you miss it as it's going by.

autoload/go/test.vim
jrahme-cci: No worries! I'd rather iterate here than have someone (me) get hit with a bug that at best messes up a workflow or at worst hides failing tests.

autoload/go/test.vim
bhcleek: There's no need to try to leverage LSP for this.

Try `:\s*\%("\([^"]\+\)"\)` and if that doesn't match anything then try that same, but with backticks instead of quotes.

> calls that are called before the calls for the test cases

What do you mean?

> strings that are transformed from the declaration in the test case to something different when running the test; `t.Run(fmt.Sprintf("Test_%s", testCase.name), func( tt *testing.T) { ...`

Yeah, that's a concern. And is one of the kinds of edges that historically kept this feature from being implemented.

jrahme-cci: > What do you mean?
It's mostly a moot point if you don't want LSP usage here, and likely an extreme corner case. But regarding 
> if > 1 string symbols exist find the next invocation of testing.Run() and check for the value passed to that and search symbols on the cursor line for a string symbol directly after that field symbol.
and identifying the `t.Run(...` that is being looped for the test cases to discover the field used for the name. 
```
func Awesome_Test(t *testing.T) {
    testCases := []struct{
        name string
        want string
     }{
         // unless we key specifically for the field or string under the cursor this is ambiguous for which
         // field should be the test name
         {name: "myTest", want: "myValue"},
         {name: "yourTest", want: "yourValue"},
     }

     // This test doesn't use the test cases but would still show a non string symbol for the value. So checking 
     // for that symbol in the fields of the test cases will at best return nothing, and at worst return the wrong value
     // to test with.
     // There's more complexity if several sets of test cases are also declared in a single test function as well. 
     subTestName := "foo"
     t.Run(subTestName, func(tt *testing.T){
       // test logic
    }

    // this would be the actual testing.Run() call we'd want to check to identify the field used for the test name in
    // in the testing struct. 
    for _, tc := range testCases {
        t.Run(tc.name, func(tt *testing.T){
            // test logic
        }
    }
}
```
But like I noted above, if you don't want LSP usage for this feature that's all moot. 


Otherwise IntelliJ(GoLand) has some [restrictions noted](https://www.jetbrains.com/go/guide/tips/run-test-in-table-test/) for using this feature that would get around issues about case manipulation between the declaration of the test cases and running them. But I'm not sure your appetite for restricting use cases for a feature. 

I think the string formatting problem can be worked around with an inclusive enough pattern in the `go test -run ...` call. But I'm going to tear this PR down and rework a couple of things to see if I can get this to be more flexible.

jrahme-cci:
Closing the PR while I think on some implementation details to make the functionality more flexible.

PR #3477, opened 2022-12-23
Fix import path for josharian/impl 

FooSoft:
Looks like josharian decided to rename the `master` branch to `main`, which unfortunately breaks the vim-go `GoInstallBinaries` / `GoUpdateBinaries` commands for that tool.

Updating the import path.

### DIFF
--- a/plugin/go.vim
+++ b/plugin/go.vim
@@ -54,7 +54,7 @@ let s:packages = {
       \ 'gorename':      ['golang.org/x/tools/cmd/gorename@master'],
       \ 'gotags':        ['github.com/jstemmer/gotags@master'],
       \ 'guru':          ['golang.org/x/tools/cmd/guru@master'],
-      \ 'impl':          ['github.com/josharian/impl@master'],
+      \ 'impl':          ['github.com/josharian/impl@main'],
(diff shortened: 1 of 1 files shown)


bhcleek:
Thank you.

What problem are you actually seeing? I'm still able to install `impl` with `:GoUpdateBinaries impl`.

FooSoft:
@bhcleek I'm not able to pull down `impl` because the `master` branch doesn't exist (and nvim complains about this when running `:GoUpdateBinaries`. I'm not sure why you aren't seeing this too. Running git version 2.38.1.

bhcleek:
Is your `GOPROXY` (i.e. `go env GOPROXY`) set to `direct` instead of `https://proxy.golang.org,direct` by chance?

FooSoft:
It is indeed just set to `direct`, which I believe is the out-of-the-box default on Fedora (where I'm using this).

bhcleek:
Thank you for confirming. I'll merge very soon.

bhcleek:
FWIW, you can unset `GOPROXY` or explicitly set it to the default value, `https://proxy.golang.org,direct`, in order to work around this problem, too.

PR #3498, opened 2023-02-04
syntax: add new `clear` builtin function

fzipp:
Support the new clear(x) builtin function added for Go 1.21.

https://github.com/golang/go/issues/56351
https://github.com/golang/go/commit/b89a840d6572c97a80ac78462b03122b83bc84e9

### DIFF
--- a/syntax/go.vim
+++ b/syntax/go.vim
@@ -47,7 +47,7 @@ hi def link     goFloats            Type
 hi def link     goComplexes         Type
 
 " Predefined functions and values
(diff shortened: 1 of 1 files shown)


bhcleek:
`clear` isn't in the language yet. There's no need for this until at least Go 1.21. This won't be merged until `clear` is a builtin in a released version of Go.

fzipp:
Sure, I just wanted to report it early, since this syntax file has several downstreams like vim, and by extension neovim, macvim, so that it will be available to most users by the time Go 1.21 is released.

Issue #3509, opened 2023-02-09
Feature Request: GoExport


sding3:
Add a new command `GoExport` which would export the current symbol if it's unexported currently, or unexport if it's exported currently.

Yes, this is possible with `GoRename` today, but it's a bit cumbersome.

bhcleek:
There's no need for a `:GoExport`; you can use the tab key with `:GoRename` to cycle between PascalCase and camelCase forms.

bhcleek:
@sding3 you may also be interested in changing your `g:go_gorename_prefill` thusly:
`let g:go_gorename_prefill = 'expand("<cword>") =~# "^[A-Z]" ? go#util#camelcase(expand("<cword>")) : go#util#pascalcase(expand("<cword>"))'`

If you do that, then the other case will be the default value when `:GoRename` is not given an argument, and you could easily enter: `:GoRename<Enter><Enter>` to do the export. In such a case, the `(go-rename)` mapping would also benefit.

sding3:
@bhcleek - thank you, that's exactly what I need!

PR #3510, opened 2023-02-10
config: change default g:go_gorename_prefill value

bhcleek:
Change the default value of `g:go_gorename_prefill` so that the default value will be the other of camelCase or PascalCase than what the word under the cursor is.

Perhaps this shouldn't be merged. The [original rationale](https://github.com/fatih/vim-go/pull/1465#issuecomment-336641007) for preserving the case where the current behavior was introduced is worth considering. #3509 is an argument in favor of changing which casing is used by default.

With this change, it's relatively simple to use the `(go-rename)` map to change an identifier from exported to unexported and vice versa (e.g. `map <leader>r <Plug>(go-rename)`)

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -393,8 +393,8 @@ endfunction
 
 function! go#config#GorenamePrefill() abort
   return get(g:, "go_gorename_prefill", 'expand("<cword>") =~# "^[A-Z]"' .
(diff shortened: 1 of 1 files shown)


bhcleek:
Closed, because user can configure `g:go_gorename_prefill` as to accomplish what this does for easy use with a mapping.

PR #3529, opened 2023-02-24
New option go_gopls_initialization_options

gduranceau:
This supports setting global gopls options, as described [here].(https://github.com/golang/tools/blob/master/gopls/doc/settings.md)

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -573,6 +573,10 @@ function! go#config#GoplsSettings() abort
   return get(g:, 'go_gopls_settings', v:null)
 endfunction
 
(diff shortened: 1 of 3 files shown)


bhcleek:
Vim-go already provides a way for you to leverage those settings with `g:go_gopls_settings`.

gduranceau:
@bhcleek 
Oh! I didn't realize they could be set this way. Maybe the documentation could mention that `go_gopls_settings` refers to [these settings](https://github.com/golang/tools/blob/master/gopls/doc/settings.md) and provide and example?

gduranceau:
Option already available

PR #3587, opened 2023-09-30
sameids: fix highlighting preceded by multibyte characters

### LINKED ISSUE #3586: Position for identifier highlight(GoSameIds) is miscalculated when unicode character is present
Presence of a single- or double-byte unicode symbol should not affect correctness of code highlighting.

Following screenshot demonstrates correct behavior. `fmt` is a Go module identifier and is **not** preceded with an unicode symbol on particular row :
image

### What happened instead?
Identifier highlight starts off the actual identifier position.

Following screenshot demonstrates the issue. `lib` is also a Go module identifier but now it **is** preceded with an unicode symbol, in a first string argument of `Println` function.
image

In the first case("bubble sort") the symbol is `⇒` (0x21d2), two-byte char so highlighting starts off by two columns.
In the second case("selection sort") the symbol is `á` (0xe1), single-byte char so highlighting starts off by one column.

Vim 9.0.1882, Linux x86_64

#### vim-go version:
current

bhcleek:
Fixes #3586

### DIFF
--- a/autoload/go/lsp.vim
+++ b/autoload/go/lsp.vim
@@ -872,7 +872,10 @@ function! s:sameIDsHandler(next, msg) abort dict
       continue
     endif
 
-    let l:result.sameids = add(l:result.sameids, [l:loc.range.start.line+1, l:loc.range.start.character+1, l:loc.range.end.character+1])
+    let l:line = getline(l:loc.range.start.line + 1)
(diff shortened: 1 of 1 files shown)


dunric:
Thanks for the fix! It seems it does the job in my cases I've tried with source codes of various projects.

I did suspected there has to be a missing position (re-)calculation for utf8/utf16 units contents.

PR #3588, opened 2023-10-11
Adds an option to disable gohtmltmpl

nkzren:
Added an option to disable automatic syntax highlighting for `gohtmltmpl`.

I wanted this behaviour because it conflicts with `.tmpl` files generated by [chezmoi](https://www.chezmoi.io/user-guide/templating/))

### DIFF
--- a/ftdetect/gofiletype.vim
+++ b/ftdetect/gofiletype.vim
@@ -7,7 +7,6 @@ set cpo&vim
 " Note: should not use augroup in ftdetect (see :help ftdetect)
 au BufRead,BufNewFile *.go setfiletype go
 au BufRead,BufNewFile *.s setfiletype asm
-au BufRead,BufNewFile *.tmpl set filetype=gohtmltmpl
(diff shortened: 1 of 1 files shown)


bhcleek:
Thank you for contributing. I think a better solution, though, may be to revert https://github.com/fatih/vim-go/pull/3146.

nkzren:
Great, makes sense to me as well since it seems to forcefully override vim's default behaviour

bhcleek:
Closing. This wasn't a revert of [3146](https://github.com/fatih/vim-go/pull/3146), and as explained there, what vim-go does here should be safe since the `template` filetype that ships with Vim is a subset of what vim-go does for the `gohtmltemplate`  filetype.

If you'd like your `.tmpl` files to have different behaviors, that can be accomplished by putting `au BufRead,BufNewFile *.tmpl set filetype=gohtmltmpl` in the right place (e.g. your vimrc or a file either in`$HOME/ftdetect` or `$HOME/after/ftdetect`) depending on where vim-go sits in your `runtimepath`.

PR #3596, opened 2023-11-05
Add installation instructions for packer.nvim

rhburt:
Add installation section for [packer.nvim](https://github.com/wbthomason/packer.nvim). Since the user will want to run the post-install hook for `:GoUpdateBinaries`, I believe it would be nice to have a copy/paste statement for this.


bhcleek:
Thank you for contributing.

I'm going to close this, though, because`packer.nvim` is no longer maintained. Also, vim-go is not going to maintain instructions for every kind of package manager out there; it is generally up to users to understand how to use their package managers.

PR #3607, opened 2023-12-14
Modify GoDoc split window key binding for GitBash compatibility

rwxguo:
The following keybinding resolved leading escape as a prefix (e.g. arrow keys),
But on `GitBash`, as there is no escape when press arraw keys, the first key-press will have no response.

```vim
nnoremap <buffer> <silent> <Esc>[ <Esc[
```

Thus, I made a GitBash platform checking to skip the key binding.


rwxguo:
This PR is created by accident, could anyone please remove it, thanks.

PR #3609, opened 2023-12-15
mingw: Add platform checking of GoDocBrowser for Cygwin on windows

rwxguo:
When using Cygwin/MSYS2 on windows, the `:GoDocBrowser` cannot popup the browser correctly.

Added `has('win32unix')` for platform compatibility.


rwxguo:
This PR is created by accident, could anyone please remove it, thanks.

PR #3610, opened 2023-12-15
doc: Add Gopls disable checking when view the `:GoDoc`

rwxguo:
When `g:go_gopls_enabled = 0`, `:GoDoc` and `:GoDocBrowser` will not work because they are still calling `go#lsp` functions.

I have added checking so that `:GoDoc` and `:GoDocBrowser` can work properly when `g:go_gopls_enabled = 0`


bhcleek:
thank you

rwxguo:
You're welcome

PR #3611, opened 2023-12-15
mingw: Add platform checking of GoDocBrowser for Cygwin on windows

rwxguo:
* When using Cygwin/MSYS2 on windows, the `:GoDocBrowser` cannot popup the browser correctly.
   * Added `has('win32unix')` for platform compatibility.
* Also resolved vim screen mess up after using `:silent` with external command
   * refer to `:h :silent`
   * https://vi.stackexchange.com/questions/2809/silent-makes-my-vim-go-blank

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -100,7 +100,7 @@ function! go#config#SnippetEngine() abort
 endfunction
 
 function! go#config#PlayBrowserCommand() abort
-    if go#util#IsWin()
+    if go#util#IsWin() || system('uname') =~ 'MINGW'
         let go_play_browser_command = '!start rundll32 url.dll,FileProtocolHandler %URL%'
     elseif go#util#IsMac()
         let go_play_browser_command = 'open %URL%'
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -100,8 +100,15 @@ function! go#config#SnippetEngine() abort
 endfunction
 
 function! go#config#PlayBrowserCommand() abort
-    if go#util#IsWin() || go#util#IsCygwin()
+    if go#util#IsWin()
         let go_play_browser_command = '!start rundll32 url.dll,FileProtocolHandler %URL%'
+    elseif go#util#IsCygwin()
+      " CYGWIN uses 'cygstart', whereas MSYS2 and GitBash uses 'start'
+      if system('uname') =~ 'CYGWIN'
+        let go_play_browser_command = '!cygstart rundll32 url.dll,FileProtoco          lHandler %URL%'
+      else
+        let go_play_browser_command = '!start rundll32 url.dll,FileProtocolHa          ndler %URL%'
+      endif
     elseif go#util#IsMac()
         let go_play_browser_command = 'open %URL%'
     elseif executable('xdg-open')
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -100,7 +100,7 @@ function! go#config#SnippetEngine() abort
 endfunction
 
 function! go#config#PlayBrowserCommand() abort
-    if go#util#IsWin() || system('uname') =~ 'MINGW'
+    if go#util#IsWin() || system('uname') =~ 'MINGW' || system('uname') =~ 'CYGWIN'
         let go_play_browser_command = '!start rundll32 url.dll,FileProtocolHandler %URL%'
     elseif go#util#IsMac()
         let go_play_browser_command = 'open %URL%'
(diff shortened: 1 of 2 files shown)


bhcleek:
Thank you

rwxguo:
welcome, happy holiday !

autoload/go/config.vim
```diff
@@ -100,7 +100,7 @@ function! go#config#SnippetEngine() abort
 endfunction
 
 function! go#config#PlayBrowserCommand() abort
-    if go#util#IsWin()
+    if go#util#IsWin() || system('uname') =~ 'MINGW' || system('uname') =~ 'CYGWIN'
```
bhcleek, line 103:
```suggestion
  if go#util#IsWin() || go#util#IsCygwin()
```

bhcleek (changes requested):
Thank you for the `redraw` and that check for cygwin. I'd like to see some changes to the cygwin detection, though.

In addition to the suggested edit I provided, can you add `go#util#IsCygwin` thusly:
```
function! IsCygwin()
  return has('win32unix`)
endfunction

Unfortunately, I don't have a windows machine to test with. Can you make the suggested edits and test to make sure they're correct?

I'm not sure vim-go should support mingw in the way that this PR and your others enable, because I think think it's possible that with mingw there's other use cases that would break with the changes here and elsewhere. I'm going to have to give this and your other PRs some thought. For now, can you remove the mingw detection in this PR?

**Comment by rwxguo** [comment]
@bhcleek Thank you for your guide and instruction. I added cygwin detection `go#util#IsCygwin`.

The reason using `uname` to check the platform is inspired by the [maven-wrapper](https://github.com/takari/maven-wrapper/blob/2b2c15adc38acfcf17533f5d7aff220ba87256ba/mvnw#L49) project, which is widely used in enterprise level java projects using SpringBoot etc.

Plus, I also using the vim built-in feature list `has('win32unix')` to make it more concise.

I have tested these factors on three major cygwin platforms, the result is for your reference:

| platform | has('win32') | has('win64') | has('win32unix') | uname |
| --- | --- | --- | --- | --- |
| Cygwin | 0 | 0 | 1 | CYGWIN_... |
| GitBash (64bit) | 0 | 0 | 1 | MINGW64_... |
| MSYS2 - mingw32.exe | 0 | 0 | 1 | MINGW32_... |
| MSYS2 - mingw64.exe | 0 | 0 | 1 | MINGW64_... |
| MSYS2 - msys2.exe | 0 | 0 | 1 | MSYS_... |

**Comment by rwxguo** [comment]
By the way, I found there is a util function `go#util#IsUsingCygwinShell` which might not correct, because it uses `go#util#IsWin`, but on all cygwin-like platform, `go#util#IsWin` always return 0.

```vim
 " Checks if using:
 " 1) Windows system,
 " 2) And has cygpath executable,
 " 3) And uses *sh* as 'shell'
function! go#util#IsUsingCygwinShell()
  return go#util#IsWin() && executable('cygpath') && &shell =~ '.*sh.*'
endfunction
```

The `go#util#IsUsingCygwinShell` is used in `path.vim`, for your information

bhcleek:
Yes, `go#util#IsUsingCygwinShell` is for a different kind of use case than you've been working on. The last I checked, it is correct for the use case for which it is intended.

autoload/go/util.vim
```diff
@@ -49,6 +49,17 @@ function! go#util#IsMac() abort
...
 
+" IsCygwin returns 1 if current OS is Cygwin/MSYS2/GitBash, 0 otherwise
+function! go#util#IsCygwin()
+  return !has('win32') &&
+        \ has('win32unix') &&
+        \ (
+        \ system('uname') =~ 'CYGWIN' ||
+        \ system('uname') =~ 'MINGW' ||
+        \ system('uname') =~ 'MSYS'
+        \ )
```
bhcleek, line 60: Please remove the `uname` checks. They are unnecessary.

There's also no need to check for `!has('win32')`, as you can see in the matrix you provided; `win32` and `win32unix` are mutually exclusive. Please remove that check.

rwxguo: Done. Thanks !

rwxguo:
Found some bug on cygwin (works fine on GitBash and MSYS2), I will re-open this PR after fixing it.

rwxguo:
As CYGWIN uses `cygstart` instead of `start`, but MSYS2 and GitBash uses `start`, I made the cygwin checking as a seprated "elseif", and use `rundll32` instead of `start rundll32` or `cygwin rundll32`. (treat cygwin as a separated system, rather than a windows variation)

Issue #3602, opened 2023-11-19
GoDef (and various other commands) don't work, potentially due to pathing issue


CrispyDrone:
### What did you do? (required: The issue will be **closed** when not provided)

Try to execute `:GoDef`.

### What did you expect to happen?

Cursor should jump to the definition of the identifier.

### What happened instead?

Nothing, except for a message `vim-go: [definition] SUCCESS` being printed in the status bar.

I suspect it is related to my trying to use bash shell provided by git, so maybe the underlying issue is the same as in this issue #3447

On that note, running `:GoRun` gives this output `"C:\Program Files\Git\bin\bash.exe" -c ""go" "run" "C:\Users\Me\Desktop\testing\go\main.go""
CreateFile C:UsersMeDesktoptestinggomain.go: The system cannot find the file specified.
shell returned 1`

#### vim-go version:

Not sure how to check this. The commit hash is 1401b576c6ac382529188d3d26cff866139f2f9a (from my plugged directory).

####  `vimrc` you used to reproduce:

<details><summary>vimrc</summary>

```vim
set shell=\"$PROGRAMW6432\Git\bin\bash.exe\"
set shellslash                                                  " use forward slash when completing path names

set rtp+=~/.vim/
call plug#begin('~/.vim/plugged')
Plug 'fatih/vim-go', { 'do': ':GoUpdateBinaries' }
call plug#end()
```
</details>

#### Vim version (first three lines from `:version`):

VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jul 29 2021 22:03:00)
MS-Windows 64-bit GUI version with OLE support
Included patches: 1-3249

#### Go version (`go version`):

go version go1.21.4 windows/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

set GO111MODULE=
set GOARCH=amd64
set GOBIN=C:\Users\Me\bin
set GOCACHE=C:\Users\Me\AppData\Local\go-build
set GOENV=C:\Users\Me\AppData\Roaming\go\env
set GOEXE=.exe
set GOEXPERIMENT=
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOINSECURE=
set GOMODCACHE=C:\Users\Me\go\pkg\mod
set GONOPROXY=
set GONOSUMDB=
set GOOS=windows
set GOPATH=C:\Users\Me\go
set GOPRIVATE=
[...]

Issue #3447, opened 2022-08-30
Incompatible with mingw64


bmccord2:
### What did you do? (required: The issue will be **closed** when not provided)
Ran plugin with mingw64 (git bash for windows).  I believe this problem likely would affect msys2 as well.

### What happened instead?
When opening a file paths are not properly formed in plugin.  I get a message like:
`vim-go: failed to load view for file:///C:/c/path/to/my/project: invalid workspace folder path: The system cannot find the path specified.; check that the casing of the configured workspace folder path agrees with the casing reported by the operating system`

Looked into the code a little bit, `has('Win32')` is used by the IsWin function in util.vim.  This returns 0 for msys2.  `has('Win32unix')` returns 1 as does `has('unix')`.

Also, GoInstallBinaries did not work without explicitly setting GOBIN variable as GOPATH/bin (not sure if this is related or not, I no longer have the error messages, but they complained about GOBIN).

#### vim-go version:
4d6962d8

####  `vimrc` you used to reproduce:

<details><summary>vimrc</summary>

```vim
set nocompatible              " be iMproved, required
filetype off                  " required

set rtp+=~/.vim/bundle/Vundle.vim
call vundle#begin()
"call vundle#begin('~/some/path/here')

Plugin 'VundleVim/Vundle.vim'


Plugin 'heavenshell/vim-pydocstring'
Plugin 'wannesm/wmgraphviz.vim'
Plugin 'vim-scripts/reload.vim'
Plugin 'vim-scripts/vim-misc'
Plugin 'romainl/Apprentice'
Plugin 'iamcco/markdown-preview.nvim'
Plugin 'fatih/vim-go'



call vundle#end()            " required
filetype plugin indent on    " required
"filetype plugin on

colorscheme default

if &diff
    colorscheme molokai
endif

:set tabstop=4
:set shiftwidth=4
:set softtabstop=4
:set expandtab
:set autoindent

augroup myvimrchooks
    au!
    autocmd bufwritepost .vimrc source ~/.vimrc
augroup END

filetype plugin on

autocmd FileType yaml setlocal shiftwidth=2 tabstop=2 softtabstop=2

nmap <silent> <c-p> <Plug>(pydocstring)

nnoremap <leader>g :YcmCompleter GoToDefinitionElseDeclaration<CR>


```
</details>

#### Vim version (first three lines from `:version`):

```
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Nov 22 2021 19:31:05)
[...]

PR #3612, opened 2023-12-16
mingw: Resolve path issue, make gopls work on GitBash (`:GoDoc`, `:GoInfo` and `:GoDef`)

rwxguo:
This patch can resolve the following issues:

* #3602 
* #3447 

The root cause is that `gopls` cannot identify the correct path sent from `vim-go`.

GitBash mixed the 2 path systems (Win & Unix) which benefit from two world, but also brought some problem. The path like `/c/mypath` is used to represent the corresponding windows path `c:\mypath`. However, when communicating with `gopls`, the path `/c:/mypath` (please note there is a `:`) is used. It is probably because the GitBash parse the path from/to the `gopls` with a forward slash prefix `/`.

I didn't modify the logic of `go#util#IsWin()` because it is called by many places in the program, to prevent other issues happen. Instead, I treated the Cygwin (including MSYS2 and GitBash) system as another system (Neither Windows, nor Unix). 

Vim provide a cygwin checking `has('win32unix')` and it should be enough. Refer to the [bhcleek's reply](https://github.com/fatih/vim-go/pull/3611#discussion_r1435775625)

I have only tested `:GoDoc`, `:GoInfo` and `:GoDef`, which I think are the most popular functions for `GitBash` users.

### DIFF
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -129,10 +129,8 @@ function! go#def#jump_to_declaration(out, mode, bin_name) abort
     let parts = split(out, '\(^[a-zA-Z]\)\@<!:')
   elseif has('win32unix')
     " remove comma in cygwin path e.g. '/c:/path'
-    if l:out[0:8] != '/cygdrive'
-      if l:out[2:3] is# ':/'
-        let l:out = l:out[0:1] . l:out[3:]
-      endif
+    if l:out[0:8] != '/cygdrive' && l:out[2:3] is# ':/'
+      let l:out = l:out[0:1] . l:out[3:]
     endif
     let parts = split(out, ':')
   else
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -127,7 +127,8 @@ function! go#def#jump_to_declaration(out, mode, bin_name) abort
   if go#util#IsWin()
     let parts = split(out, '\(^[a-zA-Z]\)\@<!:')
   elseif system('uname') =~ 'MINGW' || system('uname') =~ 'CYGWIN'
-    if l:msguri[2:3] is# ':/'
+    " remove comma from path before split. e.g. /c:/path to /c/path
+    if l:out[2:3] is# ':/'
       let l:out = l:out[0:1] . l:out[3:]
     endif
     let parts = split(out, ':')
--- a/autoload/go/def.vim
+++ b/autoload/go/def.vim
@@ -126,8 +126,10 @@ function! go#def#jump_to_declaration(out, mode, bin_name) abort
   let out = split(final_out, go#util#LineEnding())[0]
   if go#util#IsWin()
     let parts = split(out, '\(^[a-zA-Z]\)\@<!:')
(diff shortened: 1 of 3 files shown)


autoload/go/def.vim
```diff
@@ -126,6 +126,11 @@ function! go#def#jump_to_declaration(out, mode, bin_name) abort
   let out = split(final_out, go#util#LineEnding())[0]
   if go#util#IsWin()
     let parts = split(out, '\(^[a-zA-Z]\)\@<!:')
+  elseif system('uname') =~ 'MINGW' || system('uname') =~ 'CYGWIN'
+    if l:msguri[2:3] is# ':/'
```
bhcleek, line 130: `l:msguri` doesn't exist. What is this supposed to be? Is it supposed to be `l:out`?

rwxguo: Yes, my mistake. It should be `l:out`. Thanks.

rwxguo:
BTW, I would further change my cygwin checking to using the `go#util#IsCygwin` created in this [PR](https://github.com/fatih/vim-go/pull/3611#pullrequestreview-1795474217) after it's merged to make code clean.

rwxguo:
Found some bug on cygwin (works fine on GitBash and MSYS2), I will re-open this PR after fixing it.

rwxguo:
Add checking for CYGWIN as CYGWIN path has `prefix /cygdrive` compare to MSYS2/GitBash.

bhcleek:
@rwxguo I k now this and the other PRs you've submitted, #3611 and #3608, have been sitting for a while. Thank you for submitting these, and my apologies for the delay. I wanted to give you some idea for why they've been sitting unmerged, though.

These are all related and go to a common problem, so I want to merge them via a manual octopus merge instead of relying on GitHub's merge pull request functionality. I plan to do that octopus merge locally on my machine and then test this fully. I suspect these three PRs are not sufficient to fully address the problems littered throughout vim-go when trying to use a Vim compiled with mingw, and if vim-go is going to tackle this, then I'd like to address it fully if possible.

Also, I'm not convinced this is the right way to go. When using vim-go with Vim compiled with mingw, there's a strong argument that the Go version used to compile the helper binaries (e.g. `gopls`, `dlv`, etc.) should also have been compiled with mingw (e.g. https://packages.msys2.org/base/mingw-w64-go). I think that might resolve all these problems naturally, though I'm not certain. I'm sure there would be `PATH`, `GOPATH`, and `GOBIN` considerations, too, if that approach were chosen. As an alternative to that idea, users could use a Vim that's native to their system instead of trying to use a mingw Vim, vim-go, and binaries that are not derivative of mingw.

rwxguo:
@bhcleek Totally agree. Thanks for your kind reply and detailed explanation. Thanks for keeping an eye on my changes, that's very kind of you, and your direction is totally correct. I myself also use native vim with vim-go. Mingw has many problem not only in vim-go. And I will feel contented just if my workaround can help anyone who are trmporarely using vim-go with mingw at the moment. But that workaround should not be merged just as you said, which is not good way enough to resolve the issue of mingw. Take it easy, and happy coding. ;)

PR #3634, opened 2024-01-22
Update syntax for version in Go mod file directive

matthewhughes934:
Since go1.21 this directive supports the same version format as the `toolchain` directive[1] (though it doesn't support any suffixes). Specifically it supports formats such as:

* `1.20` (this was supported previously)
* `1.21.0`
* `1.21rc1`

Note: patch versions do not have release candidates (e.g. `1.21.1rc2` is not valid)

[1] https://go.dev/ref/mod#go-mod-file-go

### DIFF
--- a/syntax/gomod.vim
+++ b/syntax/gomod.vim
@@ -61,7 +61,7 @@ syntax match gomodToolchainVersion "go1\(.\d\+\)\{,2\}\(rc\d\+\)\?\([ \t-].*\)\?
 highlight default link gomodToolchainVersion Identifier
 
 " match go versions
-syntax match gomodGoVersion "1\.\d\+" contained
+syntax match gomodGoVersion "\(1.\d\+\)\(\(.\d\+\)\|\(rc\d\+\)\)\?" contained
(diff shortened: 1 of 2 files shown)


syntax/gomod.vim
```diff
@@ -61,7 +61,7 @@ syntax match gomodToolchainVersion "go1\(.\d\+\)\{,2\}\(rc\d\+\)\?\([ \t-].*\)\?
 highlight default link gomodToolchainVersion Identifier
 
 " match go versions
-syntax match gomodGoVersion "1\.\d\+" contained
+syntax match gomodGoVersion "\(1.\d\+\)\(\(.\d\+\)\|\(rc\d\+\)\)\?" contained
```
matthewhughes934, line 64: I think this is clearer with magic: `\v(1.\d+)((.\d+)|(rc\d+))?`. I think I've seen this used elsewhere in these syntax files so maybe a worthwhile change?

bhcleek: I'm ok without the magic.

bhcleek (approved):
Thank you

PR #3637, opened 2024-01-25
Fix `goPackageComment` highlighting too broad

matthewhughes934:
Previously this would highlight any comment before a line starting `package`, for example

    var (
      // This comment would be highlighted as a goPackageComment
      packages []string
    )

The package clause is only valid when followed by a space[1], so include this restriction

[1] https://go.dev/ref/spec#Package_clause

### DIFF
--- a/autoload/go/highlight_test.vim
+++ b/autoload/go/highlight_test.vim
@@ -829,6 +829,77 @@ function! Test_gomodGoVersion() abort
   endtry
 endfunc
 
+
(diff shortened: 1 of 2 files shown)


bhcleek (approved):

matthewhughes934:
@bhcleek what do you think about upstreaming this change? I've tried that with https://github.com/vim/vim/pull/14213

Issue #3664, opened 2024-06-03
Errors while installing v1.28 in NVIM v0.9.5


paskozdilar:
### What did you do? (required: The issue will be **closed** when not provided)

I executed `:GoUpdateBinaries`:

```
[...]
vim-go: Error installing golang.org/x/tools/cmd/guru@master: go: golang.org/x/tools/cmd/guru@master (in golang.org/x/tools/cmd/guru@v0.1.0-deprecated):
vim-go:         The go.mod file for the module providing named packages contains one or
vim-go:         more replace directives. It must not contain directives that would cause
vim-go:         it to be interpreted differently than if it were the main module.
[...]
vim-go: Error installing honnef.co/go/tools/cmd/keyify@master: go: honnef.co/go/tools/cmd/keyify@master: module honnef.co/go/tools@master found (v0.5.0-0.de
v.0.20240601151239-d39a04f5c28a), but does not contain package honnef.co/go/tools/cmd/keyify
[...]
```

The problem can be reproduced by building a Docker image with the following Dockerfile:

```Dockerfile
FROM ubuntu:24.04
RUN apt-get -y update
RUN apt-get -y install git golang-go neovim
RUN git clone -b v1.28 https://github.com/fatih/vim-go.git ~/.local/share/nvim/site/pack/plugins/start/vim-go
RUN nvim --headless -c GoInstallBinaries
```

### What did you expect to happen?

I expected `:GoInstallBinaries` to finish without errors.

### What happened instead?

Instead, `:GoInstallBinaries` finished with errors.

#### vim-go version:

v1.28

Same thing happens with my local version (output of `:PlugSnapshot`):

silent! let g:plugs['vim-go'].commit = 'feef9b31507f8e942bcd21f9e1f22d587c83c72d'

####  `vimrc` you used to reproduce:

None.

#### Vim version (first three lines from `:version`):

NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1703358377

Same thing happens with my local version:

NVIM v0.10.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

#### Go version (`go version`):

go version go1.22.2 linux/amd64

Same thing happens with my local version:

go version go1.22.3 linux/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/root/.cache/go-build'
GOENV='/root/.config/go/env'
[...]

PR #3665, opened 2024-06-03
Remove support for GoKeyify

paskozdilar:
Standalone keyify tool has been removed from the go-tools repository and integrated with gopls:

https://github.com/dominikh/go-tools/issues/769

This PR would close the issue https://github.com/fatih/vim-go/issues/3664

---

Since GoKeyify was already only enabled in GOPATH mode, and GOPATH mode has been [deprecated since Go 1.16](https://go.dev/wiki/GOPATH) and is planned to be removed in the future, I believe completely removing it is the right decision.


bhcleek:
thank you

PR #3668, opened 2024-06-05
Add config to exclude certain binaries from updating

EVODelavega:
The only way to skip (or set a specific version) for the binaries installed via `:GoInstallBinaries` or `:GoUpdateBinaries` so far is to pass in a specific version. This works, but is somewhat clunky. Projects I'm working often use a specific version of `golangci-lint` as part of the CI. Instead of having to create mappings, or keep track of which versions and run the correct `:GoUpdateBinaries golangci-lint@v1.2.3`, I figured it'd be much easier to let me manage the version of that binary myself, and have `:GoUpdateBinaries` skip installing/updating that binary outright.

Another gripe I have with passing in specific versions of a binary is that, in order to skip one of the binaries, I'd need to list all of the binaries I do want to update (or update all, and then revert one to a specific version). IMHO, adding an exclude list is a much simpler solution, hence the PR.

When adding the docs for this new config variable, I also noticed a small typo in the docs for `g:go_debug_breakpoint_sign_text`.


bhcleek:
Thank you for contributing again.

With this change, it looks like `:GoUpdateBinaries` cannot be used to explicitly install/update a binary that is in `g:go_exclude_binaries`. Given `let g:go_exlude_binaries = ['dlv']`, `:GoUpdateBinaries dlv` would refuse to update `dlv`, even though the user is explicitly requesting it.

I'm open to this change conceptually, but I believe users should be able to update a tool that's given as an explicit argument. WDYT?

EVODelavega:
Hi, and yes. As it stands, the `go_exclude_binaries` list does override any and all attempts to install/update binaries, even if explicitly specified. I do agree that args passed directly to the commands ought to take precedence over config, so I'll revisit the changes made here a bit, to skip the filtering if binaries are explicitly specified

plugin/go.vim
```diff
@@ -130,8 +130,14 @@ function! s:GoInstallBinaries(updateBinaries, ...)
     endfor
   else
     let l:packages = s:packages
+    " Filter packages from exclude list, if no binaries were explicitly
+    " specified.
+    for l:bin in go#config#GoExcludeBinaries()
+      call remove(l:bin, l:packages)
+    endfor
```
bhcleek, line 137: `l:packages` needs to copy `s:packages` (use the `copy` or `deepcopy`) since entries are being removed so that `s:packages` is not altered.

Also, `:GoInstallBinaries` should probably not skip anything that's missing. I think that means that instead of this removal, the check around line 160 that currently reads
```
      if a:updateBinaries == 1
        echo "vim-go: Updating " . l:binary . ". Reinstalling ". importPath . " to folder " . go_bin_path
```
should probably be something like
```
      if a:updateBinaries == 1
        if index(go#config#GoExcludeBinaries(), l:bin) != -1
            echo "vim-go: Skipping out of date " . l:binary . ". "
            continue
        endif
        echo "vim-go: Updating " . l:binary . ". Reinstalling ". importPath . " to folder " . go_bin_path
```

PR #3675, opened 2024-08-02
debug: fix double substitution for paths in goto_file from stacktrace window

gremat:
When the content of the stacktrace window is computed in s:show_stacktrace(), s:substituteRemotePath() is already used to map remote paths to local paths. goto_file() is intended to jump to the desired location in the source code which resides on the local machine. Thus it should use the _local_ path and not translate the path back to the remote version.


bhcleek:
Thank you

Issue #3453, opened 2022-09-21
Variables in GODEBUG_VARIABLES pane truncated at 20 bytes


bgarber:
### What did you do?
Trying to debug a Go program using `dlv` and `vim-go`. The variables in the `GODEBUG_VARIABLES` pane seem to be truncated at 20 bytes: 
<img width="455" alt="image" src="https://user-images.githubusercontent.com/1001525/191589162-38afa710-ea9a-4606-874f-cd67e871284e.png">

The `GODEBUG_OUTPUT` shows the entire string correctly, when I execute `:GoDebugPrint hello`:
<img width="422" alt="image" src="https://user-images.githubusercontent.com/1001525/191589515-633ef023-0c9a-4b6a-a21e-8d7bd11aa6c8.png">

I have already edited my `~/.config/dlv/config.yml` file to permit longer strings:
<img width="274" alt="image" src="https://user-images.githubusercontent.com/1001525/191589753-c0f7ca9a-0894-4669-bb5c-923446e0bbbc.png">

Am I missing something?

### What did you expect to happen?
I expected the whole string to appear in the variables pane.

### What happened instead?
Strings (and most of the data) get truncated at 20 bytes.

#### vim-go version:
latest master (22b2273cfe562ac1c1af976ce77f18a3b1776f3c)

####  `vimrc` you used to reproduce:

<details><summary>vimrc</summary>

```vim
" Setup for vim-go
let g:go_fmt_command = "goimports"
let g:go_metalinter_autosave = 1
" let g:go_debug = ['shell-commands']
" let g:go_auto_type_info = 1

if exists('$GOPATH')
    autocmd BufRead *.go
        \ let s:temp = matchlist(expand('%:p'),
           \ $GOPATH . '/src/\(github.com/bgarber/[^/]\+\)')
        \| if len(s:temp) > 1 | exe 'silent :GoGuruScope ' . s:temp[1] . '/...' | endif
        \| unlet s:temp
endif
```
</details>

#### Vim version (first three lines from `:version`):

VIM - Vi IMproved 9.0 (2022 Jun 28)
Included patches: 1-313
Compiled by 'http://www.opensuse.org/'

#### Go version (`go version`):

go version go1.18.5 linux/amd64

#### Go environment
<details><summary><code>go env</code> Output:</summary><br><pre>

GO111MODULE="auto"
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/bgarber/.cache/go-build"
[...]

PR #3676, opened 2024-08-02
debug: make local variables load config configurable

gremat:
Note though that usage of a `LoadConfig` is not consistent:
* If `debug.vim:s:eval()` is called directly, most notably from `s:expand_vars()` (i.e., when the user tries to expand a variable in the variable window), [`RPCServer.Eval()`](https://pkg.go.dev/github.com/go-delve/delve@v1.23.0/service/rpc2#RPCServer.Eval) is used which currently (i.e., delve@v1.23.0) does not allow to specify a `LoadConfig`.
* If `debug.vim:s:eval()` is called from `go#debug#Print()` (e.g., via `:GoDebugPrint`), [`RPCServer.Command()`](https://pkg.go.dev/github.com/go-delve/delve@v1.23.0/service/rpc2#RPCServer.Command) is used and [`DebuggerCommand`](https://pkg.go.dev/github.com/go-delve/delve@v1.23.0/service/api#DebuggerCommand) is provided with a `ReturnInfoLoadConfig`. However, those values and the values used in the variables window differ.

### DIFF
--- a/autoload/go/config.vim
+++ b/autoload/go/config.vim
@@ -217,6 +217,10 @@ function! go#config#LspLog() abort
   return g:go_lsp_log
 endfunction
 
+function! go#config#DebugLocalVariablesLoadConfig() abort
+  return get(g:, 'go_debug_local_variables_load_config', {'MaxStringLen': 20, 'MaxArrayValues': 20, 'MaxVariableRecurse': 10})
+endfunction
+
 function! go#config#SetDebugDiag(value) abort
   let g:go_debug_diag = a:value
 endfunction
--- a/autoload/go/debug.vim
+++ b/autoload/go/debug.vim
@@ -1305,7 +1305,7 @@ function! s:update_variables() abort
   " MaxStructFields is the maximum number of fields read from a struct, -1 will read all fields.
   let l:cfg = {
         \ 'scope': {'GoroutineID': s:goroutineID()},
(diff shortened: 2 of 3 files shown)


bhcleek:
I'm inclined to close this without merging it. Some ideal value for the values doesn't exist and what's currently used is sufficient. I talked about this some previously: https://github.com/fatih/vim-go/issues/3453#issuecomment-1255225558.

gremat:
First, I don't quite understand why to limit the string length to (some may say: only) 20 characters. The variable window has `nowrap` by default so it wouldn't even hurt to grab 40 or even 100 chars in the first place and then be fine with it when they are simply hidden.

But my main concern with this MR wasn't the string length but the other two values. And I think your argumentation is not consistent in this regard. If "Twenty characters is generally sufficient", then why have 20 (!) array/slice elements or a recursion depth of 10 (which might get _really_ huge). In particular, when you always have the option with vim-go to expand a particular array on demand. For example, I use `{'MaxStringLen': 42, 'MaxArrayValues': 3, 'MaxVariableRecurse': 5}` with my patch to get a decent overview first, and then I can still display more if required.

So my intention was in fact to limit potential discussion about those values by just making them configurable. And apparently there was already some demand for it. Anyhow: if you think this will open up too much discussion/demand for other configuration possibilities (I don't have an overview of other or past issues/MRs), then I am totally fine having this closed.

PR #3691, opened 2024-11-11
fix fatih/vim-go/issues/3690: set `noshellslash` only on win32

happy-dude:
In neovim nightly, neovim/neovim/pull/28400 was merged:

> The value of shellslash should show true now outside of Windows.
> set shellslash should work on POSIX while set noshellslash will fail

This PR for vim-go wraps `set noshellslash` between a `if has("win32")` conditional to ensure that the functions no longer error in POSIX environments.


bhcleek:
Thank you for contributing

PR #3723, opened 2025-07-23
compiler: detect `GNUmakefile`

bbasata:
From `make(1)`:

If no -f option is present, make will look for the makefiles GNUmakefile, makefile, and Makefile, in that order.

### DIFF
--- a/compiler/go.vim
+++ b/compiler/go.vim
@@ -19,7 +19,7 @@ endif
 
 let s:save_cpo = &cpo
 set cpo-=C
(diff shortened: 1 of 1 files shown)


bbasata:
cc: @bhcleek 👋 – would you be interested in accepting this PR? I work with Go projects that use `GNUmakefile`, and it's a documented option for `make(1)`.

bhcleek:
Thank you for the nudge and for contributing!

PR #3744, opened 2026-03-25
debug: fix display of string values that are parsed into blob type

gremat:
String values returned by delve may be parsed into vim's `Blob` type, e.g., if NUL bytes are encountered. This patch checks the `value`'s type and uses a string representation for blob type that is close to what golang's `%q` would print.


bhcleek (approved):

bhcleek:
All fair points. Thank you again for contributing.

PR #3751, opened 2026-07-24
syntax: fix goVar and goConst highlighting bug

### LINKED ISSUE #3699: Vim: Adding strings in multiple lines in var statement breaks Go syntax highlighting
I've opened [an issue](https://github.com/neovim/neovim/issues/32427) on NeoVim repository, and I was directed to open an issue here.

The same issue also exists on Vim. Vim `./runtime/syntax/go.vim` is (allegedly) up-to-date with `vim-go`, but I don't know how to verify that since injecting the `go.vim` file from `vim-go` into `vim` repository causes errors. I assume Vim does some kind of transformations on it?

---

Anyways, here's the details of the issue:

The following Go snippet breaks syntax highlighting in Vim:

```go
package main

var (
	host  string
	usage string = "" +
		"\"\""
)

func main() {}
```

This can be reproduced in a Docker container by downloading the latest nightly Vim AppImage:

```Dockerfile
FROM ubuntu:24.04

RUN apt-get -y update && apt-get -y install curl

WORKDIR /app

RUN curl -LO https://github.com/vim/vim-appimage/releases/download/v9.1.1106/Vim-v9.1.1106.glibc2.29-x86_64.AppImage

RUN chmod u+x Vim-v9.1.1106.glibc2.29-x86_64.AppImage

RUN ./Vim-v9.1.1106.glibc2.29-x86_64.AppImage --appimage-extract

RUN echo 'package main' >> main.go
RUN echo '' >> main.go
RUN echo 'var (' >> main.go
RUN echo '	host  string' >> main.go
RUN echo '	usage string = "" +' >> main.go
RUN echo '		"\"\""' >> main.go
RUN echo ')' >> main.go
RUN echo '' >> main.go
RUN echo 'func main() {}' >> main.go

CMD ["/app/squashfs-root/AppRun", "main.go"]
```

Running `docker build . -t test && docker run --rm -it test` shows the following screen:

Image

The statement `func main() {}` and everything after it is completely devoid of syntax highlighting.

paskozdilar:
Fix `goVar` and `goConst` hightlighting bug by excluding `goImportString` from `goVar` and `goConst` blocks.

Fixes https://github.com/fatih/vim-go/issues/3699

### DIFF
--- a/autoload/go/highlight_test.vim
+++ b/autoload/go/highlight_test.vim
@@ -900,6 +900,39 @@ function! Test_goPackageComment_highlight() abort
   endtry
 endfunc
 
+function! Test_goStringEscapedQuotesInVarBlockHighlight() abort
+  syntax on
+  let g:go_gopls_enabled = 0
+
+  let l:wd = getcwd()
+  let l:dir = gotest#write_file('highlight/escaped_quotes_var.go', [
+        \ 'package main',
+        \ '',
+        \ 'var (',
+        \ printf('%shost string', "\t"),
+        \ printf('%susage string = "" +', "\t"),
+        \ printf('%s"\"\""', "\t\t"),
+        \ ')',
+        \ '',
+        \ 'func main() {}',
+        \ ])
+
+  try
+    " Ensure string with escaped quotes inside var () is matched as goString (not goImportString)
+    call cursor(6, 3)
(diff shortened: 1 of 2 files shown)


bhcleek:
Thank you. I'm probably going to take this and the PR that I put up to fix it, too. Technically, either one will solve this problem, but having the changes in both will make the syntax file more precise.

PR #3752, opened 2026-08-03
fix: don't use GOBIN env var if not defined

carlosonunez:
This commit works around an issue I encountered wherein `gopls` fails to
load despite `:GoInstallBinaries` downloading and installing it. This
happened because I was using `asdf` to manage my Go version and `asdf`
does not set `$GOBIN` in the environment by default.

All that's being done here is getting `g:go_bin` to `$GOBIN` if its
defined in the environment (usually the case when you install `go` with
Homebrew or something like that) and sets it to the output of `go env
GOBIN` otherwise.

Signed-off-by: Carlos Nunez <13461447+carlosonunez@users.noreply.github.com>

### DIFF
--- a/autoload/go/path.vim
+++ b/autoload/go/path.vim
@@ -80,7 +80,11 @@ function! go#path#BinPath() abort
   " we can use it, otherwise use default GOPATH
   let l:bin_path = go#util#env('gobin')
   if l:bin_path isnot ''
-    let l:bin_path = $GOBIN
+    if $GOBIN isnot ''
+      return $GOBIN
+    else
+      return l:bin_path
+    endif
   else
     let l:go_paths = split(go#path#Default(), go#util#PathListSep())
     if len(l:go_paths) == 0
@@ -88,7 +92,6 @@ function! go#path#BinPath() abort
     endif
     let l:bin_path = expand(l:go_paths[0] . '/bin/')
   endif
-
   return l:bin_path
 endfunction
 


carlosonunez:
NOTE: This PR is meant to address #3753.

bhcleek:
This seems to be doing the opposite of what the description says: it's preferring `$GOBIN` regardless of what `go env gobin` returns. Is that your intent?

carlosonunez:
Yes; this preserves existing behavior (use GOBIN if "go env gobin" is set). The fix here is to use output from "go env gobin" if $GOBIN is empty.On Aug 3, 2026, at 14:37, Billie Cleek ***@***.***> wrote:﻿bhcleek left a comment (fatih/vim-go#3752)
This seems to be doing the opposite of what the description says: it's preferring $GOBIN regardless of what go env gobin returns. Is that your intent?

bhcleek:
Vim-go already uses `go env GOBIN` and `go env GOBIN` gets its value from $GOBIN. How are you getting into some state where `go env GOBIN` returns something different than $GOBIN?

carlosonunez:
﻿I use asdf to manage Golang versions. asdf doesn't export it when the golang version is reshimmed, or at least it doesn't in my setup.In this case, the .tool-versions file in my $HOME links to Go 1.26, but I was working with a project whose .tool-versions file requests Go 1.21."cd"ing into the project's directory triggers the asdf reshim to 1.21 as expected, but "echo $GOBIN" returns nothing.On Aug 3, 2026, at 16:02, Billie Cleek ***@***.***> wrote:﻿bhcleek left a comment (fatih/vim-go#3752)
Vim-go already uses go env GOBIN and go env GOBIN gets its value from $GOBIN. How are you getting into some state where go env GOBIN returns something different than $GOBIN?

carlosonunez:
So I've confirmed that this was a PEBKAC. Sort of.

The `golang` plugin for `asdf` tells you to source (or have `.bash_profile/.bashrc` source) a script after installing it; see [here](https://github.com/asdf-community/asdf-golang#use). I thought that installing the plugin was all I needed to do to handle exports.

However, assuming that the aforementioned script is not sourced, you'll get output that looks like this:

```bash
$: echo "GOBIN in env: '$GOBIN'"
GOBIN in env: ''
$: echo "GOBIN from go env: '$(go env GOBIN)'"
GOBIN from go env: '/Users/cn/.asdf/installs/golang/1.26.5/bin'
```

Which causes the `gopls` detection issue brought forth by this issue.

@bhcleek: It's your team's call on whether y'all this PR should get merged in. Consider this documentation for future readers (or their agents) otherwise.

carlosonunez:
Thanks for your consideration.

